PR #5, opened 2021-02-09
Add memory manager template parameter to allocator

stephenswat:
As we add more memory managers to the code base, it will be useful to be able to switch them out at compile time using template parameters. This commit adds a new template parameter (and a sensible default) to the allocator class, which can be used to configure the memory manager used at compile time.

### DIFF
--- a/core/include/vecmem/allocators/allocator.hpp
+++ b/core/include/vecmem/allocators/allocator.hpp
@@ -17,7 +17,7 @@
 namespace vecmem {
 
    /// STL allocator using the "currently active" memory manager
-   template< typename TYPE >
+   template< typename TYPE, typename MemMgr = memory_manager >
    class allocator {
 
    public:
@@ -40,21 +40,21 @@ namespace vecmem {
 
       /// Allocate a requested amount of memory
       pointer allocate( size_type n, const void* = nullptr ) {
-         memory_manager_interface& mmgr = memory_manager::instance().get();
+         memory_manager_interface& mmgr = MemMgr::instance().get();
          return static_cast< pointer >(
             mmgr.allocate( n * sizeof( value_type ) ) );
       }
 
       /// Deallocate a previously allocated block of memory
       void deallocate( pointer ptr, size_type ) {
-         memory_manager_interface& mmgr = memory_manager::instance().get();
+         memory_manager_interface& mmgr = MemMgr::instance().get();
          mmgr.deallocate( ptr );
       }
 
       /// Only initialise the memory if it is host-accessible.
       template< typename U, typename... Args >
       void construct( U* ptr, Args&&... args ) {
-         memory_manager_interface& mmgr = memory_manager::instance().get();
+         memory_manager_interface& mmgr = MemMgr::instance().get();
          if( mmgr.is_host_accessible() ) {
             new( ptr ) U( std::forward< Args >( args )... );
          }
@@ -64,7 +64,7 @@ namespace vecmem {
       /// Only destroy objects in host-accessible memory.
       template< typename U >
       void destroy( U* ptr ) {
-         memory_manager_interface& mmgr = memory_manager::instance().get();
+         memory_manager_interface& mmgr = MemMgr::instance().get();
          if( mmgr.is_host_accessible() ) {
             ptr->~U();
          }


krasznaa (commented):
This is not quite what I had in mind...

I thought you only wanted to make the "memory type" a template parameter. Not the memory manager instance.

At the same time I'm not sure myself what would be the best approach. :frowning: Since as I wrote in #7, I don't think we should make the memory types global.

I'll have to think a bit about it...

stephenswat:
Hmm, okay, I think we might be a little out of sync on how we want to design this library in that case. My understanding was that we wanted to have essentially drop-in replacements for memory allocation schemes, which I think would be best achieved using template parameters like this. Without that I don't see how we can test different memory managers without _a lot_ of boilerplate code all over the place. Unless you want to have dynamic memory managers at runtime, but that seems less than ideal to me.

krasznaa:
Have a look at what I did in the unit tests:

https://github.com/acts-project/vecmem/blob/main/tests/cuda/test_cuda_allocators.cpp#L61-L89

My idea was that the current memory manager should be something that's decided at runtime. This way we can actually compile our EDM code, and have it interact with different kinds of memory depending on what the runtime wants. I don't want all of our code to be templated all the time. I believe Acts has gone way overboard with its templating by now... :frowning:

What I was thinking of was something along the lines of #7. So that the "memory type" would be a template parameter of the allocator. Not the specific memory manager. I'm still just pondering how the memory type should be properly abstracted to work with all the languages...

stephenswat:
If you think ACTS has gone overboard with templating I think it's fair to say that having this stuff decided at runtime is alright for this project, as it should be significantly less sensitive to the runtime overhead of late binding. However, the `std` containers use templated allocators, of course, and hopping back to doing things on the value level when you've been doing stuff on the type level at a higher level of abstraction seems like strange design to me.

paulgessinger:
Keep in mind that depending on whether allocation operations happen frequently or not, the virtual call might be an overhead that can be prevented with templating and concrete type.

While I think @krasznaa's general weariness concerning templating in ACTS is somewhat warranted, keep in mind that we believe a **large** chunk of our template woes come from Eigen template instantiations.

krasznaa:
Let's calm down a bit...

You should create a design proposal for how you would imagine things to work then. Note that in this PR you templated the type of this singleton class:

https://github.com/acts-project/vecmem/blob/main/core/include/vecmem/memory/memory_manager.hpp

Which is not the memory manager. It's a singleton that points at runtime to the currently active memory manager. My point was that templating this really doesn't make sense.

If you would want to template the type of the memory manager, you would have to provide types like `vecmem::cuda::direct_memory_manager` and similar to the allocator class. Which **is** a possible, and defensible design. But I don't quite like the idea of either locking our EDM to a specific memory management style, or keeping the EDM templated until the very last step where the user writes his/her code.

And hence, I would rather make the memory type templated. As one should actually decide at the code structure level whether something will be managed manually between the host and device, or it will be left up to the GPU language's runtime to move data between the two automatically. But whether the allocation happens one-by-one, using an arena allocator, CUB, or anything else, should not matter to the EDM code at compile time.

paulgessinger:
All good, @krasznaa :)

I looked at the existing code, and found your design so far totally reasonable, especially for prototyping. 

I'm just chiming in.

stephenswat:
Sorry if I come across as anything other than calm here, that's absolutely not my intention. Some of the subtleties may be getting lost in the medium of Github merge requests. :smile: 

Anyway, you're right, I've misunderstood your design here. The change made in this merge request doesn't make any sense in that framework. However, I am also not really convinced that this global singleton design is the right way to go. Global state is evil, and while it works well in simple use cases like these unit tests, I foresee a lot of problems if (when) we want to start supporting heterogeneous systems, as well as multi-threaded execution: the user would have to constantly switch the global state to whichever manager they want to use at a time, and users would also have to lock the state in multi-threaded programs, right? Or perhaps you've thought of a solution to this that I'm not seeing yet.

Having the memory manager be a runtime variable is totally workable, but it becomes difficult if you want to support standard `std` containers, because they determine their allocator type at compile time, and it's that context switching from the value level (to create the vectors), to the type level (to determine their allocators), and back to the value level (to determine the memory manager) that I am most worried about.

Thinking about it this way, I can distinguish roughly three separate approaches:

1. Go for a mixed compile-time and run-time design, where the user needs to be very careful about modifying the global state of the library, as well as locking when running in a multi-threaded environment.
2. Go for a run-time design, where we can't really use `std` containers anymore, but everything else would work nicely.
3. Go for a compile-time design with a lot of templating, this would of course risk "over-templating" the entire library, and it would also require some global state, but in this case it would be hidden from the user entirely. As you say, this would also make it more difficult to make placement decisions at run-time.

Does that roughly match up with your thinking? I'm not trying to necessarily advocate for one approach here, but I do think it's worth exploring the possibilities. :slightly_smiling_face:

PR #6, opened 2021-02-09
Add a very basic .gitignore file

stephenswat:
This `.gitignore` file really just ignores the `build/` directory which we are building into. I don't think we need more than that yet.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,2 @@
+# Ignore the common build directory
+build/


.gitignore
```diff
@@ -0,0 +1,2 @@
+# Ignore the common build directory
+build/
```
krasznaa, line 2: Do you think this should be our recommended layout? Yes, the CI sets things up like this. But I myself prefer a proper out-of-source build.

I was expecting that you would be adding the files to be ignored in an in-source build. That would be useful. Though I have strongly opposed feelings against in-source builds as well...

krasznaa (commented):
Please rebase your branch on top of the head of the main branch. So that the CI tests could run. :wink:

.gitignore
stephenswat: Personally I have no particular preference for in-source builds (both kinds), or out-of-source builds; `git clean` takes care of any excess files I may want to get rid of. In my experience in-source builds are somewhat easier for people not experienced with CMake, but the difference is not at all huge.

paulgessinger: Outside the ATLAS context, I've never had a problem building anything with a `build` folder inside the git repo.

stephenswat:
Will make sure to rebase in the future, but is there a good reason why the CI system can't do this automatically for branches that can be auto-merged by git? Haven't seen this requirement on many other CI systems.

paulgessinger:
My personal experience is that CI auto-committing to dev branches is almost always a bad idea. You can usually do a merge update from the GitHub interface, which creates a merge commit, however. In ACTS we switched to squash-and-merge, which solves this problem.

krasznaa:
Neither GitLab-CI nor GitHub Actions run on the state of the repository as it would be after the merge. GitLab-CI may have introduced something along these lines recently, as we were requesting it for a long time in ATLAS, but so far that was not the case.

What these CIs do is that they build the code exactly as it is in your feature branch. Since the build happens already when you push to your feature branch, not just when you open a pull request.

In the Jenkins based ATLAS setup the CI builds **do** happen on top of the "merged state" of the merge request. But that's a Jenkins specific trick that we do. GitLab and GitHub don't do this automatically.

And this is why there's a big warning on PRs that our out-of-date with the base branch. To try to work around this shortcoming of the CI system...

krasznaa:
Superseeded by #12...

PR #7, opened 2021-02-09
Refactor memory types into single enum

stephenswat:
Currently, each allocator has its own enum representing a memory type, but this seems redundant and confusing, and I think we will be better off having one single definition of that enum to use throughout all of the code.

### DIFF
--- a/cuda/include/vecmem/memory/memory_types.hpp
+++ b/cuda/include/vecmem/memory/memory_types.hpp
@@ -1,0 +1,18 @@
+/** VecMem project, part of the ACTS project (R&D line)
+ *
+ * (c) 2021 CERN for the benefit of the ACTS project
+ *
+ * Mozilla Public License Version 2.0
+ */
+#pragma once
+
+namespace vecmem::memory {
+   enum class memory_type {
+      //! Memory allocated in device memory
+      DEVICE,
+      //! Pinned memory allocated on the host
+      HOST,
+      //! CUDA managed memory allocated on the device and host
+      MANAGED
+   };
+}
--- a/cuda/include/vecmem/memory/cuda/arena_memory_manager.hpp
+++ b/cuda/include/vecmem/memory/cuda/arena_memory_manager.hpp
@@ -7,6 +7,7 @@
 #pragma once
 
 // Local include(s).
+#include "vecmem/memory/memory_types.hpp"
 #include "vecmem/memory/memory_manager_interface.hpp"
 
 // System include(s).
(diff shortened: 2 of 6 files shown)


cuda/include/vecmem/memory/memory_types.hpp
```diff
@@ -0,0 +1,18 @@
+/** VecMem project, part of the ACTS project (R&D line)
+ *
+ * (c) 2021 CERN for the benefit of the ACTS project
+ *
+ * Mozilla Public License Version 2.0
+ */
+#pragma once
+
+namespace vecmem::memory {
+   enum class memory_type {
```
krasznaa, line 10: I'm on board with having a common definition for the CUDA memory managers. But we should not make this global.

oneAPI's USM for instance calls "the 3rd kind" of memory "shared" memory. Not "managed" memory. And HIP does an entirely different thing with "registering host memory to be accessible from the device".

https://rocmdocs.amd.com/en/latest/ROCm_API_References/HIP_API/Memory-Management.html#hiphostregister

So the easiest would be to avoid handling all languages in one strict way.

I propose making this:

```c++
namespace vecmem::cuda {
   enum class memory_type {
      ...
   };
}
```

, and putting it in `vecmem/memory/cuda/memory_type.hpp`. (I.e. let's use the name of the enum as the file name...)

krasznaa (changes requested):
As in all other cases, please rebase on top of the HEAD of the main branch.

cuda/include/vecmem/memory/memory_types.hpp
paulgessinger: So there's now, let's say, abstraction that could be used to define an enum of this kind?

krasznaa: I'm not sure... I both want to stay away from over-generalising the code, limiting our options by that. At the same time, as I wrote in https://github.com/acts-project/vecmem/pull/5#pullrequestreview-586842764, a single definition would make things a little easier.

I'm still pondering what we should be doing...

stephenswat: I'm pretty conflicted about this to be honest, same as with #5 I worry that if we don't make some steps to generalize things we are going to end up with an unmaintainable amount of boilerplate. Already with the SYCL implementation I think there is a lot of stuff there that could be de-duplicated and abstracted away in a way very similar to the CUDA implementation.

krasznaa:
:confused: Why is the CI not executing even now, after I updated your branch to the latest version?

You may need to enable GitHub actions on your fork. At least that's the only thing I can think of. Although at any rate, I think you should just switch to creating feature branches in this main repository. Then the CI should work properly for sure...

paulgessinger:
Not sure if that's the problem, but you might have to set the workflow jobs to trigger on pull requests as well.

stephenswat:
Seems to be working now! :smiley:

PR #8, opened 2021-02-09
SYCL Code Addition, main branch (2021.02.09.)

krasznaa:
Added some simple code that would interact with [SYCL](https://www.khronos.org/sycl/). Only added the simplest memory manager for it as a start, and only made sure so far that the simplest operations wouldn't make the code crash.

@stephenswat, we should design the memory managers such that this would fit in with them as well... I'll also try to add a library for [HIP](https://rocmdocs.amd.com/en/latest/index.html) tomorrow...


tests/sycl/test_sycl_allocators.cpp
```diff
@@ -0,0 +1,78 @@
...
+ * Mozilla Public License Version 2.0
+ */
+
+// Local include(s).
+#include "vecmem/allocators/allocator.hpp"
+#include "vecmem/memory/memory_manager.hpp"
+#include "vecmem/memory/sycl/direct_memory_manager.hpp"
+
+// System include(s).
+#undef NDEBUG
```
paulgessinger, line 14: This seems odd. I guess it's not dangerous in a test, but what's the reason for this?

paulgessinger: Got it, so that `assert` works even with a release build.

krasznaa: We do this in quite a few places in the Athena core software...

I didn't want to start with something like gtest right away. But that's absolutely in the cards.

paulgessinger: Cool.

paulgessinger: I personally define a few ASSERT macros for this situation, but fair enough.

krasznaa: The reason that I came to like using the system `assert(...)` function is that it very nicely works in CUDA device code as well. (All in all using `assert(...)` in CUDA code proved to be a godsend while I was working on the seedfinder code.)

paulgessinger: That's a good point. I guess the structure and reporting of something like GTest can still provide a net benefit, at least on the host.

krasznaa:
Hmm... As it turns out,

```c++
#undef NDEBUG
```

can confuse the oneAPI compiler after all. :frowning: At least when its used with device code, in a Release build. I have to admit, I didn't try doing this before...

PR #11, opened 2021-02-12
ROCm/HIP Code Addition, main branch (2021.02.12.)

krasznaa:
Copied the HIP handling code from https://gitlab.cern.ch/atlas/atlasexternals/-/tree/master/Build/AtlasHIP, making some fixes in it along the way that will need to be ported back there as well.

Then I created a `vecmem::hip` library in the same way as all the other libraries. Plus added the exact same tests for the new library that exist for the other libraries as well.

@stephenswat, this still uses the "old" style of memory management. It will also need to be updated to the new style that we discussed this week. :wink:

Note that RPATH handling doesn't work quite correctly in this setup yet. :frowning: I didn't find the correct incantation just yet for convincing CMake that the "HIP shared libraries and executables" can use the RPATH mechanism... Will probably need to get help from CMake experts on that... (This means that you need to set `$LD_LIBRARY_PATH` by hand to run the `vecmem_test_hip_containers` test at the moment.)


.vscode/settings.json
```diff
@@ -0,0 +1,5 @@
+{
+   "files.associations": {
+      "*.hip": "cpp"
+   }
+}
```
paulgessinger, line 5: Just a comment: I'm always on the fence about committing these.

krasznaa: I actually created this file by hand, it was **not** auto-created.

This is needed to have VSCode recognise the .hip extension as a C++ file. I think anybody working on this project with VSCode will want to have this set...

hip/CMakeLists.txt
```diff
@@ -0,0 +1,22 @@
...
+# (c) 2021 CERN for the benefit of the ACTS project
+#
+# Mozilla Public License Version 2.0
+
+# Enable HIP as a language.
+enable_language( HIP )
+
+# Project include(s).
+include(
+   "${CMAKE_CURRENT_SOURCE_DIR}/../cmake/vecmem-compiler-options-hip.cmake" )
```
paulgessinger, line 12: Is this preferable over modifying the cmake include search path?

krasznaa: I was pondering about that myself as well... Maybe not...

paulgessinger:
I think with a [`.gitattributes` file](https://github.com/github/linguist#using-gitattributes) you should be able to convince github/linguist to highlight `.hip` files as C++.

.vscode/settings.json
paulgessinger: That's correct. But that makes it much more difficult to have differing settings from you. VSCode puts all kinds of stuff in there, and most of them you probably don't want to commit.

krasznaa: My experience is that if the file exists already, then VSCode is much more measured in this. It does try to create other files in `.vscode` automatically. But I don't think it would ever modify this file itself without the user asking for it.

I see your point with this, but until the project gets more complicated, when the users may indeed want to have different settings in their own clones, I thought it would be safe enough to add this simple setting to the repository.

krasznaa:
> I think with a [`.gitattributes` file](https://github.com/github/linguist#using-gitattributes) you should be able to convince github/linguist to highlight `.hip` files as C++.

Unfortunately it didn't have an effect in the PR's diff-view. But maybe after it gets merged in... :thinking:

.vscode/settings.json
krasznaa: So, how strongly do you feel about this? As from my side I'm ready with this PR at this point.

paulgessinger: Hadn't seen this, not strongly anyway.

PR #12, opened 2021-02-12
VSCode DevContainer Setup, main branch (2021.02.12.)

krasznaa:
@paulgessinger, since you didn't like me adding a `.vscode/settings.json` file to the repository before, I really wonder what you'll think about this one... :stuck_out_tongue:

I've been developing the code largely in the environment that this configuration sets up. Which allows VSCode to find the ROCm/HIP and SYCL headers correctly, leading to a much nicer code development experience. Like:

vecmem_devcontainer

@stephenswat, in the end I just followed your earlier suggestion, and added `build/` as an ignored directory in the source. Since with this DevContainer setup it's a lot easier to place the build directory inside of the source directory...

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,11 @@
+# VecMem project, part of the ACTS project (R&D line)
+#
+# (c) 2021 CERN for the benefit of the ACTS project
+#
+# Mozilla Public License Version 2.0
+
+# Base the development image on an ATLAS specific image for the time being.
+FROM gitlab-registry.cern.ch/akraszna/atlas-gpu-devel-env:centos7-1-3-1
+
+# Set up the command to run when starting the image.
+CMD /bin/bash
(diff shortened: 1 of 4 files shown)


paulgessinger:
This is good stuff. If been remoting into the Threadripper machine using VSCode for a bit, it's a pretty good experience. Other than letting CMake Tools configure Intellisense I haven't gotten the latter to work properly though, and it still randomly stops working once in a while. But it's pretty good overall.

paulgessinger:
Just to make clear that I'm not just being grumpy about the `settings.json` file: this is what my file looks like from just doing that, and letting the CMake extension configure Intellisense:

image

The thing is, I can configure these things at three levels: User, Machine, Workspace. If you commit the `settings.json` file, realistically, I can only do it at the former two, which for obvious reasons is unworkable.

paulgessinger:
The devcontainer config I'm more neutral about. This is more or less project specific and makes sense to be shared between developers, as opposed to almost everything going into `settings.json`.

PR #14, opened 2021-02-15
Use modern C++ memory resource semantics

stephenswat:
This is a draft merge request to implement some of the architectural changes which I discussed with @krasznaa last week. The gist of it is that we use, as much as possible, the memory semantics of modern C++, specifically the C++17 standard, including `std::memory_resource`. Currently, only the core module and the CUDA module have been altered to use the new architecture, but obviously we plan to convert the HIP and SYCL modules as well.

The primary motivation of this merge request was to step away from the shared global state architecture that was in place before, because we were worried about maintainability implications of such an architecture. This new implementation is completely run-time-based, in vector storage location as well as vector memory type. For this, we take a note from the `std::pmr` library, especially `std::pmr::polymorphic_allocator` which powers `std::pmr::vector`. We cannot use these classes directly because we need some additional information about our memory resources which are not exposed by default, but our code is heavily inspired by the polymorphic STL code.

Our new code uses a stackable allocation scheme design, just like the STL implentation. In fact, STL allocation schemes should be trivially compatible with our code. This means that we distinguish pool resources and terminal resources, which respectively distribute memory from an upstream resource or interface directly with some memory manager like `malloc` or `cudaMalloc`.

#### Concerns

Like the previous allocator code, we rely on overloads of `construct` and `destroy` to ensure that the CPU does not try to write to device-only memory, thus triggering a segmentation fault. However, the existence of `construct` and `destroy` in the allocator is deprecated in C++17, and removed entirely in C++20. This could cause trouble with our code in the future. We might need to find a different approach for this, for example using `std::allocator_traits`.

#### Todo

The following tasks need to be completed before this merge request can be merged in:

- [ ] Add implementation for HIP
- [ ] Add implementation for SYCL
- [ ] Document the code


paulgessinger:
Just trying to follow along here. *Stackable* here means I can use multiple allocators, one which for example allocates a block of memory, and another one on top which allocates within the block for individual structs etc. Is that correct?

Other than that, you conform to the `std::pmr` interface for the allocators? And these allocators are still compatible with plain old `std::vector`?

cmake/vecmem-compiler-options-cuda.cmake
```diff
@@ -11,7 +11,7 @@ include_guard( GLOBAL )
 include( vecmem-functions )
 
 # Set up the used C++ standard(s).
-set( CMAKE_CUDA_STANDARD 14 CACHE STRING "The (CUDA) C++ standard to use" )
+set( CMAKE_CUDA_STANDARD 17 CACHE STRING "The (CUDA) C++ standard to use" )
```
krasznaa, line 14: I hope this is not needed, otherwise we'll be in trouble.

To be able to work with SYCL on NVidia devices, we need to keep using CUDA 10.X. Which does not allow for C\+\+17. :frowning:

stephenswat:
@paulgessinger, that's the core idea of it, yes! This is also how the standard library does it, for example the [`std::pmr::synchronized_pool_resource`](https://en.cppreference.com/w/cpp/memory/synchronized_pool_resource): it doesn't do any allocation as such by itself, but it relies on an _upstream_ memory resource which either does the allocation, or relies on another resource further upstream.

As for compatibility with `std::vector`, we define vectors in this new architecture as templates of `std::vector`, yes. However, the problem with the allocator template type is that two vectors of the same value type but with different allocators do not type check. That's the problem that `std::pmr::vector` tries to solve using its polymorphic allocator. We're far from a perfect solution here, and even the standard library doesn't have one. But at least we can consistently use the `vecmem::memory::polymorphic_allocator` here, and as long as we do that our vectors should type check. :slightly_smiling_face:

core/include/vecmem/memory/allocator.hpp
```diff
@@ -0,0 +1,50 @@
...
+        void deallocate(T * ptr, std::size_t size) {
+            alloc.deallocate(ptr, size);
+        };
+
+        template<typename U, typename... Args>
+        void construct(U* ptr, Args&&... args) {
+            if (is_host_accessible()) {
+                alloc.construct(ptr, std::forward<Args>(args)...);
+            }
+        };
```
krasznaa, line 33: We discussed this quickly last week, but amongst all the things it's inevitable to forget about a few.

I don't think we need to bother with non-host-accessible memory types in our allocators. As we discussed at that point, using an STL allocator on top of a memory block that's not host-accessible is just not making a lot of sense. (Now that I played with it for a bit.) So I think we can just use [std::pmr::polymorphic_allocator](https://en.cppreference.com/w/cpp/memory/polymorphic_allocator) as is...

core/include/vecmem/memory/resources/base_resource.hpp
```diff
@@ -0,0 +1,10 @@
+#pragma once
+
+#include <memory_resource>
+
+namespace vecmem::memory::resources {
+    class base_resource : public std::pmr::memory_resource {
+    public:
+        virtual bool is_host_accessible() const = 0;
+    };
```
krasznaa, line 9: Note that while we don't need this information in our allocators, I think it still makes sense to flag which resources produce host-accessible memory. Since we'll be using these resources not only in the STL-like allocators.

core/src/memory/resources/terminal/malloc_resource.cpp
```diff
@@ -0,0 +1,31 @@
...
+        std::size_t,
+        std::size_t
+    ) {
+        free(p);
+    }
+
+    bool malloc_resource::do_is_equal(
+        const std::pmr::memory_resource & other
+    ) const noexcept {
+        return false;
```
krasznaa, line 25: Why would two malloc resources never be equal to each other? I would've thought they would always be equal... :confused:

cuda/include/vecmem/memory/cuda/resources/resources.hpp
```diff
@@ -0,0 +1,9 @@
+#include <memory_resource>
+
+#include "vecmem/memory/resources/base_resource.hpp"
+
+namespace vecmem::memory::resources {
+    base_resource * get_terminal_cuda_device_resource() noexcept;
+    base_resource * get_terminal_cuda_host_resource() noexcept;
+    base_resource * get_terminal_cuda_managed_resource() noexcept;
```
krasznaa, line 8: I don't like this way of providing the resources to the users. :frowning:

The resources, eventually, will need to be tied to particular devices. So that in a system that has multiple GPUs, we could specify which resource handles which of them.

So I would just forego any centrally provided resources, and make it a task for the final application to create the resource objects for the GPUs / other resources that they want to use.

cuda/src/memory/cuda/resources/terminal/device_resource.cpp
```diff
@@ -0,0 +1,35 @@
...
+        std::size_t,
+        std::size_t
+    ) {
+        cudaFree(p);
+    }
+
+    bool cuda_device_resource::do_is_equal(
+        const std::pmr::memory_resource & other
+    ) const noexcept {
+        return false;
```
krasznaa, line 29: I would think that logically, whenever two resource objects are set up to talk to the same GPU, they should be considered equal, no?

cuda/src/memory/cuda/resources/terminal/device_resource.cpp
```diff
@@ -0,0 +1,35 @@
...
+        cudaMalloc(&res, bytes);
+        return res;
+    }
+
+    void cuda_device_resource::do_deallocate(
+        void * p,
+        std::size_t,
+        std::size_t
+    ) {
+        cudaFree(p);
```
krasznaa, line 23: A couple of points:
  - I would call the class `vecmem::cuda::device_resource`. Let's not get too crazy with the namespaces. (The whole project is about memory management. I don't think we should add a `memory` sub-namespace at all.)
  - With that class name this file should be called `device_resource.cu`. So that CMake would recognise it as a CUDA file. (And also, pay attention to naming the files the same as the class that they hold!)
  - Please use `VECMEM_CUDA_ERROR_CHECK(...)` on all of the CUDA calls.

krasznaa (changes requested):
This is just a first round. I'll have more comments later on. :stuck_out_tongue:

cmake/vecmem-compiler-options-cuda.cmake
stephenswat: That's a little awkward. Much of the code relies on C++17 features, so if we are to use C++17 for all of the standard C++ compilation and C++14 for the CUDA code, we would need to define a really clear split between the headers of the two... :disappointed:

core/include/vecmem/memory/allocator.hpp
stephenswat: That would definitely make life a lot easier, as you say we could just use the standard `std::pmr::polymorphic_allocator`.

cuda/include/vecmem/memory/cuda/resources/resources.hpp
stephenswat: That's a fair comment. This design mirrors roughly what the standard library does [(`std::pmr::get_default_resource`)](https://en.cppreference.com/w/cpp/memory/get_default_resource), but that doesn't mean it's the best design. Putting the onus of creating the resources on the user is fine with me.

cuda/src/memory/cuda/resources/terminal/device_resource.cpp
stephenswat: This was mostly a placeholder decision until we fleshed out the architecture for dealing with multiple devices a little more, but yes, in principle all resources dedicated to the same GPU can be considered equal (for now).

cmake/vecmem-compiler-options-cuda.cmake
krasznaa: :thinking: It should not be too bad. `vecmem::cuda::device_resource` (and friends) will need to be a C\+\+ class, with a `.cpp` implementation file. Then internally the CUDA library can create helper functions for allocating memory on a GPU. These helper functions can be implemented in `.cu` files, compiled with C\+\+14. It's really just one degree of separation that we need.

cuda/src/memory/cuda/resources/terminal/device_resource.cpp
stephenswat: Are you sure you want this to be compiled with the CUDA compiler? There is no device code here, so I don't personally see the need for it.

core/include/vecmem/memory/allocator.hpp
stephenswat: Agreed. In that case we can simply use `std::pmr::polymorphic_allocator` directly. However, I am struggling to see how the CUDA example code would then look like.

cuda/src/memory/cuda/resources/terminal/device_resource.cpp
krasznaa: It makes the CMake configuration way easier if we use the `.cu` file extension. As CMake then automatically makes the code see `<cuda_runtime.h>`, and link the code against the CUDA runtime library. Otherwise we need to do all of those by hand.

So while this is a slight overhead, I just got into the habit of putting all of my CUDA code (even the code that's only running on the host) into `.cu` files.

Also, I've set up the CMake configuration to do the same for `.hip` and `.sycl` files. There I also want to use the special extensions for any code that interacts with the runtime of those languages...

core/include/vecmem/memory/allocator.hpp
krasznaa: You should just remove all of the `std::vector` tests with non-host-accessible memory. Those are really useless tests...

The unit test should just make sure that the memory resources can provide non-null pointers for device-only memory blocks.

stephenswat:
On a related note, `std::pmr::memory_resource` is not properly supported in GCC until 9.1.0. Do we have any compiler targets to work around? I see the CI is using 7.5.0; is that something we want to keep that way?

krasznaa:
> On a related note, `std::pmr::memory_resource` is not properly supported in GCC until 9.1.0. Do we have any compiler targets to work around? I see the CI is using 7.5.0; is that something we want to keep that way?

Yeah, I saw. Unfortunately I think we won't be able to use these new types directly. I think we'll have to settle for re-implementing them in our code.

Not supporting GCC 7-8 with our code I think would be a bad idea. Ubuntu 18.04, which the GCC 7 build is based on, will remain a "standard platform" for some time to come. I think this project should support is natively...

krasznaa:
Though maybe the situation is not quite that bad... We could put something like this into the code:

```c++
#if SOMETHING
#  include <memory_resource>
#else
#  include <experimental/memory_resource>
#
```

With some CMake magic we could make this robust enough... (I just checked, GCC 7 holds this header under `<experimental/memory_resource>`...)

cuda/src/memory/cuda/resources/terminal/device_resource.cpp
```diff
@@ -0,0 +1,36 @@
+#include <memory>
+
+#include <cuda_runtime_api.h>
+
+#include "vecmem/memory/resources/memory_resource.hpp"
+#include "vecmem/memory/cuda/resources/terminal/device_resource.hpp"
+#include "vecmem/utils/cuda_error_handling.hpp"
```
krasznaa, line 7: We don't have any coding guidelines for the project yet, but let's just adopt the ATLAS one. I.e. include stuff in the following order:
  - Project headers;
  - Headers from external projects;
  - System headers.

So, going from the most specific to the least specific. This tends to be a good way of detecting problems in one's header files.

tests/cuda/test_cuda_containers.cpp
```diff
@@ -0,0 +1,61 @@
+/** VecMem project, part of the ACTS project (R&D line)
```
krasznaa, line 1: Splitting this executable into 2 source files seems to just be overcomplicating things. To do this properly you would also need to introduce a 3rd file, `test_cuda_containers_kernels.hpp`. But why do the split in the first place? What was the problem with having all the test code compiled by `nvcc`? :confused:

krasznaa (commented):
We should try to talk a bit in the afternoon...

cuda/include/vecmem/utils/cuda_error_handling.hpp
```diff
@@ -19,9 +20,6 @@
                                             __LINE__ );                        \
       }                                                                        \
    } while( false )
-#else
-#   define VECMEM_CUDA_ERROR_CHECK( EXP ) do {} while( false )
-#endif // __CUDACC__
```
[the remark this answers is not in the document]
stephenswat: This is a workaround for the fact that the CI compiler is configured to throw compilation errors for unused variables, and silently discarding variables like this macro does is a real compilation trap combined with that behavior. Of course a simple `(void) EXP` would suffice also, but is there a good reason why this should only be defined for `nvcc`? Nothing in here is dependent on device code and gcc can compile it just fine, so I think it leads to much more predictable compilation if this macro is compiler-independent. But perhaps I'm missing something.

core/include/vecmem/utils/types.hpp
```diff
@@ -6,13 +6,6 @@
  */
 #pragma once
 
-// CUDA include(s).
-#ifdef __CUDACC__
-#   include <cuda_runtime.h>
-#else
-typedef int cudaError_t; ///< Dummy type for @c cudaError_t in C++ code
-#endif // __CUDACC__
-
```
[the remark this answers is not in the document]
stephenswat: I had to rip this out temporarily because it was causing a ton of duplicate definition errors. Of course we should put it back in when we're in a more stable position (if it is still necessary)!

cuda/CMakeLists.txt
```diff
@@ -10,14 +10,28 @@ enable_language( CUDA )
...
+   "src/memory/cuda/resources/terminal/host_resource.cpp"
+   "include/vecmem/memory/cuda/resources/terminal/managed_resource.hpp"
+   "src/memory/cuda/resources/terminal/managed_resource.cpp"
+   "include/vecmem/memory/cuda/resources/resources.hpp"
+   "src/memory/cuda/resources/resources.cpp"
    # Utilities.
    "include/vecmem/utils/cuda_error_handling.hpp"
    "src/utils/cuda_error_handling.cu" )
+
+target_link_libraries( vecmem_cuda PUBLIC CUDA::cudart )
```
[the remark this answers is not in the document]
stephenswat: That's fair, in that case we link `vecmem_cuda` against the runtime privately and then link the CUDA tests against it as well.

cuda/src/memory/cuda/resources/terminal/device_resource.cpp
```diff
@@ -0,0 +1,36 @@
+#include <memory>
+
+#include <cuda_runtime_api.h>
```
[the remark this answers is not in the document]
stephenswat: Principle of least action, if `cuda_runtime_api.h` has everything we need (which I believe it does) then I would rather import that than the heavier `cuda_runtime.h`, since both are designed to be outward-facing headers for user code. However for the sake of consistency I am more than happy to use `cuda_runtime.h`! :slightly_smiling_face:

tests/cuda/test_cuda_containers.cpp
stephenswat: This is unfortunately related to the C++14 requirement of the CUDA code. As we discussed earlier, `nvcc` won't compile many of the C++17 features we use here, so we need to compartmentalize. It seemed easier to put the CUDA code into its own little pocket like this than to fence off all the rest of the code, just in terms of implementation effort. That would also impact the code for people not using the CUDA plugin at all! I think it's a good idea to minimize the use of `nvcc` as much as possible, don't you agree?

cuda/CMakeLists.txt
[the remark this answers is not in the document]
krasznaa: In this moment the library **does** publicly depend on CUDA. So the CMake configuration is correct. My point is that I want all public headers of the library to be usable without the CUDA headers.

tests/cuda/test_cuda_containers.cpp
krasznaa: I see.

In that case we need a header file as well. :stuck_out_tongue:

PR #15, opened 2021-02-18
Add basic definitions for memory resource architecture

stephenswat:
I've decided to split up #14 into three separate merge requests for the sake of clarity. This is the first of three, upon which the others will build.

This commit lays the groundwork for the rest of the migration to a memory resource architecture. That is to say, it defines some of the common definitions and infrastructure on which we will build the memory resources for CPU code, CUDA, and other paradigms. More specifically it imports the required `std::pmr` types in a compiler agnostic way, and it defines our custom abstract subclass of `std::pmr::memory_resource`. It also defines a useful alias for creating vectors.

### DIFF
--- a/core/include/vecmem/memory/resources/memory_resource.hpp
+++ b/core/include/vecmem/memory/resources/memory_resource.hpp
@@ -1,0 +1,37 @@
+/** VecMem project, part of the ACTS project (R&D line)
+ *
+ * (c) 2021 CERN for the benefit of the ACTS project
+ *
+ * Mozilla Public License Version 2.0
+ */
+
+#pragma once
+
+/*
+ * The purpose of this file is to provide uniform access (on a source-code
+ * level) to the memory_resource and polymorphic_allocator types from the
+ * standard library. These are either in the std::pmr namespace or in the
+ * std::experimental::pmr namespace depending on the GCC version used, so we try
+ * to unify them by aliassing depending on the compiler feature flags.
+ */
+#if __has_include(<memory_resource>)
+#include <memory_resource>
+
+namespace vecmem {
+    using memory_resource = std::pmr::memory_resource;
+
+    template<typename T>
+    using polymorphic_allocator = std::pmr::polymorphic_allocator<T>;
+}
+#elif __has_include(<experimental/memory_resource>)
+#include <experimental/memory_resource>
+
+namespace vecmem {
+    using memory_resource = std::experimental::pmr::memory_resource;
+
+    template<typename T>
+    using polymorphic_allocator = std::experimental::pmr::polymorphic_allocator<T>;
+}
+#else
+#error "vecmem requires C++17 LFTS V1 (P0220R1) component memory_resource!"
+#endif
--- a/core/CMakeLists.txt
+++ b/core/CMakeLists.txt
@@ -13,11 +13,13 @@ vecmem_add_library( vecmem_core core SHARED
    "include/vecmem/containers/device_vector.hpp"
    "include/vecmem/containers/host_vector.hpp"
    "include/vecmem/containers/static_vector.hpp"
+   "include/vecmem/containers/vector.hpp"
    # Memory management.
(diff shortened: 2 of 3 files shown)


core/include/vecmem/memory/resources/base_resource.hpp
```diff
@@ -0,0 +1,35 @@
...
+     * @todo It may be difficult to determine the host accessibility for
+     * non-terminal memory resources, and we need a good solution for this.
+     */
+    class base_resource : public memory_resource {
+    public:
+        /**
+         * @brief Check whether this memory resource can be accessed by the host
+         * device.
+         */
+        virtual bool is_host_accessible() const = 0;
```
krasznaa, line 33: I guess we're keeping this function after all... :confused:

I just still don't see where we'll be able to make use of it... Since `std::pmr::polymorphic_allocator` doesn't care about it. So even if we pass a "wrong type" of memory resource to a `vecmem::vector`, that will still not result in any clean failure.

So I would still just get rid of this class, and implement our resources directly on top of `vecmem::resource::memory_resource`. Which at that point I would really rather just call `vecmem::memory_resource`...

core/include/vecmem/memory/resources/memory_resource.hpp
```diff
@@ -0,0 +1,35 @@
...
+#if __has_include(<memory_resource>)
+#include <memory_resource>
+
+namespace vecmem::resource {
+    using memory_resource = std::pmr::memory_resource;
+
+    template<typename T>
+    using polymorphic_allocator = std::pmr::polymorphic_allocator<T>;
+}
+#else
```
krasznaa, line 26: I would rather do:

```c++
#elif __has_include(<experimental/memory_resource>)
```

And then at the end of the block you can do:

```c++
#else
#   error "VecMem does not support the compiler in use!"
#endif
```

core/include/vecmem/memory/resources/memory_resource.hpp
```diff
@@ -0,0 +1,35 @@
...
+ */
+#if __has_include(<memory_resource>)
+#include <memory_resource>
+
+namespace vecmem::resource {
+    using memory_resource = std::pmr::memory_resource;
+
+    template<typename T>
+    using polymorphic_allocator = std::pmr::polymorphic_allocator<T>;
+}
```
krasznaa, line 25: I would prefer the simpler `vecmem::memory_resource` and `vecmem::polymorphic_allocator` names. Since eventually these will be called `std::memory_resource` and `std::polymorphic_allocator` (at least I'm pretty sure that they will be...), we shouldn't try to make the name more elaborate than it needs to be...

krasznaa (changes requested):
On top of all of these I would also just prefer to move `vecmem/memory/resources/memory_resource.hpp` to `vecmem/memory/memory_resource.hpp`.

I know, I might be a bit overly opinionated on these things, but I'm really just trying not to over-complicate our naming convention out of the gate...

core/include/vecmem/memory/resources/base_resource.hpp
stephenswat: I thought we discussed in our last meeting that we should keep this, but I guess I was wrong. I'll get rid of it!

krasznaa: My thinking was just that we would for sure introduce memory resource types that would not be host-accessible. We would just never use those with `vecmem::vector`, only with some other, simpler constructs. (Like [Acts::Cuda::device_array](https://github.com/acts-project/acts/blob/master/Plugins/Cuda/include/Acts/Plugins/Cuda/Utilities/Arrays.hpp).)

So while the ability to pair the wrong resource type with a given higher-level type will be ever-present, I just don't know how we could avoid this. At least in `vecmem::vector`, where it really counts...

krasznaa (approved):
The header is still called `vecmem/memory/resources/memory_resource.hpp`, but let's just go with that for now. :wink:

PR #16, opened 2021-02-18
Separate CUDA container test kernel from its driver

stephenswat:
As it is right now, the CUDA container test contains both device code and host driver code in the same translation unit, which forces us to compile both with `nvcc`. This is normally not a problem, but as discussed in #14 we need to compile our C++ code with the C++17 standard while we can only compile our CUDA code with the C++14 standard. It is therefore necessary to enforce a separation between these two contexts, and thus we split it into two translation units. The C++ TU will handle the driving of the kernel and interact with the C++17-based architecture of the `vecmem` library, while the kernel itself will be compiled with C++14 and be agnostic of whatever C++ standard is being used outside.

### DIFF
--- a/tests/cuda/CMakeLists.txt
+++ b/tests/cuda/CMakeLists.txt
@@ -15,5 +15,6 @@ vecmem_add_test( cuda_allocators test_cuda_allocators.cpp
    LINK_LIBRARIES vecmem::core vecmem::cuda )
 
 # Tests for the core library's device vector types, using the CUDA backend.
-vecmem_add_test( cuda_containers test_cuda_containers.cu
+vecmem_add_test( cuda_containers test_cuda_containers.cpp
+   test_cuda_containers_kernels.cu test_cuda_containers_kernels.cuh
    LINK_LIBRARIES vecmem::core vecmem::cuda )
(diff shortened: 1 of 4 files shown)


tests/cuda/CMakeLists.txt
```diff
@@ -15,5 +15,6 @@ vecmem_add_test( cuda_allocators test_cuda_allocators.cpp
    LINK_LIBRARIES vecmem::core vecmem::cuda )
 
 # Tests for the core library's device vector types, using the CUDA backend.
-vecmem_add_test( cuda_containers test_cuda_containers.cu
+vecmem_add_test( cuda_containers test_cuda_containers.cpp
+   test_cuda_containers_kernels.cu
```
krasznaa, line 19: Just one comment on this one. :stuck_out_tongue: Please add `test_cuda_Containers_kernels.cuh` to the list here. That will allow some IDEs (like Xcode and VisualStudio) to list the files belonging to the executable correctly.

I'll still need to add some CMake code to make those IDEs like our setup, but at least I won't have to do this later on then. :smile:

krasznaa (approved):
Looks good!

PR #17, opened 2021-02-18
Migrate core CPU code to use memory resources

stephenswat:
Now that the groundwork is laid, we can start to convert the different modules from the old memory management to the new memory resources. This merge request does exactly that for the `core` CPU code: it adds a single new memory resource, wrapping `malloc`, and it updates the tests to work with the new architecture.

### DIFF
--- a/core/CMakeLists.txt
+++ b/core/CMakeLists.txt
@@ -21,5 +21,7 @@ vecmem_add_library( vecmem_core core SHARED
    "include/vecmem/memory/host_memory_manager.hpp"
    "src/memory/host_memory_manager.cpp"
    "include/vecmem/memory/resources/memory_resource.hpp"
+   "src/memory/host_memory_resource.cpp"
+   "include/vecmem/memory/host_memory_resource.hpp"
    # Utilities.
    "include/vecmem/utils/types.hpp" )
--- a/core/include/vecmem/memory/host_memory_resource.hpp
+++ b/core/include/vecmem/memory/host_memory_resource.hpp
@@ -1,0 +1,41 @@
+/**
+ * VecMem project, part of the ACTS project (R&D line)
+ *
+ * (c) 2021 CERN for the benefit of the ACTS project
+ *
+ * Mozilla Public License Version 2.0
+ */
+
+#pragma once
+
+#include "vecmem/memory/resources/memory_resource.hpp"
+
+#include <memory>
+#include <cstddef>
+
+namespace vecmem {
+    /**
+     * @brief Memory resource which wraps the malloc standard library call.
+     *
+     * This is probably the simplest memory resource you can possibly write. It
+     * is a terminal resource which does nothing but wrap malloc and free. It
+     * is state-free (on the relevant levels of abstraction).
+     */
+    class host_memory_resource : public vecmem::memory_resource {
+    private:
+        virtual void * do_allocate(
+            std::size_t,
+            std::size_t
+        ) override;
+
+        virtual void do_deallocate(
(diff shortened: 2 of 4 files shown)


core/include/vecmem/memory/resources/terminal/malloc_resource.hpp
```diff
@@ -0,0 +1,33 @@
+#pragma once
```
krasznaa, line 1: Please add the "Acts header".

core/include/vecmem/memory/resources/terminal/malloc_resource.hpp
```diff
@@ -0,0 +1,33 @@
...
+
+#include "vecmem/memory/resources/memory_resource.hpp"
+
+#include <memory>
+#include <cstdlib>
```
krasznaa, line 6: I believe you should only use `<cstddef>` here. Just to ensure that `std::size_t` is defined. Even though `vecmem::memory_resource` would ensure even that... So you may as well just not add any std includes here.

core/include/vecmem/memory/resources/terminal/malloc_resource.hpp
```diff
@@ -0,0 +1,33 @@
...
+
+namespace vecmem::resource {
```
krasznaa, line 8: I would not use the `resource` sub-namespace here. As we talked about it before, I foresee calling the CUDA resources `vecmem::cuda::device_memory`. Possibly `vecmem::cuda::device_memory_resource`. I don't want to call it `vecmem::cuda::resource::device_resource`...

So with that in mind I myself would call this class `vecmem::host_memory`... (With CUDA having a `vecmem::cuda::host_memory` later on.)

core/src/memory/resources/terminal/malloc_resource.cpp
```diff
@@ -0,0 +1,33 @@
...
+    bool malloc_resource::do_is_equal(
+        const memory_resource &
+    ) const noexcept {
+        /*
+         * All malloc resources are equal to each other, because they have no
+         * internal state. Of course they have a shared underlying state in the
+         * form of the underlying C library memory manager, but that is not
+         * relevant for us.
+         */
+        return true;
```
krasznaa, line 31: :smile: Now that I pondered about this a bit more... I think this should be implemented as:

```c++
bool host_memory::do_is_equal( const memory_resource& other ) const noexcept {
   const host_memory* otherPtr = dynamic_cast< const host_memory* >( &other );
   return ( otherPtr != nullptr );
}
```

Since STL may very well be comparing this object instance to some GPU memory resource...

The GPU resources would then not only make sure that the "other" object is the same type as they are but that it is actually also talking to the same GPU as they are.

core/CMakeLists.txt
```diff
@@ -21,5 +21,7 @@ vecmem_add_library( vecmem_core core SHARED
    "include/vecmem/memory/host_memory_manager.hpp"
    "src/memory/host_memory_manager.cpp"
    "include/vecmem/memory/resources/memory_resource.hpp"
+   "src/memory/resources/terminal/malloc_resource.cpp"
+   "include/vecmem/memory/resources/terminal/malloc_resource.hpp"
```
krasznaa, line 25: I would rather call these `include/vecmem/memory/host_memory.hpp` and `src/memory/host_memory.cpp`...

krasznaa (commented):
I'm still mostly just harping on about the naming...

krasznaa (approved):
Looks good!

PR #18, opened 2021-02-18
Link against and include the CUDA runtime library for C++ code

stephenswat:
As we discussed several times before, compiling C++14 and C++17 code alongside each other is a bad idea, and we'd be best off minimizing the use of `nvcc` so we can rely on another compiler to compile the standard C++ code. However, we were previously relying a lot on `nvcc`'s ability to automatically include and link the CUDA runtime, which we now lose. This merge request remedies that situation by properly linking against and including the CUDA runtime in C++ host code where possible.

### DIFF
--- a/cuda/CMakeLists.txt
+++ b/cuda/CMakeLists.txt
@@ -10,6 +10,8 @@ enable_language( CUDA )
 # Project include(s).
 include( vecmem-compiler-options-cuda )
 
+find_package( CUDAToolkit REQUIRED )
+
 # Set up the build of the VecMem CUDA library.
 vecmem_add_library( vecmem_cuda cuda SHARED
    # Memory management.
@@ -20,4 +22,8 @@ vecmem_add_library( vecmem_cuda cuda SHARED
    # Utilities.
    "include/vecmem/utils/cuda_error_handling.hpp"
    "src/utils/cuda_error_handling.cu" )
-target_link_libraries( vecmem_cuda PUBLIC vecmem::core )
+
+target_link_libraries( vecmem_cuda
+   PUBLIC vecmem::core
+   PRIVATE CUDA::cudart
+)
--- a/cuda/CMakeLists.txt
+++ b/cuda/CMakeLists.txt
@@ -20,8 +20,8 @@ vecmem_add_library( vecmem_cuda cuda SHARED
    "include/vecmem/memory/cuda/direct_memory_manager.hpp"
    "src/memory/cuda/direct_memory_manager.cu"
    # Utilities.
-   "include/vecmem/utils/cuda_error_handling.hpp"
-   "src/utils/cuda_error_handling.cu" )
+   "src/utils/cuda_error_handling.cpp"
+   "src/utils/cuda_error_handling.hpp" )
 
 target_link_libraries( vecmem_cuda
    PUBLIC vecmem::core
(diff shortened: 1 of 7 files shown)


cuda/CMakeLists.txt
```diff
@@ -18,6 +20,9 @@ vecmem_add_library( vecmem_cuda cuda SHARED
...
    "src/memory/cuda/direct_memory_manager.cu"
    # Utilities.
-   "include/vecmem/utils/cuda_error_handling.hpp"
-   "src/utils/cuda_error_handling.cu" )
+   "src/utils/cuda_error_handling.cpp"
+   "src/utils/cuda_error_handling.hpp" )
+
+target_link_libraries( vecmem_cuda PRIVATE CUDA::cudart )
+
 target_link_libraries( vecmem_cuda PUBLIC vecmem::core )
```
krasznaa, line 28: You should write this in a bit more compact form:

```cmake
target_link_libraries( vecmem_cuda
   PUBLIC vecmem::core
   PRIVATE CUDA::cudart )
```

tests/cuda/test_cuda_containers_kernels.cu
```diff
@@ -33,6 +33,11 @@ void linearTransformKernel( std::size_t size, const int* input, int* output ) {
...
 
-   VECMEM_CUDA_ERROR_CHECK( cudaGetLastError() );
-   VECMEM_CUDA_ERROR_CHECK( cudaDeviceSynchronize() );
+   if (cudaGetLastError() != cudaSuccess) {
+      throw std::runtime_error("Failure!");
+   }
+
+   if (cudaDeviceSynchronize() != cudaSuccess) {
+      throw std::runtime_error("Failure!");
+   }
```
krasznaa, line 42: :confused: Why didn't you use the (now) private `cuda_error_handling.hpp` header in the end? I would still go that route with these unit tests... Unless you want to introduce a separate checker macro for the tests. But I think that's not necessary.

Later on we will hopefully have a lot of tests with a lot of return codes to check. We shouldn't start checking them one by one like this.

krasznaa (commented):
I think at this point you should just rename the `.cu` files of the CUDA library to `.cpp`. Since as we discussed, none of them hold any device code.

cuda/CMakeLists.txt
```diff
@@ -10,6 +10,8 @@ enable_language( CUDA )
 # Project include(s).
 include( vecmem-compiler-options-cuda )
 
+find_package( CUDAToolkit )
```
krasznaa, line 13: One more thing... You should add `REQUIRED` here. The configuration should just bomb if CUDA is not available.

cuda/src/utils/cuda_error_handling.hpp
```diff
@@ -6,32 +6,26 @@
  */
 #pragma once
 
+#include <cuda_runtime_api.h>
+
 // Detray Data Model include(s).
 #include "vecmem/utils/types.hpp"
```
krasznaa, line 12: This include is no longer necessary...

stephenswat: The macro `VECMEM_CUDA_ERROR_CHECK` uses some CUDA runtime definitions like `errorCode` and `cudaSuccess`, so I think it's a good idea to keep this import.

tests/cuda/test_cuda_containers_kernels.cu
stephenswat: I'm extremely hesitant to use relative include paths across the test-source border, but we can always address that problem when we run into it later when we have more tests. So I'll include the macro by relative path for now.

PR #20, opened 2021-03-01
 Migrate CUDA code to use memory resources

stephenswat:
This merge request follows #17, but for the CUDA code. We add three new memory managers, for host memory, managed memory, and device memory. The usefulness of that last one is not super clear right now, but the first two should work transparently and flawlessly with the existing code.


cuda/src/memory/cuda/device_memory_resource.cpp
```diff
@@ -0,0 +1,42 @@
...
+ * VecMem project, part of the ACTS project (R&D line)
+ *
+ * (c) 2021 CERN for the benefit of the ACTS project
+ *
+ * Mozilla Public License Version 2.0
+ */
+
+#include "../../utils/cuda_error_handling.hpp"
+
+#include "vecmem/memory/resources/memory_resource.hpp"
```
krasznaa, line 11: I was pondering for a moment why you would put this include here. Did you on purpose put it here because `do_is_equal(...)` uses the `vecmem::memory_resource` type?

I myself would not have bothered, but I can sort of see the point of making this include explicitly here...

cuda/src/memory/cuda/device_memory_resource.cpp
```diff
@@ -0,0 +1,42 @@
...
+        VECMEM_CUDA_ERROR_CHECK(cudaFree(p));
+    }
+
+    bool device_memory_resource::do_is_equal(
+        const memory_resource & other
+    ) const noexcept {
+        const device_memory_resource * c;
+        c = dynamic_cast<const device_memory_resource *>(&other);
+
+        return c != nullptr;
```
krasznaa, line 40: A simple dynamic cast is enough for the host and managed memory resources. But for `vecmem::cuda::device_memory_resource` we need to be more explicit. These resources should be explicitly tied to a specific device.

So:
  - The class should receive an `int` device identifier in its constructor.
    * I would use the same logic here as in the "memory manager" classes. I.e. latch on to the "current" device if the user doesn't specify a device explicitly. (I would go with a `-1` default argument for the `int` constructor argument, and pick up the "current" device if the value is negative.)
  - In this function the class should check whether the `int` device identifiers are the same for the two memory resources.

cuda/src/memory/cuda/device_memory_resource.cpp
```diff
@@ -0,0 +1,42 @@
...
+
+#include <cuda_runtime_api.h>
+
+namespace vecmem::cuda {
+    void * device_memory_resource::do_allocate(
+        std::size_t bytes,
+        std::size_t
+    ) {
+        void * res;
+        VECMEM_CUDA_ERROR_CHECK(cudaMalloc(&res, bytes));
```
krasznaa, line 22: With `m_device` added, you should add

```c++
VECMEM_CUDA_ERROR_CHECK(cudaSetDevice(m_device));
```

before the `cudaMalloc(...)` call.

tests/cuda/test_cuda_allocators.cpp
```diff
@@ -41,52 +39,17 @@ void run_host_tests() {
...
-   run_device_tests();
+   run_host_tests(host_resource);
+   run_host_tests(managed_resource);
 
-   vecmem::memory_manager::instance().set(
-      std::make_unique< vecmem::cuda::direct_memory_manager >(
-         vecmem::cuda::direct_memory_manager::memory_type::device ) );
-   run_device_tests();
+   run_host_tests(host_resource);
+   run_host_tests(managed_resource);
```
krasznaa, line 52: I believe these two are not needed here.

krasznaa (changes requested):
On top of the specific comments, please add some documentation / comments to the code. :wink:

cuda/src/memory/cuda/device_memory_resource.cpp
stephenswat: Yes, that's the reason. Personally I believe in including things even if they are technically redundant due to being included in another header, just because I believe it expresses more clearly what the file depends on and in my experience it reduces the chance of strange include errors when headers change or get removed. So personally I prefer it this way, but I don't feel very strongly about it. If you'd prefer it removed let me know, and I'll get rid of it!

cuda/src/memory/cuda/device_memory_resource.cpp
stephenswat: Agreed. However, we also have the option of returning the CUDA device to what is was before the allocation, so that the call to this function serves as a sort of stack, and the CUDA device is reset when it exits. What do you think is best? Personally I like the idea of setting it back to what it was at the end of the call, since it reduces the number of outwardly observable side-effects.

tests/cuda/test_cuda_allocators.cpp
stephenswat: Yes, the reason they're still here is that something similar was being done in the old tests as well. I figured that might have been there to check if everything still works after switching back to host allocations? But I'm more than happy to remove these calls.

cuda/src/memory/cuda/device_memory_resource.cpp
stephenswat: There's a few design choices we should decide on before we move to make this change:

1. Do we pass the device identifier as an integer argument directly, or do we encapsulate it in a configuration struct? The latter is the approach taken by `std::pmr::synchronized_pool_resource` and other standard library memory resource implementations. It might make our life a lot easier if we decide to add more parameters in the future which we can't foresee yet. On the other hand implementing it as an integer is obviously easier.
2. Do we evaluate the default device eagerly or lazily? Very concretely, say the current device is set to _m_ at the time the memory resource is created, but the device is set to _n_ before the first allocation call is made. Does the allocation happen on device _m_ or on device _n_? Is the default device evaluated at memory resource construction time or at allocation time? I would personally prefer construction time, as having immutable device identifiers would make our implementation significantly easier.

cuda/src/memory/cuda/device_memory_resource.cpp
krasznaa: Sure, this is fine. I was just surprised for a moment about the extra include...

cuda/src/memory/cuda/device_memory_resource.cpp
krasznaa: Sure, I didn't even think of that. But I guess in the end the safest would be something like:

```c++
static std::mutex s_memAllocMutex;
...
void * device_memory_resource::do_allocate( std::size_t bytes, std::size_t ) {
   std::lock< std::mutex > lock( s_memAllocMutex );
   int currentDevice = 0;
   VECMEM_CUDA_ERROR_CHECK( cudaGetDevice( &currentDevice ) );
   VECMEM_CUDA_ERROR_CHECK( cudaSetDevice( m_device ) );
   void* res = nullptr;
   VECMEM_CUDA_ERROR_CUECK( cudaMalloc( &res, bytes ) );
   VECMEM_CUDA_ERROR_CUECK( cudaSetDevice( currentDevice ) );
   return res;
}
```

I.e. we would even need to use a mutex to avoid parallel threads doing something unexpected.

tests/cuda/test_cuda_allocators.cpp
krasznaa: Previously the test used 2 different "memory managers". Once we'll have multiple "memory resources" (built on top of the basic ones), we can add back the extra tests. But for now just testing `host_resource` and `managed_resource` once should be enough. :wink:

cuda/src/memory/cuda/device_memory_resource.cpp
krasznaa: I think at this point introducing an extra struct would be over-engineering. I can see the value of an extra struct once one has >5 parameters controlling how the memory resource should behave. But until then I believe individual constructor parameters should be easier on everyone. (Our code and the users' code alike.)

I believe the code should evaluate the "current device" (if the user didn't specify one explicitly) at the construction of `vecmem::cuda::device_memory_resource`. To me this seems much more natural than the code deciding about the device during the first request for memory.

Plus it makes the memory resource code a lot simpler if we do it this way...

cuda/src/memory/cuda/device_memory_resource.cpp
stephenswat: In fact, the CUDA device selector API is already thread-safe and thread-local, so this should not be necessary!

cuda/src/memory/cuda/device_memory_resource.cpp
```diff
@@ -0,0 +1,70 @@
...
+    ) {
+        int current_device = 0;
+
+        VECMEM_CUDA_ERROR_CHECK(cudaGetDevice(&current_device));
+        VECMEM_CUDA_ERROR_CHECK(cudaSetDevice(m_device));
+
+        VECMEM_CUDA_ERROR_CHECK(cudaFree(p));
+
+        VECMEM_CUDA_ERROR_CHECK(cudaSetDevice(current_device));
+    }
```
krasznaa, line 55: I think `cudaFree` does not need all this machinery with switching the current device around.

But...

Since you introduced `cuda_wrappers.hpp` / `cuda_wrappers.cpp` anyway, I wonder if it would be possible to introduce a helper construct for these sort of calls. I would imagine the following kind of UI:

```c++
details::run_on_device( m_device )( []{
      VECMEM_CUDA_ERROR_CHECK( cudaFree( p ) );
   } );
```

Naturally `vecmem::cuda::details::run_on_device` would be (sort of) a functor, that could execute "something executable" while setting the current device to the value that it received in its constructor.

Such a construct could come in handy in some other places as well later on...

krasznaa (commented):
Other than the one comment about the current device handling, I'm happy with everything else.

cuda/src/memory/cuda/device_memory_resource.cpp
stephenswat: Hmm, that's an interesting idea. I was actually thinking of an RAII-style device selection mechanism similar to `std::lock_guard` earlier, where all the setting and restoring would happen in the constructor and destructor of an object, respectively:

```cpp
void * device_memory_resource::do_allocate(
    std::size_t bytes,
    std::size_t
) {
    details::select_device(m_device);

    void * res;
    VECMEM_CUDA_ERROR_CHECK(cudaMalloc(&res, bytes));

    return res;
}
```

I think these are both elegant ways to handle the problem at hand here.

krasznaa: :thinking: You're right, that would be a little easier to implement. And I myself have also used that sort of code for handling "ROOT directories" in my analysis code in the past. I just didn't think of it in this instance...

stephenswat: If you're happy with the code as it is now, I'd suggest we merge this and I can start to work on a nicer way to manage the device selection. :smile:

krasznaa (approved):
Sure, let's conclude on this PR...

PR #21, opened 2021-03-01
Remove the default value for the memory manager

stephenswat:
In the old global state memory manager architecture, there is a default memory manager which is currently set to the host resource. However, as we move towards the new memory resource architecture this is no longer necessary, and I'd like to get rid of the existing core memory manager. For that, we have to remove the default value for the memory manager.


krasznaa (approved):
Sure, let's do this.

PR #22, opened 2021-03-01
Make Github Actions workflows compatible with act

stephenswat:
In order to test the code locally according to the rules defined in the Github Actions workflows, I am using act (not to be confused with ACTS, see https://github.com/nektos/act) to simulate the workflows. However, it seems that not all Github Actions code works properly with act. There are some problems with build path substitutions that act doesn't seem to accept. Some minor changes can make the workflows work with both act and Github actions, which will make development of HIP and SYCL code a lot easier for me.

### DIFF
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -28,21 +28,17 @@ jobs:
     steps:
     # Use a standard checkout of the code.
     - uses: actions/checkout@v2
-    # Set up the build directory.
-    - name: Create Build Environment
-      run: cmake -E make_directory ${{github.workspace}}/build
     # Run the CMake configuration.
     - name: Configure
-      working-directory: ${{github.workspace}}/build
-      run: cmake -DCMAKE_BUILD_TYPE=${{ matrix.BUILD_TYPE }} $GITHUB_WORKSPACE
+      run: cmake -DCMAKE_BUILD_TYPE=${{ matrix.BUILD_TYPE }} -S $GITHUB_WORKSPACE -B build
     # Perform the build.
     - name: Build
-      working-directory: ${{github.workspace}}/build
-      run: cmake --build .
+      run: cmake --build build
     # Run the unit test(s).
     - name: Test
-      working-directory: ${{github.workspace}}/build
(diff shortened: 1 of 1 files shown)


.github/workflows/builds.yml
```diff
@@ -28,21 +28,17 @@ jobs:
...
     # Use a standard checkout of the code.
     - uses: actions/checkout@v2
-    # Set up the build directory.
-    - name: Create Build Environment
-      run: cmake -E make_directory ${{github.workspace}}/build
     # Run the CMake configuration.
     - name: Configure
-      working-directory: ${{github.workspace}}/build
-      run: cmake -DCMAKE_BUILD_TYPE=${{ matrix.BUILD_TYPE }} $GITHUB_WORKSPACE
+      run: cmake -DCMAKE_BUILD_TYPE=${{ matrix.BUILD_TYPE }} -S $GITHUB_WORKSPACE -B build
```
krasznaa, line 33: It **really** surprised me that this worked. And I'm still a bit confused as to what's going on...

While GitHub has been teasing us for some time now that their default/native runners will switch to Ubuntu 20.04, they are still running 18.04. And Ubuntu 18.04 comes with CMake 3.10.

https://packages.ubuntu.com/bionic/cmake

Or at least it **came** with CMake 3.10 ever since I can remember. But I just now logged into one of the CERN Ubuntu 18.04 machines, and was surprised to see:

```
[bash][atspot01]:~ > which cmake
/usr/bin/cmake
[bash][atspot01]:~ > cmake --version
cmake version 3.13.4

CMake suite maintained and supported by Kitware (kitware.com/cmake).
[bash][atspot01]:~ >
```

So it seems that very recently Ubuntu 18.04 switched to CMake 3.13. Which, as it happens, supports the `-B` / `-S` flags. (While CMake 3.10 didn't yet.)

I guess I'm just trying to explain why the different platforms used different incantations to CMake. (The Docker images use a much newer version of CMake.) But as long as this formalism works at the current moment, let's go with this one.

krasznaa (approved):
Looks good.

PR #24, opened 2021-03-02
Remove the old-style core memory manager

stephenswat:
The old style memory manager is no longer necessary for the core code, and can be removed.  This commit removes the old code in order to clean up the repository.


krasznaa (approved):
Sure, let's start removing the old style code.

PR #25, opened 2021-03-02
vecmem::array, main branch (2021.03.02.)

krasznaa:
Created `vecmem::array`, which is meant to mimic [std::array](https://en.cppreference.com/w/cpp/container/array) for managing memory on any resource.

While the memory allocation for `vecmem::array` happens on the heap at runtime (unlike what [std::array](https://en.cppreference.com/w/cpp/container/array) does), the UI of this type (as opposed to `vecmem::vector`) should allow for some clever pieces of templating in client code.

Pinging @stephenswat. :wink:


core/include/vecmem/containers/array.hpp
```diff
@@ -0,0 +1,154 @@
...
+      /// for the size template parameter.
+      ///
+      array( memory_resource& resource );
+
+      /// Constructor with a size and a memory resource to use
+      ///
+      /// Can only be used if the user uses the default (zero) value for the
+      /// size template parameter.
+      ///
+      array( std::size_t size, memory_resource& resource );
```
paulgessinger, line 110: Maybe SFINAE it out then in case `N!=0`? Also, I'd consider reversing the argument order.

paulgessinger: I mean, the `static_assert` does almost the same tbf

krasznaa: My first thought was also SFINAE. :smile: But...

```
[ 57%] Building CXX object tests/core/CMakeFiles/vecmem_test_core_arrays.dir/test_core_arrays.cpp.o
In file included from /home/krasznaa/ATLAS/projects/vecmem/vecmem/tests/core/test_core_arrays.cpp:9:
/home/krasznaa/ATLAS/projects/vecmem/vecmem/core/include/vecmem/containers/array.hpp:139:68: error: cannot add a default template argument to the definition of a member of a class template
   template< std::enable_if_t< is_different< N, 0 >::value, bool > = true >
                                                                   ^ ~~~~
/home/krasznaa/ATLAS/projects/vecmem/vecmem/core/include/vecmem/containers/array.hpp:148:64: error: cannot add a default template argument to the definition of a member of a class template
   template< std::enable_if_t< is_equal< N, 0 >::value, bool > = true >
                                                               ^ ~~~~
```

Apparently you can't just provide a single default template argument for SFINAE to work. There would have to be a specific template parameter for the constructor as well.

So after some fumbling I went with `static_assert(...)`...

paulgessinger: Yeah I think it would have to be something like 

```cpp
template <typename _N = N, std::enable_if_t< _N ==0 >::value, bool > = true
array(std::size_t, memory_resource& resource) {
  // ...
}
```

to lift it into a constructor template. (Pseudo code, this won't compile). In the end it's not really important anyway.

PR #26, opened 2021-03-02
Implement a binary paging memory manager

stephenswat:
So far, we only have allocating memory resources in `vecmem`, such as the CUDA or `malloc` allocators. We don't yet have an efficient mechanism to manage that memory after it is allocated, outside of the `std::pmr` namespace. This merge request aims to change that by porting the binary paging memory manager that I previously implemented for the main ACTS repository to the new memory resource architecture. We provide some tests to show basic functionality, and the way in which this memory resource can be used.


core/src/memory/binary_page_memory_resource.cpp
```diff
@@ -0,0 +1,348 @@
...
+        /*
+         * All malloc resources are equal to each other, because they have no
+         * internal state. Of course they have a shared underlying state in the
+         * form of the underlying C library memory manager, but that is not
+         * relevant for us.
+         */
+        const binary_page_memory_resource * c;
+        c = dynamic_cast<const binary_page_memory_resource *>(&other);
+
+        return this == c;
```
krasznaa, line 171: :thinking: What's the purpose of this check in the end? I was thinking of this as a test of memory type, and not so much as a test of the exact memory resource class. But I might be thinking about it incorrectly...

What I'm trying to get at is that I thought that in the "high level" memory resources we would compare the "low level" memory resources that they are using. Like:

```c++
return m_upstream->is_equal( *( c->m_upstream ) );
```

But I'm honestly not sure what the correct behaviour should be here... :confused:

core/src/memory/binary_page_memory_resource.cpp
```diff
@@ -0,0 +1,348 @@
...
+        /*
+         * Making too many small allocations here would be a bad idea, so we
+         * take a minimum size of 1 megabyte.
+         */
+        size = std::max(size, static_cast<std::size_t>(1048576));
+
+        /*
+         * Allocate the memory upstream and gather the address.
+         */
+        void * addr = m_upstream->allocate(size);
```
krasznaa, line 238: I wonder if we should pass along the alignment value received by `do_allocate(...)`. We don't use the alignment for anything yet, but as long as [std::pmr::memory_resource](https://en.cppreference.com/w/cpp/memory/memory_resource) explicitly deals with memory alignment, our "high level" memory resources should also be aware of it.

core/include/vecmem/memory/binary_page_memory_resource.hpp
```diff
@@ -0,0 +1,159 @@
...
+     * creates a binary tree of pages which can be either vacant, occupied, or
+     * split.
+     */
+    class binary_page_memory_resource : public memory_resource {
+    public:
+        /**
+         * @brief Initialize a binary page memory manager depending on an
+         * upstream memory resource.
+         */
+        binary_page_memory_resource(memory_resource *);
```
krasznaa, line 35: I prefer using references in functions that must receive an argument. The member variable should still be a pointer, as that is a lot easier on the compiler for generating copy/move constructors/assignment operators automatically. But here you should receive the upstream `memory_resource` rather as a reference.

krasznaa (changes requested):
I'm pretty happy with this code. Took me a while to understand why you used `std::unique_ptr` for the main vector of `page` objects, but I think I get it now. :stuck_out_tongue:

core/include/vecmem/memory/binary_page_memory_resource.hpp
stephenswat: I totally agree with you. I went for a pointer argument here because that's what the standard library does, but I honestly have no idea why.

core/src/memory/binary_page_memory_resource.cpp
stephenswat: The C++ standard requires `do_is_equal` to return true if and only if the one memory resource can free memory allocated with the other and vice versa. It's possible to have two of these memory resources with completely different internal states, so I figured this would be a good-enough solution to check that equality constraint.

core/src/memory/binary_page_memory_resource.cpp
stephenswat: Yep, this is a good point.

core/src/memory/binary_page_memory_resource.cpp
krasznaa: :thinking: Okay, I'm fine with that.

But in that case I think we could spare ourselves the dynamic cast all together, and just do a pointer value comparison. Wouldn't

```c++
return ( this == &other );
```

work without anything else?

core/include/vecmem/memory/binary_page_memory_resource.hpp
krasznaa: I think the PMR implementation can accept `nullptr` as the memory resource. In which case it would fall back on some standard memory resource. (Which it manages through those helper functions that you pointed out earlier.)

So in their design the pointer value makes sense. (I think...) But in this project, as we discussed before, let's not bother with default memory resources...

core/src/memory/binary_page_memory_resource.cpp
stephenswat: Thinking about it, due to the way this allocator works alignment is both a little harder and a little easier than normal... Passing on the argument here won't work as such, I think. Would you mind crossing that bridge when we get to it at a later date?

krasznaa (approved):
Fine, let's leave the alignment aside for now...

PR #27, opened 2021-03-03
Add an RAII CUDA device switching mechanism 

stephenswat:
This new class allows us to switch CUDA devices in an elegant, modern C++ fashion that is scope-safe. When an instance of this class is constructed, the CUDA device is set to whatever was passed as a parameter, and the previously selected CUDA device is restored when the object is destroyed.


.vscode/settings.json
```diff
@@ -1,6 +1,70 @@
 {
    "files.associations": {
       "*.hip": "cpp",
-      "*.sycl": "cpp"
+      "*.sycl": "cpp",
+      "memory_resource": "cpp",
```
krasznaa, line 5: I know, I know... This is exactly what @paulgessinger warned me about... :frowning:

Still, this should not be added into the repository...

cuda/src/memory/cuda/device_memory_resource.cpp
```diff
@@ -44,14 +39,9 @@ namespace vecmem::cuda {
         std::size_t,
         std::size_t
     ) {
-        int current_device = 0;
-
-        VECMEM_CUDA_ERROR_CHECK(cudaGetDevice(&current_device));
-        VECMEM_CUDA_ERROR_CHECK(cudaSetDevice(m_device));
+        details::select_device dev(m_device);
```
krasznaa, line 42: I still don't think that this is necessary in this function... :thinking:

krasznaa (commented):
As discussed in #20, I'm all for this update...

.vscode/settings.json
paulgessinger: Ha! I get to say "I told you so!". :D

stephenswat: Oh no! I usually use `git add -i` for everything to avoid this, but the one time I commit the crime of `git commit -a` this happend! :sob:

cuda/src/memory/cuda/device_memory_resource.cpp
stephenswat: I'm not sure, there doesn't seem to be a clear answer to this anywhere, but even [some of NVIDIA's own examples](https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/examples.html) set the device before calling free.

krasznaa: Fine, you convinced me. (I thought the comment would be added when pressing "Resolve conversation"...)

krasznaa (approved):
Looks good.

PR #28, opened 2021-03-03
Remove old-style memory management in CUDA code

stephenswat:
Like with the core CPU code, we still have a bit of the old-style CUDA memory management code in the repository which is now unused. This commit removes some of that code in the name of cleanup.


tests/cuda/test_cuda_containers.cpp
```diff
@@ -11,7 +11,6 @@
 #include "vecmem/allocators/allocator.hpp"
 #include "vecmem/containers/vector.hpp"
 #include "vecmem/memory/memory_manager.hpp"
```
krasznaa, line 13: I think you could remove the "memory_manager.hpp" include as well. We can't remove that header until **all** of the "old style" memory management is removed, but at least this test should not include it anymore.

krasznaa (commented):
Apart from the tiny suggestion, looks good.

tests/cuda/test_cuda_containers.cpp
stephenswat: Good one, fixed!

krasznaa (approved):
Looks good.

PR #29, opened 2021-03-03
Implement a contiguous downstream memory resource

stephenswat:
There are a lot of use cases in which you might want to allocate contiguous blocks of memory, but this is not supported by default by many allocators. This new downstream memory resource guarantees that consecutive allocations will be exactly contiguous in memory, which makes it easier to allocate memory with efficient data copies and locality in mind.

### DIFF
--- a/core/include/vecmem/memory/contiguous_memory_resource.hpp
+++ b/core/include/vecmem/memory/contiguous_memory_resource.hpp
@@ -66,7 +66,6 @@ namespace vecmem {
         memory_resource & m_upstream;
         const std::size_t m_size;
         void * const m_begin;
-        void * const m_end;
         void * m_next;
     };
 }
--- a/core/include/vecmem/memory/contiguous_memory_resource.hpp
+++ b/core/include/vecmem/memory/contiguous_memory_resource.hpp
@@ -1,0 +1,72 @@
+/**
+ * VecMem project, part of the ACTS project (R&D line)
+ *
+ * (c) 2021 CERN for the benefit of the ACTS project
+ *
+ * Mozilla Public License Version 2.0
+ */
+
+#pragma once
+
+#include "vecmem/memory/resources/memory_resource.hpp"
+
+#include <cstddef>
+
+namespace vecmem {
+    /**
+     * @brief Downstream allocator that ensures that allocations are contiguous.
+     *
+     * When programming for co-processors, it is often desriable to keep
+     * allocations contiguous. This downstream allocator fills that need. When
+     * configured with an upstream memory resource, it will start out by
+     * allocating a single, large, chunk of memory from the upstream. Then, it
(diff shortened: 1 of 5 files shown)


tests/core/test_core_contiguous_memory_resource.cpp
```diff
@@ -0,0 +1,39 @@
...
+   vec1.reserve(100);
+   vec2.reserve(100);
+   vec3.reserve(100);
+   vec4.reserve(100);
+   vec5.reserve(100);
+
+   assert(static_cast<void *>(vec2.data()) == static_cast<void *>(static_cast<int *>(vec1.data()) + 100));
+   assert(static_cast<void *>(vec3.data()) == static_cast<void *>(static_cast<char *>(vec2.data()) + 100));
+   assert(static_cast<void *>(vec4.data()) == static_cast<void *>(static_cast<double *>(vec3.data()) + 100));
+   assert(static_cast<void *>(vec5.data()) == static_cast<void *>(static_cast<float *>(vec4.data()) + 100));
```
krasznaa, line 36: I don't think so much casting is necessary here. :confused: I think the following should be enough:

```c++
assert( vec2.data() == ( vec1.data() + 100 ) );
...
```

core/include/vecmem/memory/contiguous_memory_resource.hpp
```diff
@@ -0,0 +1,72 @@
...
+        ) override;
+
+        virtual bool do_is_equal(
+            const memory_resource &
+        ) const noexcept override;
+
+        memory_resource & m_upstream;
+        const std::size_t m_size;
+        void * const m_begin;
+        void * const m_end;
```
krasznaa, line 69: 2 of these 3 should be enough to implement the desired behaviour. I would go with just keeping `m_begin` and `m_size`.

krasznaa: In fact since you anyway need `m_begin` and `m_size` for the de-allocation, it's pretty clear that `m_end` is the one that's unnecessary...

krasznaa (commented):
Other than the minor technical comments, I'm all for this!

tests/core/test_core_contiguous_memory_resource.cpp
krasznaa: Okay, it wasn't quite that easy. Casting everything to `void*` is necessary in the end, since the different vectors point at different types, and comparing pointers of different types is a no-no.

But at least it was possible to get rid of one cast per line... :stuck_out_tongue:

krasznaa (approved):
I decided to try whether I can tweak your PR. :stuck_out_tongue:

Hopefully you don't object to either of my changes...

stephenswat:
I was just working on this, but this solves the problem too! Looks good! :smile:

PR #30, opened 2021-03-03
SYCL Memory Resources, main branch (2021.03.03.)

krasznaa:
This PR updates the `vecmem::sycl` library to provide "memory resources" instead of its current "memory manager" for access to SYCL managed memory.

As a technicality, in order to allow the memory resources not to depend on SYCL in their public UI, I introduced `vecmem::sycl::queue_wrapper`. This type, through `vecmem::sycl::details::opaque_queue`, allows us to pass queue objects around without depending on their concrete type in the public interface of `vecmem::sycl`.

This is just the beginning of this development at the moment, I still have 2 more memory resources to add, and the tests to update.


sycl/src/utils/sycl/opaque_queue.hpp
```diff
@@ -0,0 +1,20 @@
...
+ */
+#pragma once
+
+// SYCL include(s).
+#include <CL/sycl.hpp>
+
+namespace vecmem::sycl::details {
+
+   /// Helper class for managing queue objects in memory
+   class opaque_queue : public cl::sycl::queue {
```
stephenswat, line 15: What is the motivation for this class?

sycl/include/vecmem/memory/sycl/details/memory_resource_base.hpp
```diff
@@ -0,0 +1,41 @@
...
+   /// SYCL memory resource base class
+   ///
+   /// This class is used as base by all of the oneAPI/SYCL memory resource
+   /// classes. It holds functionality that those classes all need.
+   ///
+   class memory_resource_base : public memory_resource {
+
+   public:
+      /// Constructor on top of a user-provided queue
+      memory_resource_base( const queue_wrapper& queue = { "" } );
```
stephenswat, line 24: I not sure I really like this, but I understand it's necessary due to the way SYCL work, and I don't see another way of doing. How do we envision this working in actual SYCL code? Does the user intersperse their standard SYCL queue operations with creating `vecmem` vectors and such? For example:

```cpp
sc::sycl::queue q;
vecmem::sycl::device_memory_resource mem(q);
vemem::vector vec(&mem);
vec = {41, 51, 79};
q.submit([&](...){...});
```

Is that roughly the design we're going for?

sycl/src/utils/sycl/opaque_queue.hpp
krasznaa: Ha... This is something that I took from https://github.com/root-project/root/commit/289c1d701d8d36cecd8376f03ffc9651644dd315. :smile:

Intel (as we see in TBB's example) likes to define its types in fairly complicated ways. At the moment `cl::sycl::queue` is a "real class". But at any point it can become a typedef. This sort of design is meant to guard us against that. While still making it possible to use `std::unique_ptr` for managing the queue's memory.

sycl/include/vecmem/memory/sycl/details/memory_resource_base.hpp
krasznaa: In the diffs it's pretty difficult to see this, but this is one of the things that I liked about this design the most. The SYCL test does the following in this PR:

```c++
   // Create the SYCL queue that we'll be using in the test.
   cl::sycl::queue queue{ vecmem::sycl::device_selector() };
   std::cout << "Using device:" << std::endl;
   std::cout << "    name: "
             << queue.get_device().get_info< cl::sycl::info::device::name >()
             << std::endl;
   std::cout << "  vendor: "
             << queue.get_device().get_info< cl::sycl::info::device::vendor >()
             << std::endl;
   std::cout << " version: "
             << queue.get_device().get_info< cl::sycl::info::device::version >()
             << std::endl;

   // Set up the memory resource for the test.
   vecmem::sycl::shared_memory_resource resource( &queue );

   // Create an input and an output vector.
   auto inputvec = make_input_vector( resource );
   auto outputvec = make_output_vector( resource );
```

https://github.com/acts-project/vecmem/blob/SyclMemoryResource-main-20210303/tests/sycl/test_sycl_containers.sycl#L36-L54

I was trying to achieve exactly what you wrote, to be able to seamlessly give a user-created `cl::sycl::queue` object to the memory resources, while also allowing the resources to be created without any SYCL code dependence as well, in case that is necessary.

sycl/include/vecmem/utils/sycl/queue_wrapper.hpp
```diff
@@ -0,0 +1,62 @@
...
+   namespace details {
+      class opaque_queue;
+   }
+
+   /// Wrapper class for @c cl::sycl::queue
+   ///
+   /// It is necessary for passing around SYCL queue objects in code that should
+   /// not be directly exposed to the SYCL headers.
+   ///
+   class queue_wrapper {
```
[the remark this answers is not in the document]
krasznaa: This is really just a matter of being able to handle `vecmem::sycl::device_memory_resource` and friends in "pure C\+\+ code". Anything that includes `<CL/sycl.hpp>` needs to be compiled using the `-fsycl` flag of the Intel compiler at the moment. I really only want clients to have to use that flag in code in which they **really** need to interact with SYCL.

krasznaa:
We really need to conclude on this, as I can't commence with https://github.com/acts-project/vecmem/tree/DeviceVectorUpdate-main-20210304 before merging this and #31 in first...

If you have any remaining issues with these changes please let me know very quickly...

stephenswat:
It's good to go ahead for merging as far as I'm concerned!

PR #31, opened 2021-03-04
HIP Memory Resources, main branch (2021.03.04.)

krasznaa:
Similar to #30 this PR introduces two new memory resources (`vecmem::hip::device_memory_resource`, `vecmem::hip::host_memory_resource`), along with all the infrastructure needed to make them work.

Just like for `vecmem::cuda`, I had to give up on using the `.hip` extension in the `vecmem::hip` source files. As `HIP_PLATFORM=nvcc` would not succeed, as that maxes out at C\+\+14. But while CMake provides [FindCUDAToolkit.cmake](https://cmake.org/cmake/help/latest/module/FindCUDAToolkit.html), it does no such thing for HIP. :frowning: So I had to write `FindHIPToolkit.cmake` myself...

@stephenswat, note that I did add `vecmem::hip::details::run_on_device` in this PR. :smile: I just wanted to see how it would work. It didn't really make the code any shorter, but I kind of like how it can be used in `device_memory_resource.cpp`... (I did also introduced `vecmem::hip::details::select_device`, as RAII is definitely the thing to use underneath to, for instance, handle exceptions correctly.)

I then updated the HIP tests to use these new memory resources, and removed the old `vecmem::hip::direct_memory_manager` completely.


krasznaa:
I'll just plow forward with this one as well...

PR #32, opened 2021-03-04
GoogleTest Introduction, main branch (2021.03.04.)

krasznaa:
Taught the project how to build [googletest](https://github.com/google/googletest) (when testing is enabled), and updated the core library's tests to make use of it.

@stephenswat, with the `test_core_allocators.cpp` (we should rename it at one point...) test testing all of the memory resources with this update, do we need the separate `test_core_binary_page.cpp` test? The specific test for `vecmem::contiguous_memory_resource` of course makes a lot of sense, so I updated it to the "new formalism". But I'm not sure it makes sense to update `test_core_binary_page.cpp`...

Note that I decided to build `ExternalProject_Add(...)` instead of making [googletest](https://github.com/google/googletest) a submodule of this repository. This way, if somebody builds our project with the unit testing disabled (which could be a very reasonable thing to do for projects using vecmem), they don't have to bother with downloading those sources. But I'm open to disagreement on this. :stuck_out_tongue:


stephenswat:
Personally I would prefer to compile all the tests per module into one binary, instead of keeping all the binaries we have now. How would you feel about this?

krasznaa:
> Personally I would prefer to compile all the tests per module into one binary, instead of keeping all the binaries we have now. How would you feel about this?

Sure, why not. We'll have to see if all of the tests in the "GPU libraries" will fit into single executables, but at least for `vecmem::core` this was very doable.

krasznaa:
Okay, let me consider it accepted then. :wink:

PR #33, opened 2021-03-05
Device Vector Updates, main branch (2021.03.05.)

krasznaa:
In preparation for introducing more complicated (2 dimensional...) data types, I first spent a bit of time cleaning up the implementation of `vecmem::device_vector` and `vecmem::const_device_vector`.

Since for 2-dimensional types we will need to ship a non-trivial amount of data to the GPUs, I thought I would introduce "data types" for the 1-dimensional types as well. These became `vecmem::device_vector_data` and `vecmem::const_device_vector_data`. These can be created from any host vector type, and then passed trivially to CUDA/HIP/SYCL device code as they are. The `vecmem::device_vector` and `vecmem::const_device_vector` types can then be created from them directly.

While doing all of this, I also introduced a hand-written `vecmem::reverse_iterator` type. This really is just a copy-paste of `std::reverse_iterator`, it "just" has all the CUDA/HIP markings necessary to make it usable in device code. With that added I could implement the "iterator functionality" of the device vector types to actually work in device code.

Feedback is very welcome on these changes!


krasznaa:
And boom goes the dynamite... :smile:

I really disliked for quite a while how I ended up creating separate `vecmem::device_vector` and `vecmem::const_device_vector` types. Especially while cleaning up these types in the PR, it became quite apparent that maintaining them separately is really not a good idea.

But as it turns out, we really don't need to! Since now that `vecmem::device_vector` is implemented in a slightly smarter way than it used to be, `vecmem::const_device_vector<T>` was really just the same as `vecmem::device_vector<const T>`. I really thought that it would be more of a struggle to "merge" the two types, but it wasn't.

I still kept the `vecmem::const_device_vector<T>` type around, as I believe it will be useful in some templating in algorithmic code later on. But by now at least it's just a typedef, so practically no maintenance effort should be involved!

I'm pretty pleased to end the week on this. :smile:

krasznaa:
Okay, **now** the PR should be ready...

I introduced `vecmem::get_data(...)` functions to more conveniently be able to create `vecmem::device_vector_data` objects. Without them the SYCL example needed to be a lot more explicit with its type names than I would've liked.

I had to be a bit careful with this, since I also needed to teach the code how to convert from `vecmem::device_vector_data<T>` to `vecmem::device_vector_data<const T>` (the artist formerly known as `vecmem::const_device_vector_data<T>` :stuck_out_tongue:). Since in exactly the SYCL code's example this is necessary. But it mustn't have been done in a way that would interfere with the default constructability of these objects. And so a pretty difficult SFINAE constructor was born...

Any feedback is still very welcome.

krasznaa:
Since I'll need some of these updates for my next developments, I'll just plow ahead now... Let me know afterwards if you think something should be tweaked.

core/include/vecmem/containers/device_vector.ipp
```diff
@@ -0,0 +1,279 @@
...
+   device_vector< TYPE >::operator=( const device_vector& rhs ) {
+
+      // Prevent self-assignment.
+      if( this == &rhs ) {
+         return *this;
+      }
+
+      // Copy the other object's payload.
+      m_size = rhs.m_size;
+      m_ptr = rhs.m_ptr;
```
stephenswat, line 40: I would expect this operation to do a copy, not alias the pointer of the right hand side. Might be confusing to users who are more used to the standard library behaviour. I would suggest documenting this well, or adding a move assignment operator.

core/include/vecmem/containers/device_vector.ipp
```diff
@@ -0,0 +1,279 @@
...
+      return *this;
+   }
+
+   template< typename TYPE >
+   VECMEM_HOST_AND_DEVICE
+   typename device_vector< TYPE >::reference
+   device_vector< TYPE >::at( size_type pos ) {
+
+      // Check if the index is valid.
+      assert( pos < m_size );
```
stephenswat, line 52: I worry about this failing silently when `NDEBUG` is set. However, we can't really mirror the behaviour of the standard library because we can't really throw in device code. Explicitly documenting might be key here.

core/include/vecmem/containers/device_vector_data.ipp
```diff
@@ -0,0 +1,65 @@
...
+   device_vector_data( std::size_t size, pointer ptr )
+   : m_size( size ), m_ptr( ptr ) {
+
+   }
+
+   template< typename TYPE >
+   template< typename ALLOC >
+   VECMEM_HOST
+   device_vector_data< TYPE >::
+   device_vector_data( std::vector< value_type, ALLOC >& vec )
```
stephenswat, line 23: I don't think this constructor should exist like this, rather it should take its `std::vector` argument as an rvalue reference.

core/include/vecmem/containers/device_vector_data.ipp
```diff
@@ -0,0 +1,65 @@
...
+   : m_size( vec.size() ), m_ptr( vec.data() ) {
+
+   }
+
+   template< typename TYPE >
+   template< typename ALLOC >
+   VECMEM_HOST
+   device_vector_data< TYPE >::
+   device_vector_data( const std::vector< value_type, ALLOC >& vec )
```
stephenswat, line 32: Same as above.

core/include/vecmem/containers/device_vector_data.ipp
```diff
@@ -0,0 +1,65 @@
...
+
+   template< typename TYPE >
+   template< typename OTHERTYPE,
+             std::enable_if_t< ( ! std::is_same< TYPE, OTHERTYPE >::value ) &&
+                               std::is_convertible< OTHERTYPE*, TYPE* >::value,
+                               bool > >
+   VECMEM_HOST_AND_DEVICE
+   device_vector_data< TYPE >::
+   device_vector_data( const device_vector_data< OTHERTYPE >& parent )
+   : m_size( parent.m_size ), m_ptr( parent.m_ptr ) {
```
stephenswat, line 45: Are you sure this works? What if `OTHERTYPE` is a larger child class of `TYPE`, then the pointer arithmetic would no longer add up, would it?

tests/core/test_core_device_containers.cpp
```diff
@@ -0,0 +1,39 @@
...
+// System include(s).
+#include <type_traits>
+
+/// Test case for the custom device container types
+class core_device_container_test : public testing::Test {
+
+}; // class core_device_container_test
+
+// Test that the "simple" data types are trivially constructible.
+TEST_F( core_device_container_test, trivial_construct ) {
```
stephenswat, line 24: Can't we do these tests at compile time? Not that it really matters...

stephenswat (commented):
Looks excellent! There's a few things I would like to briefly discuss, but when these points are cleared up I am happy to merge this in.

core/include/vecmem/containers/device_vector.ipp
krasznaa: The behaviour **is** different from STL containers. Since these types do not own the memory that they operate on. As such, a copy can only do this. To create another `vecmem::device_vector` instance that would operate on the same array in memory.

Creating move operators/constructors could be useful, mainly for code documentation purposes. But performance-wise they will not bring us anything...

core/include/vecmem/containers/device_vector.ipp
krasznaa: I'm open to suggestions... Unfortunately I couldn't come up with anything better. I'm just not aware of anything cleaner for error handling in device code than `assert(...)` statements. (Not even separately for the individual platforms...)

We could possibly put in something that would trigger a memory error right there and then. That could possibly simplify debugging in optimised code a bit more. :thinking:

core/include/vecmem/containers/device_vector_data.ipp
krasznaa: I'm not sure what you mean. Unless my C\+\+ vocabulary is completely broken, the constructor **is** receiving an rvalue reference already. The way the two constructors differentiate between which of them is allowed to grab a non-const pointer to the vector's data and which is only allowed to grab a constant one, is by using rvalue references... :confused:

core/include/vecmem/containers/device_vector_data.ipp
krasznaa: Same comment as above. :stuck_out_tongue:

core/include/vecmem/containers/device_vector_data.ipp
krasznaa: I didn't run too rigorous tests, but I think automatic pointer conversion really only works for going from `FOO*` to `const FOO*`. Everything else needs to be explicitly declared as a conversion, no? :confused:

tests/core/test_core_device_containers.cpp
krasznaa: :thinking: Are you thinking of adding some `static_assert(...)`-s? Sure, we could do that as well...

stephenswat: It's a minor point, but catching these errors at compile time will surely make debugging a lot easier. We all know how... Unhelpful the C++ compiler can be when you make errors in your type-level programming. These tests might very well not even compile under certain error conditions.

core/include/vecmem/containers/device_vector_data.ipp
stephenswat: But we're throwing away the original type of the vector without actually converting anything on a bit-level. So I would think it possible for the array pointed at by `m_ptr` to consist of objects of a different size from `TYPE`.

core/include/vecmem/containers/device_vector_data.ipp
stephenswat: Sorry, I should be more clear, what I am suggesting is defining this constructor as `device_vector_data(std::vector<value_type, ALLOC> && vec)` (with two ampersands) to accept rvalue references. If we accept lvalue references like we do now it's possible for the size and content of the original container to go out of sync with the associated `device_vector_data`, which would be bad. Or at least, it could lead to some very unexpected behaviour. If we accept only rvalue references the compiler would protect the user against those mistakes.

core/include/vecmem/containers/device_vector.ipp
stephenswat: Right, I totally understand the need for this. But semantically I think this class should strictly be constructed and assigned from rvalue vectors. This is actually the exact same discussion we're having a few comments down. :smile:

core/include/vecmem/containers/device_vector_data.ipp
krasznaa: No, this does seem to be reasonably safe. You should try the following:
  - This does compile:

```c++
vecmem::vector< int > hostvec;
vecmem::device_vector_data< const int > test( hostvec );
```

  - But the following fails miserably:

```c++
vecmem::vector< int > hostvec;
vecmem::device_vector_data< short > test( hostvec );
```

Like:

```
/home/krasznaa/ATLAS/projects/vecmem/vecmem/tests/core/test_core_containers.cpp:56:40: error: no matching constructor for initialization of 'vecmem::device_vector_data<short>'
   vecmem::device_vector_data< short > dummy( m_reference_vector );
                                       ^      ~~~~~~~~~~~~~~~~~~
/home/krasznaa/ATLAS/projects/vecmem/vecmem/core/include/vecmem/containers/device_vector_data.hpp:26:11: note: candidate constructor (the implicit copy constructor) not viable: no known conversion from 'vecmem::vector<int>' (aka 'vector<int, polymorphic_allocator<int>>') to 'const vecmem::device_vector_data<short>' for 1st argument
   struct device_vector_data {
          ^
/home/krasznaa/ATLAS/projects/vecmem/vecmem/core/include/vecmem/containers/device_vector_data.hpp:26:11: note: candidate constructor (the implicit move constructor) not viable: no known conversion from 'vecmem::vector<int>' (aka 'vector<int, polymorphic_allocator<int>>') to 'vecmem::device_vector_data<short>' for 1st argument
/home/krasznaa/ATLAS/projects/vecmem/vecmem/core/include/vecmem/containers/device_vector_data.hpp:42:7: note: candidate template ignored: could not match 'short' against 'int'
      device_vector_data( std::vector< value_type, ALLOC >& vec );
      ^
/home/krasznaa/ATLAS/projects/vecmem/vecmem/core/include/vecmem/containers/device_vector_data.hpp:46:7: note: candidate template ignored: could not match 'short' against 'int'
      device_vector_data( const std::vector< value_type, ALLOC >& vec );
      ^
/home/krasznaa/ATLAS/projects/vecmem/vecmem/core/include/vecmem/containers/device_vector_data.hpp:60:7: note: candidate template ignored: could not match 'device_vector_data' against 'vector'
      device_vector_data( const device_vector_data< OTHERTYPE >& parent );
      ^
/home/krasznaa/ATLAS/projects/vecmem/vecmem/core/include/vecmem/containers/device_vector_data.hpp:34:7: note: candidate constructor not viable: requires 0 arguments, but 1 was provided
      device_vector_data() = default;
      ^
/home/krasznaa/ATLAS/projects/vecmem/vecmem/core/include/vecmem/containers/device_vector_data.hpp:37:7: note: candidate constructor not viable: requires 2 arguments, but 1 was provided
      device_vector_data( std::size_t size, pointer ptr );
      ^
1 error generated.
```

stephenswat: But what if you do:

```cpp
vecmem::vector<int> hostvec;
vecmem::device_vector_data<long> test(hostvec);
```

core/include/vecmem/containers/device_vector_data.ipp
stephenswat: Thinking about it more there are a few gaps in the specification of `std::vector` that makes this a little difficult... For example, the fact that `.data()` is not guaranteed to return the beginning of the allocation block. Makes taking ownership of it a whole lot harder. Seems like there was a proposal for something like this in [N4359](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2015/n4359.pdf) but it was abandoned. Shame!

krasznaa: :confused: We are grabbing these pointers from `vecmem::vector` objects. Which do need to stay alive until the end of the GPU kernel. Since we want to receive the results of the GPU's calculation in them.

Yes, it is very important that host code would stop modifying "host containers" before handing them off to the device code, and only touch these containers again after the device code has finished.

If we would force the host's `vecmem::vector` objects to become invalid once we hand them off to device code, the host code would need to receive a terrible amount of boilerplate. Like:

```c++
vecmem::vector< int > input, output;
vecmem::device_vector_data< const int > device_input( input );
vecmem::device_vector_data< int > device_output( output );
// Hand "device_input" and "device_output" to a GPU kernel, with "input" and "output" having become invalid.
vecmem::vector< int > output_after_gpu_kernel( ? );
```

Since `vecmem::vector` always manages its own memory in the current setup, I have no idea how we would look at the output of a GPU kernel in such a setup in any convenient way. :confused:

core/include/vecmem/containers/device_vector_data.ipp
krasznaa:
```
/home/krasznaa/ATLAS/projects/vecmem/vecmem/tests/core/test_core_containers.cpp:56:39: error: no matching constructor for initialization of 'vecmem::device_vector_data<long>'
   vecmem::device_vector_data< long > dummy( m_reference_vector );
                                      ^      ~~~~~~~~~~~~~~~~~~
/home/krasznaa/ATLAS/projects/vecmem/vecmem/core/include/vecmem/containers/device_vector_data.hpp:26:11: note: candidate constructor (the implicit copy constructor) not viable: no known conversion from 'vecmem::vector<int>' (aka 'vector<int, polymorphic_allocator<int>>') to 'const vecmem::device_vector_data<long>' for 1st argument
   struct device_vector_data {
          ^
/home/krasznaa/ATLAS/projects/vecmem/vecmem/core/include/vecmem/containers/device_vector_data.hpp:26:11: note: candidate constructor (the implicit move constructor) not viable: no known conversion from 'vecmem::vector<int>' (aka 'vector<int, polymorphic_allocator<int>>') to 'vecmem::device_vector_data<long>' for 1st argument
/home/krasznaa/ATLAS/projects/vecmem/vecmem/core/include/vecmem/containers/device_vector_data.hpp:42:7: note: candidate template ignored: could not match 'long' against 'int'
      device_vector_data( std::vector< value_type, ALLOC >& vec );
      ^
/home/krasznaa/ATLAS/projects/vecmem/vecmem/core/include/vecmem/containers/device_vector_data.hpp:46:7: note: candidate template ignored: could not match 'long' against 'int'
      device_vector_data( const std::vector< value_type, ALLOC >& vec );
      ^
/home/krasznaa/ATLAS/projects/vecmem/vecmem/core/include/vecmem/containers/device_vector_data.hpp:60:7: note: candidate template ignored: could not match 'device_vector_data' against 'vector'
      device_vector_data( const device_vector_data< OTHERTYPE >& parent );
      ^
/home/krasznaa/ATLAS/projects/vecmem/vecmem/core/include/vecmem/containers/device_vector_data.hpp:34:7: note: candidate constructor not viable: requires 0 arguments, but 1 was provided
      device_vector_data() = default;
      ^
/home/krasznaa/ATLAS/projects/vecmem/vecmem/core/include/vecmem/containers/device_vector_data.hpp:37:7: note: candidate constructor not viable: requires 2 arguments, but 1 was provided
      device_vector_data( std::size_t size, pointer ptr );
      ^
1 error generated.
```

Pointers **do not convert easily**...

stephenswat: Okay, hold on, this needs to become a little more complex:

```cpp
class A {
    int m_a;
};

class B : public A {
    int m_b;
};

vecmem::vector<B> hostvec;
vecmem::device_vector_data<A> test(hostvec);
```

Does that also throw an error?

krasznaa: No, that probably does not.

Fine, let's declare this constructor more narrowly. The code does need to be able to convert from `vecmem::device_vector_data<FOO>` to `vecmem::device_vector_data<const FOO>`. What's your proposal, what should the SFINAE expression be for only enabling that sort of `OTHERTYPE`?

core/include/vecmem/containers/device_vector.ipp
krasznaa: The question is not really the same in the two cases...

The question here is whether we should allow multiple `vecmem::device_vector` objects to operate on the same memory array in the GPU kernel. I believe this can be a useful feature. The N threads of the GPU will already use N `vecmem::device_vector` instances to access the same memory array. What harm does it do if the device code also has multiple instances of the object inside the same thread as well? (Other than allowing for inexperienced developers to produce some "sneakily broken" code.)

core/include/vecmem/containers/device_vector_data.ipp
stephenswat: Something like this, perhaps?

```cpp
template<typename OTHERTYPE,
    std::enable_if_t<
        (!std::is_same<TYPE, OTHERTYPE>::value) &&
        std::is_same<TYPE, std::add_const<OTHER_TYPE>::type>::value,
        bool
    >
>
```

krasznaa: :thinking: Okay, that's probably the way to go... I was thinking of using [std::remove_cv](https://en.cppreference.com/w/cpp/types/remove_cv) in some way, but that would still allow for some invalid combinations...

PR #34, opened 2021-03-08
Device Array Addition, main branch (2021.03.08.)

krasznaa:
This unfortunately grew into a bigger PR than I would've liked, but I thought it may still go through more quickly than if I would've split it up into multiple PRs that would all need each other.

First off, the PR makes some changes in how the `vecmem::(const_)device_vector_data` objects are created/handled, following the discussion in https://github.com/acts-project/vecmem/pull/33#discussion_r589258915.

Also based on that discussion, I stopped `vecmem::(const_)device_vector_data` from being constructible directly from `std::vector` types, as I wanted to re-use the same data type for communicating the payload of `vecmem::array` to `vecmem::(const_)device_array`. So the code was re-written to always require an explicit `vecmem::get_data(...)` call to get a "simple data object" from some host container in the host code.

With all those in place, I introduced the `vecmem::(const_)device_array` type(s), and updated all of the tests to make use of it.

@stephenswat, we should coordinate the discussion between #33 and this PR in some way, since this is where your suggestions from #33 could be implemented...


core/include/vecmem/containers/array.hpp
```diff
@@ -209,6 +211,18 @@ namespace vecmem {
 
    }; // class array
 
+   /// Helper function creating a @c vecmem::device_vector_data object
+   template< typename T, std::size_t N >
+   VECMEM_HOST
+   device_vector_data< T >
+   get_data( array< T, N >& a );
```
stephenswat, line 218: Why are these helper functions not methods of the `vecmem::array` class?

core/include/vecmem/containers/const_device_array.hpp
```diff
@@ -0,0 +1,18 @@
...
+#pragma once
+
+// Local include(s).
+#include "vecmem/containers/device_array.hpp"
+
+namespace vecmem {
+
+   /// Class mimicking a host-filled constant @c std::array in "device code"
+   template< typename T, std::size_t N >
+   using const_device_array = device_array< const T, N >;
```
stephenswat, line 16: I'm honestly not sure if it's worth having these aliases in the namespace, it's not too hard (and in my opinion actually clearer) for the user to set the template variable to `const T`, right? There is no `std::const_vector` in the standard library either. :thinking:

core/include/vecmem/containers/device_array.ipp
```diff
@@ -0,0 +1,271 @@
...
+      return m_ptr[ pos ];
+   }
+
+   template< typename T, std::size_t N >
+   VECMEM_HOST_AND_DEVICE
+   typename device_array< T, N >::reference
+   device_array< T, N >::front() {
+
+      // Make sure that there is at least one element in the vector.
+      static_assert( N > 0, "Cannot return first element of empty array" );
```
stephenswat, line 94: Should we put these static asserts in the class instead of in its methods? Creating an array of length 0 makes little sense at the type level.

core/include/vecmem/containers/vector.hpp
```diff
@@ -23,4 +26,20 @@ namespace vecmem {
     */
    template<typename T>
    using vector = std::vector<T, vecmem::polymorphic_allocator<T>>;
-}
+
+   /// Helper function creating a @c vecmem::device_vector_data object
+   template< typename TYPE, typename ALLOC >
+   VECMEM_HOST
+   device_vector_data< TYPE >
+   get_data( std::vector< TYPE, ALLOC >& vec );
```
stephenswat, line 34: Same question as before, why do we want these to be class-external functions instead of static?

stephenswat (approved):

core/include/vecmem/containers/vector.hpp
krasznaa: Let me answer this one first...

Since `vecmem::vector` is (currently) just a specialisation of `std::vector`, we need to make `get_data(...)` external to it. Since C++ doesn't allow us to extend existing classes with additional functions. (I don't want to make `vecmem::vector` inherit from `std::vector`, as inheriting from STL types is generally not a good idea.)

Since for `vecmem::vector` we have to have `get_data(...)` as an external function, I thought it would be a nice design to just have `vecmem::get_data(...)` functions for all of the types that vecmem handles/supports. For instance I'm now playing with the following type:

```c++
namespace vecmem {
   template< typename T >
   using jagged_vector = vector< vector< T > >;
}
```

For which I'm introducing a special `vecmem::jagged_vector_data` type (I'll want to propose a change to the naming of the `vecmem::device_vector_data` types in an upcoming PR...), which would be created by another specialisation of the `vecmem::get_data(...)` function. :wink:

core/include/vecmem/containers/device_array.ipp
krasznaa: Possibly... Though every other functionality of the code works fine with a zero sized container. And I could imagine using some type of template meta-programming that could result in instantiating zero sized arrays.

core/include/vecmem/containers/const_device_array.hpp
krasznaa: I'm thinking of the type of templating here that I had to use in the early version of my demonstrator code.

https://github.com/krasznaa/detray_data_model/blob/main/tests/detector.hpp

I just think that it could come in handy to easily template some "high level EDM type" on this type instead of on `vecmem::device_array`. Like:

```c++
template< template< typename, std::size_t > class array_type = vecmem::array >
struct condition_edm_type {
   array_type< float, 50 > m_data;
};
```

For constant input data, this could just be specialised with:

```c++
condition_edm_type< vecmem::const_device_array > object_in_device_code;
```

This may not pan out in the end, but this sort of a thing was why I kept the "constant types" around.

PR #36, opened 2021-03-08
Add a front-end allocator to wrap memory resources

stephenswat:
Currently, we support only allocation using memory resources via vectors and other such containers. Sometimes this is too high-level, and we (or users) may want to allocate objects individually using a memory resource. This commit adds a new allocator class which, when initialized with a memory resource, allows allocation of bytes directly, size for object arrays, or constructs new objects.

### DIFF
--- a/tests/core/test_core_allocator.cpp
+++ b/tests/core/test_core_allocator.cpp
@@ -1,0 +1,107 @@
+/*
+ * VecMem project, part of the ACTS project (R&D line)
+ *
+ * (c) 2021 CERN for the benefit of the ACTS project
+ *
+ * Mozilla Public License Version 2.0
+ */
+
+#include "vecmem/memory/allocator.hpp"
+#include "vecmem/memory/host_memory_resource.hpp"
+
+#include <gtest/gtest.h>
+
+#include <algorithm>
+#include <type_traits>
+#include <vector>
+
+class test_class {
+    public:
+    test_class() : m_int_2(11), m_bool_2(true) {}
+    test_class(int n) : m_int_2(n), m_bool_2(false) {}
+    test_class(int n, int m) : m_int_2(n), m_bool_2(m > 100) {}
+
+    int m_int_1 = 5;
+    int m_int_2;
+
+    bool m_bool_1 = false;
+    bool m_bool_2;
+};
+
+class core_allocator_test : public testing::Test {
+    protected:
+    vecmem::host_memory_resource m_upstream;
+    vecmem::allocator * m_alloc;
+
+    void SetUp() override {
+        m_alloc = new vecmem::allocator(m_upstream);
+    }
+};
+
+TEST_F(core_allocator_test, basic) {
+    void * p = m_alloc->allocate_bytes(1024);
+
+    EXPECT_TRUE(p != nullptr);
+
+    m_alloc->deallocate_bytes(p, 1024);
+}
+
+TEST_F(core_allocator_test, primitive) {
+    int * p = m_alloc->allocate_object<int>();
+
+    ASSERT_TRUE(p != nullptr);
+
+    *p = 5;
+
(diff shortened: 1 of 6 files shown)


krasznaa (commented):
Interesting...

Why did you choose to not make the class templated, but instead make every one of its functions individually templated?

From what I understand, you wanted to access some features of [std::pmr::polymorphic_allocator](https://en.cppreference.com/w/cpp/memory/polymorphic_allocator) that were only introduced with C\+\+20, right? Since the interface you introduced looks a lot like that.

But if so, I would rather do something along the following lines:

```c++
namespace vecmem {
#if __cplusplus >= 20200000
   template< typename T >
   using allocator = std::pmr::polymorphic_allocator< T >;
#else
   template< typename T >
   using allocator = details::allocator< T >;
#endif
}
```

Where `vecmem::details::allocator<T>` would be a copy of C\+\+20's version of [std::pmr::polymorphic_allocator](https://en.cppreference.com/w/cpp/memory/polymorphic_allocator).

Or did you have some explicit reason for implementing the class's templating like this?

stephenswat:
You make a good point, but I think if we examine the interface of the standard library itself, we can (I think) quite clearly see that the standards committee itself regrets their decision of making the `std::pmr::polymorphic_allocator` a templated class. Of the six methods added in the C++20 standard, two (`allocate_bytes` and `deallocate_bytes`) are type-agnostic, and the other four (`allocate_object`, `deallocate_object`, `new_object`, `delete_object`) completely ignore the template argument of the polymorphic allocator: they all rely on template parameters at the function level. The notes on cppreference say the following:

_"This function was introduced for use with the fully-specialized allocator `std::polymorphic_allocator<>`, but it may be useful in any specialization as a shortcut to avoid having to rebind from `std::polymorphic_allocator<T>` to `std::polymorphic_allocator<U>`."_

Additional evidence comes in the fact that C++20 introduces a default `std::byte` type argument for the polymorphic allocator, indicating that they want to move away from having any templates on it at all. At least, that's how I read it.

And that decision makes sense to me: templating the polymorphic allocator just adds unnecessary restrictions for the user, and it is much less flexible than having the templating at a function level. Personally I can't really see any advantage to having such an allocator class be templated at all.

So all in all, I think having this class be non-templated is actually more in line with what the standards committee wants for the polymorphic allocator.

tests/core/test_core_allocator.cpp
```diff
@@ -0,0 +1,107 @@
...
+    test_class * p2 = m_alloc->new_object<test_class>(12);
+    test_class * p3 = m_alloc->new_object<test_class>(21, 611);
+    test_class * p4 = m_alloc->new_object<test_class>(21, 15);
+
+    ASSERT_TRUE(p1 != nullptr);
+    ASSERT_TRUE(p2 != nullptr);
+    ASSERT_TRUE(p3 != nullptr);
+    ASSERT_TRUE(p4 != nullptr);
+
+    EXPECT_TRUE(p1->m_int_1 == 5);
```
krasznaa, line 88: This is just for information purposes, you don't need to change the code in this PR.

Yesterday I bumped into the `EXPECT_EQ` macro, which provides better error reporting for equality checks. So in upcoming PRs we should probably replace all such tests to use that macro instead.

krasznaa (approved):
Let's go ahead with the non-template allocator then.

At one point we could still add an extra layer on top of it, to make it behave exactly as [std::pmr::polymorphic_allocator](https://en.cppreference.com/w/cpp/memory/polymorphic_allocator) does. But we can cross that bridge when necessary...

tests/core/test_core_allocator.cpp
stephenswat: That's good to know. I'm mostly used to Catch2 for C++ testing, so I'm happy to learn about Google Test. :smile:

PR #39, opened 2021-03-09
vecmem::array Memory Handling Update, main branch (2021.03.09.)

krasznaa:
Following up from @stephenswat's comment in #37, I had to realize that the implementation of `vecmem::array` didn't allow for any complex element types.

To fix this, I updated `vecmem::array` to make use of an allocator instead of `vecmem::memory_resource` directly. Leaving it up to the allocator to properly instantiate/delete complex objects.

To do this, I now introduced 2 helper types:
  - `vecmem::details::deallocator` is a rename of the previous `vecmem::details::deleter` type. Hopefully better conveying with its name that it is only deallocating the memory given to it.
  - `vecmem::details::deleter` is not a template, making use of an allocator itself for deleting the object from memory.

Then I updated all the unit tests for these UI changes. Note that with this change the constructor of `vecmem::vector` and `vecmem::array` became a bit more consistent. :wink:

### DIFF
--- a/core/include/vecmem/containers/array.hpp
+++ b/core/include/vecmem/containers/array.hpp
@@ -10,12 +10,11 @@
 #include "vecmem/containers/details/vector_view.hpp"
 #include "vecmem/memory/resources/memory_resource.hpp"
 #include "vecmem/utils/deleter.hpp"
+#include "vecmem/utils/reverse_iterator.hpp"
 #include "vecmem/utils/types.hpp"
 
 // System include(s).
 #include <cstddef>
-#include <iterator>
-#include <memory>
 
 namespace vecmem {
 
@@ -37,7 +36,8 @@ namespace vecmem {
    /// it does provide an optional second template argument, which can be used
    /// to set the fixed size of the array at compile time.
    ///
-   template< typename T, std::size_t N = details::array_invalid_size >
+   template< typename T, std::size_t N = details::array_invalid_size,
+             typename Allocator = vecmem::polymorphic_allocator< T > >
    class array {
 
    public:
@@ -46,6 +46,8 @@ namespace vecmem {
 
       /// Type of the array elements
       typedef T                  value_type;
+      /// Type of the allocator to use
+      typedef Allocator          allocator_type;
       /// Size type for the array
       typedef std::size_t        size_type;
       /// Pointer difference type
@@ -67,9 +69,14 @@ namespace vecmem {
       typedef const_pointer_type const_iterator;
 
       /// Reverse iterator type
-      typedef std::reverse_iterator< iterator > reverse_iterator;
+      typedef vecmem::reverse_iterator< iterator > reverse_iterator;
       /// Constant reverse iterator type
-      typedef std::reverse_iterator< const_iterator > const_reverse_iterator;
+      typedef vecmem::reverse_iterator< const_iterator > const_reverse_iterator;
+
+      /// Type of the object holding on to the allocated memory
+      typedef std::unique_ptr< value_type,
+                               details::deleter< value_type, allocator_type > >
+         memory_type;
 
       /// @}
 
@@ -78,14 +85,14 @@ namespace vecmem {
       /// Can only be used if the user chose a non-default value for the size
       /// template parameter.
       ///
-      array( memory_resource& resource );
+      array( allocator_type alloc );
 
       /// Constructor with a size and a memory resource to use
       ///
       /// Can only be used if the user uses the default (invalid) value for the
       /// size template parameter.
       ///
-      array( memory_resource& resource, size_type size );
+      array( allocator_type alloc, size_type size );
 
       /// Destructor
       ~array() = default;
@@ -181,21 +188,21 @@ namespace vecmem {
       /// The size of the allocated array
       size_type m_size;
       /// The allocated array
-      std::unique_ptr< value_type, details::deleter > m_memory;
+      memory_type m_memory;
 
    }; // class array
 
    /// Helper function creating a @c vecmem::details::vector_view object
-   template< typename T, std::size_t N >
+   template< typename T, std::size_t N, typename Allocator >
    VECMEM_HOST
    details::vector_view< T >
-   get_data( array< T, N >& a );
+   get_data( array< T, N, Allocator >& a );
 
    /// Helper function creating a @c vecmem::details::vector_view object
-   template< typename T, std::size_t N >
+   template< typename T, std::size_t N, typename Allocator >
    VECMEM_HOST
    details::vector_view< const T >
-   get_data( const array< T, N >& a );
+   get_data( const array< T, N, Allocator >& a );
 
 } // namespace vecmem
 
--- a/core/include/vecmem/containers/details/vector_buffer.hpp
+++ b/core/include/vecmem/containers/details/vector_buffer.hpp
@@ -9,7 +9,7 @@
 // Local include(s).
 #include "vecmem/containers/details/vector_view.hpp"
 #include "vecmem/memory/resources/memory_resource.hpp"
-#include "vecmem/utils/deleter.hpp"
+#include "vecmem/utils/deallocator.hpp"
 #include "vecmem/utils/types.hpp"
 
 // System include(s).
@@ -39,7 +39,7 @@ namespace vecmem::details {
 
    private:
       /// Data object owning the allocated memory
-      std::unique_ptr< TYPE, deleter > m_memory;
+      std::unique_ptr< TYPE, deallocator > m_memory;
 
    }; // class vector_buffer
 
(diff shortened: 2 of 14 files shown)


core/include/vecmem/containers/array.hpp
```diff
@@ -78,14 +85,14 @@ namespace vecmem {
...
-      array( memory_resource& resource );
+      array( allocator_type alloc );
 
       /// Constructor with a size and a memory resource to use
       ///
       /// Can only be used if the user uses the default (invalid) value for the
       /// size template parameter.
       ///
-      array( memory_resource& resource, size_type size );
+      array( allocator_type alloc, size_type size );
```
krasznaa, line 95: I'm not super happy about this formalism, but this seemed to be the most appropriate.

This class only needs the allocator object in its constructor. So it could declare it with a const reference as well. But since it doesn't make a copy of the allocator itself, and the `allocate(...)` function of the allocator is non-const, I decided that making the copy as part of the function signature was the way to go.

STL containers just hide this copy a bit more. (I was looking at the implementation of `std::vector` for quite a while to figure out how it passes the received allocator object around...)

core/include/vecmem/utils/deleter.hpp
```diff
@@ -38,11 +45,14 @@ namespace vecmem::details {
...
 
    private:
-      /// The number of bytes allocated for the memory block
-      std::size_t m_bytes;
-      /// The memory resource used for deleting the memory block
-      memory_resource* m_resource;
+      /// The number of elements allocated for the memory block
+      size_type m_elements;
+      /// The allocator used to delete the elements
+      allocator_type m_allocator;
```
krasznaa, line 51: I wasn't absolutely sure about this either. But I think just making a simple copy of the allocator received in the constructor is the way to go here. But I'm open to suggestions...

krasznaa (commented):
I just thought I would point out some of the skeletons myself. :stuck_out_tongue:

core/include/vecmem/containers/array.hpp
stephenswat: Yeah, I have spent more time than I would like to admit reading through the documentation to find out how exactly `std::pmr::vector` constructs itself using a memory resource. Either something is off with the documentation there, or there is some implicit construction of allocator happening. I must admit I don't really know.

core/include/vecmem/containers/details/vector_buffer.hpp
```diff
@@ -39,7 +39,7 @@ namespace vecmem::details {
 
    private:
       /// Data object owning the allocated memory
-      std::unique_ptr< TYPE, deleter > m_memory;
+      std::unique_ptr< TYPE, deallocator > m_memory;
```
stephenswat, line 42: Why are using `deallocator` here and not `deleter`? If I understand correctly, `deleter` is now the class that actually does the destruction that you would expect it to do, right?

core/include/vecmem/utils/deleter.hpp
stephenswat: Will `m_elements` ever be something other than 1 here?

core/include/vecmem/utils/deleter.ipp
```diff
@@ -0,0 +1,24 @@
...
+   deleter< T, Allocator >::
+   deleter( size_type elements, const allocator_type& allocator )
+   : m_elements( elements ), m_allocator( allocator ) {
+
+   }
+
+   template< typename T, typename Allocator >
+   void deleter< T, Allocator >::operator()( void* ptr ) {
+
+      m_allocator.deallocate( static_cast< T* >( ptr ), m_elements );
```
stephenswat, line 21: This still doesn't work, I think. `std::allocator<T>::deallocate` states that it calls `::operator delete(void*, std::align_val_t)`, which is a [class-agnostic](https://en.cppreference.com/w/cpp/memory/new/operator_delete) variant of operator delete which does not delete.

stephenswat (changes requested):

core/include/vecmem/containers/details/vector_buffer.hpp
krasznaa: It's the same issue that we had with the "old design", and why the memory managers needed to declare whether they provide host-accessible or non-host-accessible memory.

`vecmem::details::vector_buffer` is mainly meant to be used with non-host-accessible memory. As such, `vecmem::details::deleter` would try to access memory from the host that it cannot.

I myself imagine the following things here:
  - For CUDA/HIP/SYCL "managed" memory we can let the allocation/deallocation call the constructor/destructor of non-trivial types as it wants. For non-host-accessible memory we can't do either, we can just allocate and deallocate the memory needed for those types.
  - For "simple" types, which is mostly what we should be using here anyway, this should be enough as-is.
  - For "complex" types this does put a limitation on what we can and cannot do.

In general I do not believe that any "GPU friendly" data type should ever do anything important in its destructor. These data types are not allowed to manage their own memory. So their destructor should always be a default one. (We could very well add a `static_assert` check for that...)

Having non-default constructors is another matter however. As those will be used for any operations that need to set the memory describing the object "to a ground/default state".

Now... For device-only memory, for input variables, this is not a problem. There you first need to create the objects in host memory, using their constructors. Then the memory used by these host-created objects can be copies verbatim to the device using one of the copy functions that we'll be adding in addition to the CUDA ones. After this, the objects created around that memory block in the device code should see the objects in a "reasonable state".

However for output variables this may prove a bit more tricky. Since the code currently only allocates memory for the output variables. It doesn't initialise that variable in any shape or form. The `vecmem::device_vector` and `vecmem::device_array` types would need to learn how to use the constructors of the types that they describe to initialise the memory given to them.

But... Very long story short, the `vecmem::details::vector_buffer` type, in my mind, must only do memory allocations and de-allocations. It must not interact with the types that would eventually use these allocated memory blocks...

core/include/vecmem/containers/array.hpp
```diff
@@ -37,7 +36,8 @@ namespace vecmem {
    /// it does provide an optional second template argument, which can be used
    /// to set the fixed size of the array at compile time.
    ///
-   template< typename T, std::size_t N = details::array_invalid_size >
+   template< typename T, std::size_t N = details::array_invalid_size,
+             typename Allocator = vecmem::polymorphic_allocator< T > >
```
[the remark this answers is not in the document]
krasznaa: `vecmem::array` is meant to be able to work on top of memory allocated through `vecmem::memory_resource`. That's why I chose to use `vecmem::polymorphic_allocator`. As I hoped that it would take care of constructing/destructing the array elements properly, even for complex types.

Using the full-blown `vecmem::polymorphic_allocator` is indeed a bit of an overkill. Since a `vecmem::array` object will issue exactly one allocation and one de-allocation call to the allocator. I just wanted to inherit the logic for doing this properly for complex types from `vecmem::polymorphic_allocator`.

core/include/vecmem/utils/deleter.hpp
krasznaa: It's used by `vecmem::array`. (At the moment. It may see usage outside of it later on as well.) So `m_elements` is meant to be the size of the array. (Defined either at compile, or at runtime. Remember that the size of `vecmem::array` can also be defined at runtime, since it needs to use `vecmem::memory_resource` for the allocation anyway.)

core/include/vecmem/containers/array.hpp
krasznaa: Note that now that `vecmem::array` uses an allocator instead of a memory resource directly, I even more strongly think that the order of these two constructor arguments should be reversed. So that it would provide the same interface as `vecmem::vector` when the size of the array is decided at runtime.

I was convinced that either you or @paulgessinger requested this argument order in #25. But I can't find that comment anymore. :confused:

core/include/vecmem/utils/deleter.ipp
stephenswat: I'm afraid not. The [documentation for `std::pmr::polymorphic_allocator<T>::deallocate`](https://en.cppreference.com/w/cpp/memory/polymorphic_allocator/deallocate) states that it is equivalent to `this->resource()->deallocate(p, n * sizeof(T), alignof(T));`. That doesn't do any destruction. What you want is [`std::pmr::polymorphic_allocator<T>::delete_object`](https://en.cppreference.com/w/cpp/memory/polymorphic_allocator/delete_object), but that's a C++20 feature... :disappointed:

core/include/vecmem/containers/array.hpp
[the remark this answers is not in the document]
stephenswat: I'm realize now that the answer is (probably) arrays with dynamic size.

krasznaa:
As discussed with Stephen, let's try this in a different way. (In a different PR...)

PR #41, opened 2021-03-10
Memory Handling Update, main branch (2021.03.10.)

krasznaa:
This is meant as a replacement for #39.

Instead of bothering with allocators, `vecmem::array` is taught directly to call the constructors of non-trivial types, and is updated to use its own custom deleter type once again to make sure that custom destructors would be called correctly.

At the same time renamed `vecmem::details::deleter` to `vecmem::details::deallocator`, as the latter better reflects what this type does. It just deallocates memory, it is not able to call any custom destructors. (Since it is used for non-host-accessible memory resources as well.)

Updated both `vecmem::array` and `vecmem::details::vector_buffer` to explicitly require certain properties from their value types.

Finally, added some tests to make sure that both `vecmem::vector` and `vecmem::array` would call the constructor and destructor of its elements correctly. (For `vecmem::vector` it's a bit unnecessary, but it could come in handy if we do some bigger changes to how `vecmem::vector` is defined.)

### DIFF
--- a/core/CMakeLists.txt
+++ b/core/CMakeLists.txt
@@ -27,6 +27,8 @@ vecmem_add_library( vecmem_core core SHARED
    "include/vecmem/memory/allocator.hpp"
    "include/vecmem/memory/allocator.ipp"
    "src/memory/allocator.cpp"
+   "include/vecmem/memory/deallocator.hpp"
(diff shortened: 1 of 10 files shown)


stephenswat (approved):
This looks great. :smile:

PR #42, opened 2021-03-11
Forward declare C++17 types for NVCC

stephenswat:
Once again NVCC's dependence on C++14 rears its ugly head. We can't really include the memory_resources.hpp file from any code that might be included from CUDA right now, but we need to do that for some new types. This commit adds a preprocessor clause that forward-declares the memory_resource and polymorphic_allocator types in code compiled by NVCC. That allows NVCC to accept the type for pointers and references without having to include more C++17-only headers.

### DIFF
--- a/core/include/vecmem/memory/resources/memory_resource.hpp
+++ b/core/include/vecmem/memory/resources/memory_resource.hpp
@@ -14,7 +14,14 @@
  * std::experimental::pmr namespace depending on the GCC version used, so we try
  * to unify them by aliassing depending on the compiler feature flags.
  */
-#if __has_include(<memory_resource>)
+#ifdef __CUDACC__
+namespace vecmem {
+    class memory_resource;
+
+    template<typename T>
+    class polymorphic_allocator;
+}
+#elif __has_include(<memory_resource>)
 #include <memory_resource>
 
 namespace vecmem {


core/include/vecmem/memory/resources/memory_resource.hpp
```diff
@@ -14,7 +14,14 @@
...
  */
-#if __has_include(<memory_resource>)
+#ifdef __CUDACC__
+namespace vecmem {
+    class memory_resource;
+
+    template<typename T>
+    class polymorphic_allocator;
+}
+#elif __has_include(<memory_resource>)
```
krasznaa, line 24: :thinking: I wonder if we couldn't make this a bit more robust...

How about introducing a `memory_resource_fwd.hpp` file that only has these forward declarations? And then we use that everywhere where we don't need to see the full definitions. Regardless whether CUDA or whatever is being used for the compilation.

In the `jagged_vector_view.hpp` file then you could also hide the constructor needing this type using `__cplusplus` instead of `__CUDACC__`.

I'm just very much on the mind now, thanks to you, that depending on the language specific macros should be avoided as much as possible...

stephenswat: We could alternatively have the check be `#if __cplusplus < 201703L` which might be more robust. Would that be okay for you?

krasznaa: I don't think that would work in this file. :thinking:

My problem is that I don't like that a given header would either provide the full definition of a type or just a forward declaration of it, based on some circumstances. I think that's not a good design. :frowning:

I believe it would be cleaner if we coded this up like:

`jagged_vector_view.hpp`:
```c++
#include "vecmem/memory/resources/memory_resource_fwd.hpp"
...
      jagged_vector_view( memory_resource& resource );
...
```

`jagged_vector_view.ipp`
```c++
#if __cplusplus >= 201703L
#   include "vecmem/memory/resources/memory_resource.hpp"
#endif
...
template<...>
jagged_vector_view<...>::jagged_vector_view( memory_resource& resource ) {
#if __cplusplus >= 201703L
   // Do whatever needs to be done.
#else
   static_assert( false, "This constructor is not available with <C++17" );
#endif
}
```

stephenswat: I understand your hesitation to conditionally provide a forward declaration, but unfortunately I don't think your solution would work either. Typedeffing a class after forward-declaring it is a compilation error, so if we were to go this way we would need to make absolutely sure that no file ever includes, directly or indirectly, both `memory_resource.hpp` and `memory_resource_fwd.hpp`. That makes developing the library as well as using it really difficult. :disappointed:

And I really do agree with you that I think this is not great, but it's the only way I have come up with so far that makes my jagged vector code work.

krasznaa: Okay... How about we switch to declaring `vecmem::memory_resource` like this then?

```c++
namespace vecmem {
   class memory_resource : public std::pmr::memory_resource {
   public:
      using std::pmr::memory_resource::memory_resource;
   };
}
```

(With the additional logic for "older" compilers of course.) Then we could freely forward declare it as we'd like.

stephenswat: The problem with that is that we would have to do the same for `std::pmr::polymorphic_allocator` which, unlike `std::pmr::memory_resource`, is not a virtual class designed to be subclassed. And we all know that inheriting from `std` classes not designed for that is a dangerous idea.

stephenswat: I guess we could split the `memory_resource.hpp` file in two, one for the `memory_resource` class and one for the `polymorphic_allocator` class. I can give it a try, but I worry we might be getting ourselves in a really hairy situation.

krasznaa: I was wondering about that since a while as well. I understand that since STL put both types into [<memory_resource>](https://en.cppreference.com/w/cpp/header/memory_resource), you did the same in our header.

For following the ATLAS coding guidelines, I would actually prefer to have 2 separate headers for `vecmem::memory_resource` and `vecmem::polymorphic_allocator<...>`...

Could you give it a try? I agree that it's not a super-clean, super-nice setup. But I really fear adding conditional compilation into the core library based on `__CUDACC__`. (The types in `types.hpp` don't count. :stuck_out_tongue:)

stephenswat:
Re-opening, because after some investigation the subclassic approach isn't going to work for `vecmem::memory_resource`: the `std::pmr::memory_resource` has a `do_is_equal` method, which requires a `std::pmr::memory_resource`, not a `vecmem::memory_resource`. That means that each memory resource must know about the `std::pmr::memory_resource`, but also the `vecmem` version...

This is making my head spin! :confounded:

krasznaa:
Ahh, yeah... Of course STL would put its own spin on its real class names... :frowning:

```
In file included from /home/krasznaa/ATLAS/projects/vecmem/vecmem/tests/core/test_core_arrays.cpp:9:
In file included from /home/krasznaa/ATLAS/projects/vecmem/vecmem/core/include/vecmem/containers/array.hpp:240:
In file included from /home/krasznaa/ATLAS/projects/vecmem/vecmem/core/include/vecmem/containers/array.ipp:10:
/home/krasznaa/ATLAS/projects/vecmem/vecmem/core/include/vecmem/memory/memory_resource.hpp:27:48: error: reference to 'pmr' is ambiguous
    using memory_resource = std::experimental::pmr::memory_resource;
                            ~~~~~~~~~~~~~~~~~~~^
/home/krasznaa/ATLAS/projects/vecmem/vecmem/core/include/vecmem/memory/memory_resource_fwd.hpp:21:52: note: candidate found by name lookup is 'std::experimental::pmr'
namespace std { namespace experimental { namespace pmr {
                                                   ^
/opt/lcg/gcc/8.3.0-eda0e/x86_64-centos7/lib/gcc/x86_64-pc-linux-gnu/8.3.0/../../../../include/c++/8.3.0/experimental/memory_resource:43:11: note: candidate found by name lookup is 'std::experimental::fundamentals_v2::pmr'
namespace pmr {
          ^
```

I thought I could maybe get away with forward declaring the typename in the STL namespace, and then add a `using` statement for it. But that can also only be done if we're willing to hardcode implementation details like that into our code. Which I am just not...

This sort of comes back to #30. I introduced `vecmem::sycl::details::opaue_queue` exactly to prepare for these sort of shenanigans... :confused:

The problem is that that solution also required inheriting from the upstream type. Which I have to admit I still don't fully understand is not possible with `std::pmr::memory_resource`. :confused: Which implementation of `do_is_equal` would be confused by that exactly?

stephenswat:
Well, the problem I see is that, to declare a custom memory resource, you need the full definition of the `vecmem::memory_resource` (to inherit from), as well as at least a partial definition of `std::pmr::memory_resource` (to declare the `do_is_equal` method). Right now that works fine because those are the same class. If we're going to start subclassing, we would need to have another definition for the base `std::pmr::memory_resource`.

Anyway, I got really confused by this yesterday but it's a fresh new day, so I'll see if I can do something clever.

stephenswat:
Alright, the problems continue. My declaration of `jagged_vector_view` requires the compiler to know about at least a forward declaration of `vecmem::vector`. That gives us two options:

1. Import the full `vecmem::vector` header in `jagged_vector_view.hpp`
2. Import a slimmed down `vector_fwd.hpp` forward declaration header in `jagged_vector_view.hpp`

And of course neither of those work. In the first case, it straight up doesn't compile because CUDA doesn't know how to compile vectors with polymorphic allocators. In the second case, we run into the issue that you can't use a typedef to fill a forward declaration, so you again need to make sure that no file ever includes both `vector.hpp` and `jagged_vector_view.hpp` which is just a silly requirement. :angry:

PR #43, opened 2021-03-11
Split memory resource header in two parts

stephenswat:
Inspired by the discussion in #42, let's split the `memory_resource.hpp` file into two parts, one for `memory_resource` and one for `polymorphic_allocator`. I've also taken the liberty of moving the `memory_resource.hpp` up one level so it doesn't sit all by itself in the `resources/` directory anymore.


krasznaa (approved):
Looks good to me!

PR #44, opened 2021-03-11
Static Vector Update, main branch (2021.03.11.)

krasznaa:
I decided that while you work on the jagged vector concept @stephenswat, I would do some house cleaning in `vecmem::static_vector`. With 2 goals in mind:
  - Make it handle custom (non-trivial) constructors and destructors on the types that it wraps;
  - Implement the rest of `std::vector` in it. At least the parts that make sense.

The main feature of the update is that the vector now uses a `char` array for storing the payload of its elements. This is to avoid the requirement of having a default constructor on the template type.

The code is for sure not bug-free at this point. It just has too many memory operations for me not to have made any mistakes in them. But I wanted to open a PR at this point already, to let you comment on the overall ideas.

Note that I added a new test that allows me to test different features of `vecmem::static_vector` for different template types through GoogleTest. Unfortunately the way it is implemented in GoogleTest required me to disable a GNU specific warning in the build. (Otherwise the Debug build would fail.) Still, this automated testing for different template types should come in handy in some of our other tests as well. :wink:

### DIFF
--- a/core/include/vecmem/containers/static_vector.hpp
+++ b/core/include/vecmem/containers/static_vector.hpp
@@ -7,170 +7,271 @@
 #pragma once
 
 // Local include(s).
+#include "vecmem/utils/reverse_iterator.hpp"
 #include "vecmem/utils/types.hpp"
 
 // System include(s).
-#include <algorithm>
-#include <cassert>
-#include <vector>
+#include <cstddef>
 
 namespace vecmem {
 
    /// Class mimicking @c std::vector on top of a fixed sized array
+   ///
+   /// This can come in handy when needing vector arithmetics in device code,
+   /// without resorting to heap allocations.
+   ///
+   /// The type does come with a significant limitation over @c std::vector.
+   /// It has a maximal/fixed size that needs to be chosen at compile time.
+   ///
    template< typename TYPE, std::size_t MAX_SIZE >
    class static_vector {
 
    public:
-      /// @name Type definitions mimicking @c std::vector
+      /// @name Type definitions, mimicking @c std::vector
       /// @{
-      typedef std::size_t    size_type;
-      typedef std::ptrdiff_t difference_type;
-      typedef TYPE*          pointer;
-      typedef const TYPE*    const_pointer;
-      typedef TYPE&          reference;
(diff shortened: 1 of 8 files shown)


core/include/vecmem/containers/static_vector.hpp
```diff
@@ -7,170 +7,271 @@
...
+      static_vector();
+      /// Construct a vector with a specific size
+      VECMEM_HOST_AND_DEVICE
+      static_vector( size_type size, const_reference value = value_type() );
+      /// Construct a vector with values coming from a pair of iterators
+      /*
+      template< typename InputIt >
+      VECMEM_HOST_AND_DEVICE
+      static_vector( InputIt other_begin, InputIt other_end );
+      */
```
krasznaa, line 81: Do you know what's the best way to test that `InputIt` is "and iterator type"?

https://en.cppreference.com/w/cpp/container/vector/vector

describes the exact problem that I ran into with this. That without any type requirements, this constructor is preferred over the one defined just before it.

Unfortunately STL implements its type requirements with non-public code. :frowning: So for now I rather just removed these lines until I get a better idea of how to put in a bit of SFINAE on this function...

stephenswat: I'm not sure what you mean. If you leave the second constructor uncommented it will implicitly upcast and use that one instead of the one on line 75? In that case, you should make the constructor of InputIt explicit.

krasznaa:
This was surprising... I ran into a limitation/bug in GCC with this code. :confused: The iterator-using-functions, that I talked about earlier, have to be implemented inline. (6424d9d) Otherwise GCC gets very confused about them. (Or at least I only found this one solution for making GCC accept this code.)

Since I was developing the code using Clang, this only broke in the CI. This was quite unexpected...

krasznaa:
Okay, I'm happy with the current state of the code. I didn't create any initializer list constructors or member functions, but we should be able to live without those for now. (They can be added later on when/if necessary.)

krasznaa:
Since this whole PR is just an "implementation detail", let me just go forward with it. It should anyway not affect any of the "important" design questions on the table at the moment...

PR #46, opened 2021-03-13
Core Test Cleanup, main branch (2021.03.13.)

krasznaa:
This PR updates the core library's tests to make use of parametrised tests where appropriate/possible, and overall make the naming of the tests and source files a bit more consistent.

A couple of the core tests can meaningfully be executed using different memory resources. Making it potentially possible to re-use that same code for testing (host-accessible) GPU memory resources. So I re-wrote the code to use `testing::TestWithParam` in those cases.

At the same time introduced some code for making the printouts coming from GTest a little nicer. Which I admit, may have been overdoing things a little...

Note that the `vecmem_testing_common` static library is **not** set up to be installed with the project. It is kept only in the build directory, and used only by the tests. (It is a static library because GTest is by default a non-position-independent static library as well.)


stephenswat:
Looks good, but I don't know enough about GTest to provide any meaningful comments, I'm afraid.

krasznaa:
Unfortunately neither do I. :frowning: For instance I also played a bit with the CUDA tests since (I'll open a separate PR about those when this goes in), but picking up "test fixtures" from the core library tests to use in the CUDA library tests will need a bit more of a build re-organisation.

So on first order I won't be shooting for that. I'll just be migrating the "device code tests" to GTest as well, in fairly simple ways.

PR #47, opened 2021-03-15
CUDA Test Cleanup, main branch (2021.03.15.)

krasznaa:
This is a follow-up from #46.

Migrated the `vecmem::cuda` tests to GoogleTest, merging all of the test code into a single executable, and re-writing it to use GTest fixtures and tests.

For now simply copied the host-accessible memory resource tests from the core library's tests. Later on these could be set up to be shared between the tests.


stephenswat (approved):

krasznaa:
I guess you know that this will mess with #45. :frowning: But what can we do...

PR #49, opened 2021-03-15
SYCL Test Cleanup, main branch (2021.03.15.)

krasznaa:
Updated the SYCL tests in the same way as what I did in #47 and #48.

The one addition is that I had to set a high "linker preference" (https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_LINKER_PREFERENCE.html) for the SYCL language. By default CMake used "the C\+\+ linker" for linking any binaries that have C\+\+ and any sort of other code in them. This works well for CUDA and HIP, as they both explicitly link against the CUDA/HIP runtime library. So "C\+\+ linking" works out fine for the test executables.

But since the new `vecmem_test_sycl` executable contains SYCL device code, the way that the oneAPI linker works, one must provide the `-fsycl` flag to the linker. Which is done by linking the test as a "SYCL binary".

Unfortunately RPATH handling is still not working well for "SYCL linking". :frowning: But I'll have to look at that in a separate PR.


stephenswat (approved):
Looks fine, sadly I don't know enough about SYCL linking to comment, sadly.

PR #52, opened 2021-03-24
Jagged Vector Data Updates, main branch (2021.03.24.)

krasznaa:
In this PR I re-designed `vecmem::data::jagged_vector_view` and `vecmem::data::jagged_vector_data` a bit.

In the base class I only made some stylistic changes, but left the design the same as it was. (Since there's little that one can change in that class. :stuck_out_tongue:)

However I re-wrote `vecmem::data::jagged_vector_data` not to depend on `vecmem::jagged_vector` directly anymore. Instead it is created using primitive variables, leaving it up to the newly introduced `vecmem::get_data(...)` overloads to construct proper(ly filled) objects.

At the same time I also modified `vecmem::data::jagged_vector_data`'s memory management to use [std::unique_ptr](https://en.cppreference.com/w/cpp/memory/unique_ptr). Memory ownership in this type, and the `vecmem::data::jagged_vector_buffer` that I plan to introduce next, is just a bit iffy. I hope that by using [std::unique_ptr](https://en.cppreference.com/w/cpp/memory/unique_ptr), we'll get notified more clearly about trying to use unsafe copy/move operations.

The tests in the end only had to be adjusted slightly to account for these changes.

### DIFF
--- a/tests/cuda/test_cuda_jagged_vector_view.cpp
+++ b/tests/cuda/test_cuda_jagged_vector_view.cpp
@@ -12,7 +12,6 @@
 #include "vecmem/containers/vector.hpp"
 #include "vecmem/containers/jagged_vector.hpp"
 #include "vecmem/containers/data/jagged_vector_data.hpp"
-#include "vecmem/containers/jagged_device_vector.hpp"
 
 #include <gtest/gtest.h>
 
@@ -33,7 +32,7 @@ class cuda_jagged_vector_view_test : public testing::Test {
             vecmem::vector<int>(&m_mem),
             vecmem::vector<int>({12, 13, 14, 15, 16}, &m_mem)
         }, &m_mem),
-        m_data(m_vec, &m_mem)
+        m_data(vecmem::get_data(m_vec, &m_mem))
     {
     }
 };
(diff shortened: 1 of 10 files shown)


tests/cuda/test_cuda_jagged_vector_view.cpp
```diff
@@ -33,30 +32,28 @@ class cuda_jagged_vector_view_test : public testing::Test {
...
+    EXPECT_EQ(m_vec.at(2).at(0), 14);
+    EXPECT_EQ(m_vec.at(2).at(1), 16);
+    EXPECT_EQ(m_vec.at(2).at(2), 18);
+    EXPECT_EQ(m_vec.at(2).at(3), 20);
+    EXPECT_EQ(m_vec.at(3).at(0), 22);
+    EXPECT_EQ(m_vec.at(5).at(0), 24);
+    EXPECT_EQ(m_vec.at(5).at(1), 26);
+    EXPECT_EQ(m_vec.at(5).at(2), 28);
+    EXPECT_EQ(m_vec.at(5).at(3), 30);
+    EXPECT_EQ(m_vec.at(5).at(4), 32);
```
krasznaa, line 58: After some pondering I decided that it makes more sense to check the values in the "original" jagged vector instead of looking at the contents of `m_data` using another instances of `vecmem::jagged_device_vector`. But let me know if you think otherwise @stephenswat.

stephenswat: Works for me, as long as we design everything such that the underlying `std::vector` is the ultimate owner of all the data, and that they are responsible for freeing all of it. That's a fairly natural design anyway, in this case, because `std::vector` doesn't really allow you to release its memory from it. Not without some fairly unpleasant code, anyway.

krasznaa: The code already tests `vecmem::jagged_device_vector`'s behaviour in `test_core_jagged_vector_view.cpp`. In this CUDA test I believe we want to test that the CUDA kernel is able to modify the values inside of the "original" `vecmem::jagged_vector` in-situ. That's why I thought the tests would make more sense expressed through the `m_vec` object.

Any other comments about these proposed changes in `vecmem::data::jagged_vector_data`? :thinking:

krasznaa:
I realise this is not too collaborative, but we really need to soldier on with this... (I have changes waiting for `vecmem::jagged_device_vector` that depend on these changes.)

PR #54, opened 2021-03-25
vecmem::jagged_device_vector Update, main branch (2021.03.25.)

krasznaa:
Okay, with this you really need to hold on to your hat... :stuck_out_tongue:

After some pondering I decided to use a hack for representing jagged vectors in a nice way in device code. Relying on the fact that `vecmem::data::vector_view<T>` and `vecmem::device_vector<T>` have the same data layout in memory. So, as ugly/scary as it seems, the following works:

```c++
vecmem::data::vector_view<T>* ptr1 = ...;
vecmem::device_vector<T>* ptr2 = reinterpret_cast< vecmem::device_vector<T>* >( ptr1 );
```

I was thinking of other solutions for a while about how `vecmem::jagged_device_vector` should present a nice vector-of-vectors interface, but everything else would've been way more expensive. Since remember, allocating memory on the heap in device code should only be done as a last resort.

I added some runtime tests in `test_core_device_containers.cpp` to make sure that this conversion would be valid. Though we may want to turn that into a compile-time check later on as well. :thinking:

I'm really curious about what you think on this one...

### DIFF
--- a/core/include/vecmem/containers/details/jagged_device_vector_iterator.hpp
+++ b/core/include/vecmem/containers/details/jagged_device_vector_iterator.hpp
@@ -14,7 +14,7 @@
 
 // System include(s).
 #include <cstddef>
-#include <type_traits>
+#include <iterator>
 
 namespace vecmem { namespace details {
 
@@ -49,10 +49,39 @@ namespace vecmem { namespace details {
       typedef device_vector< TYPE > value_type;
       /// (Pointer) Difference type
       typedef std::ptrdiff_t difference_type;
-      /// Pointer type to the underlying (virtual) value
-      typedef value_type* pointer;
-      /// Reference type to the underlying (virtual) value
-      typedef value_type& reference;
+      /// "Reference" type to the underlying (virtual) value
+      typedef value_type reference;
+
+      /// Helper class for returning "pointer-like" objects from the iterator
+      ///
+      /// Since the iterator returned everything by value as temporary objects,
+      /// in order to provide a proper return type for its @c operator->, this
+      /// custom type needs to be used.
+      ///
+      class pointer {
+
+      public:
+         /// Constructor from a data pointer
+         ///
+         /// Used a pointer instead of a reference to make the rest of the code
+         /// in @c vecmem::details::jagged_device_vector_iterator as unaware of
+         /// the existence of this type as possible.
+         ///
+         VECMEM_HOST_AND_DEVICE
+         pointer( const data_pointer data );
+
+         /// Return a pointer to a device vector (non-const)
+         VECMEM_HOST_AND_DEVICE
+         value_type* operator->();
+         /// Return a pointer to a device vector (const)
+         VECMEM_HOST_AND_DEVICE
+         const value_type* operator->() const;
+
+      private:
+         /// Temporary device vector created on the stack
+         value_type m_vec;
+
+      }; // class pointer
 
       /// @}
 
@@ -62,13 +91,6 @@ namespace vecmem { namespace details {
       /// Constructor from an underlying data object
       VECMEM_HOST_AND_DEVICE
       jagged_device_vector_iterator( data_pointer data );
-      /// Construct from a non-const underlying data object
-      template< typename OTHERTYPE,
-                std::enable_if_t<
-                   details::is_same_nc< TYPE, OTHERTYPE >::value,
-                   bool > = true >
-      VECMEM_HOST_AND_DEVICE
-      jagged_device_vector_iterator( data::vector_view< OTHERTYPE >* data );
       /// Copy constructor
       VECMEM_HOST_AND_DEVICE
       jagged_device_vector_iterator(
@@ -133,22 +155,17 @@ namespace vecmem { namespace details {
       /// @{
 
       /// Check for the equality of two iterators
+      VECMEM_HOST_AND_DEVICE
       bool operator==( const jagged_device_vector_iterator& other ) const;
       /// Check for the inequality of two iterators
+      VECMEM_HOST_AND_DEVICE
       bool operator!=( const jagged_device_vector_iterator& other ) const;
 
       /// @}
 
    private:
-      /// Ensure that the internal value object is valid
-      void ensure_valid() const;
-
       /// Pointer to the data (in an array)
       data_pointer m_ptr;
-      /// Flag showing whether the value object is up to date with the pointer
-      mutable bool m_value_is_valid;
-      /// Helper object returned by the iterator
-      mutable value_type m_value;
 
    }; // class jagged_device_vector_iterator
 
@@ -176,6 +193,7 @@ namespace std {
       typedef typename
          vecmem::details::jagged_device_vector_iterator< T >::reference
          reference;
+      typedef std::random_access_iterator_tag iterator_category;
    };
 
 } // namespace std
--- a/core/include/vecmem/containers/details/jagged_device_vector_iterator.hpp
+++ b/core/include/vecmem/containers/details/jagged_device_vector_iterator.hpp
@@ -1,0 +1,184 @@
+/*
+ * VecMem project, part of the ACTS project (R&D line)
+ *
+ * (c) 2021 CERN for the benefit of the ACTS project
+ *
+ * Mozilla Public License Version 2.0
+ */
+#pragma once
+
+// Local include(s).
+#include "vecmem/containers/data/vector_view.hpp"
+#include "vecmem/containers/device_vector.hpp"
+#include "vecmem/utils/types.hpp"
+
+// System include(s).
+#include <cstddef>
+#include <type_traits>
+
+namespace vecmem { namespace details {
+
+   /// Custom iterator for @c vecmem::jagged_device_vector
+   ///
+   /// In order for @c vecmem::jagged_device_vector to be able to offer
+   /// iteration over its elements in an efficient and safe way, it needs to use
+   /// this custom iterator type.
+   ///
+   /// It takes care of converting between the underlying data type and the
+   /// type presented towards the users for access to the data. On top of
+   /// providing all the functionality that an iterator has to.
+   ///
+   template< typename TYPE >
+   class jagged_device_vector_iterator {
+
+   public:
+      /// @name Types describing the underlying data
+      /// @{
+
+      /// Type of the data object that we have an array of
+      typedef data::vector_view< TYPE > data_type;
+      /// Pointer to the data object
+      typedef data_type* data_pointer;
+
+      /// @}
+
+      /// @name Type definitions, mimicking STL iterators
+      /// @{
+
+      /// Value type being (virtually) iterated on
+      typedef device_vector< TYPE > value_type;
+      /// (Pointer) Difference type
+      typedef std::ptrdiff_t difference_type;
+      /// Pointer type to the underlying (virtual) value
+      typedef value_type* pointer;
+      /// Reference type to the underlying (virtual) value
+      typedef value_type& reference;
+
+      /// @}
+
+      /// Default constructor
+      VECMEM_HOST_AND_DEVICE
+      jagged_device_vector_iterator();
+      /// Constructor from an underlying data object
+      VECMEM_HOST_AND_DEVICE
+      jagged_device_vector_iterator( data_pointer data );
+      /// Construct from a non-const underlying data object
+      template< typename OTHERTYPE,
+                std::enable_if_t<
+                   details::is_same_nc< TYPE, OTHERTYPE >::value,
+                   bool > = true >
+      VECMEM_HOST_AND_DEVICE
+      jagged_device_vector_iterator( data::vector_view< OTHERTYPE >* data );
+      /// Copy constructor
+      VECMEM_HOST_AND_DEVICE
+      jagged_device_vector_iterator(
+         const jagged_device_vector_iterator& parent );
+      /// Copy constructor
+      template< typename T >
+      VECMEM_HOST_AND_DEVICE
+      jagged_device_vector_iterator(
+         const jagged_device_vector_iterator< T >& parent );
+
+      /// Copy assignment operator
+      VECMEM_HOST_AND_DEVICE
+      jagged_device_vector_iterator&
+      operator=( const jagged_device_vector_iterator& rhs );
+
+      /// @name Value accessor operators
+      /// @{
+
+      /// De-reference the iterator
+      VECMEM_HOST_AND_DEVICE
+      reference operator*() const;
+      /// Use the iterator as a pointer
+      VECMEM_HOST_AND_DEVICE
+      pointer operator->() const;
+
+      /// @}
+
+      /// @name Iterator updating operators
+      /// @{
+
+      /// Decrement the underlying iterator (with '++' as a prefix)
+      VECMEM_HOST_AND_DEVICE
+      jagged_device_vector_iterator& operator++();
+      /// Decrement the underlying iterator (wuth '++' as a postfix)
+      VECMEM_HOST_AND_DEVICE
+      jagged_device_vector_iterator operator++( int );
+
+      /// Increment the underlying iterator (with '--' as a prefix)
+      VECMEM_HOST_AND_DEVICE
+      jagged_device_vector_iterator& operator--();
+      /// Increment the underlying iterator (with '--' as a postfix)
+      VECMEM_HOST_AND_DEVICE
+      jagged_device_vector_iterator operator--( int );
+
+      /// Decrement the underlying iterator by a specific value
+      VECMEM_HOST_AND_DEVICE
+      jagged_device_vector_iterator operator+( difference_type n ) const;
+      /// Decrement the underlying iterator by a specific value
+      VECMEM_HOST_AND_DEVICE
+      jagged_device_vector_iterator& operator+=( difference_type n );
+
+      /// Increment the underlying iterator by a specific value
+      VECMEM_HOST_AND_DEVICE
+      jagged_device_vector_iterator operator-( difference_type n ) const;
+      /// Increment the underlying iterator by a specific value
+      VECMEM_HOST_AND_DEVICE
+      jagged_device_vector_iterator& operator-=( difference_type n );
+
+      /// @}
+
+      /// @name Comparison operators
+      /// @{
+
+      /// Check for the equality of two iterators
+      bool operator==( const jagged_device_vector_iterator& other ) const;
+      /// Check for the inequality of two iterators
+      bool operator!=( const jagged_device_vector_iterator& other ) const;
+
+      /// @}
+
+   private:
+      /// Ensure that the internal value object is valid
+      void ensure_valid() const;
+
+      /// Pointer to the data (in an array)
+      data_pointer m_ptr;
+      /// Flag showing whether the value object is up to date with the pointer
+      mutable bool m_value_is_valid;
+      /// Helper object returned by the iterator
+      mutable value_type m_value;
+
+   }; // class jagged_device_vector_iterator
+
+} } // namespace vecmem::details
+
+namespace std {
+
+   /// Specialisation of @c std::iterator_traits
+   ///
+   /// This is necessary to make @c vecmem::reverse_iterator functional on top
+   /// of @c vecmem::details::jagged_device_vector_iterator.
+   ///
+   template< typename T >
+   struct
+   iterator_traits< vecmem::details::jagged_device_vector_iterator< T > > {
+      typedef typename
+         vecmem::details::jagged_device_vector_iterator< T >::value_type
+         value_type;
+      typedef typename
+         vecmem::details::jagged_device_vector_iterator< T >::difference_type
+         difference_type;
+      typedef typename
+         vecmem::details::jagged_device_vector_iterator< T >::pointer
+         pointer;
+      typedef typename
+         vecmem::details::jagged_device_vector_iterator< T >::reference
+         reference;
+   };
+
+} // namespace std
+
+// Include the implementation.
+#include "vecmem/containers/impl/jagged_device_vector_iterator.ipp"
--- a/core/CMakeLists.txt
+++ b/core/CMakeLists.txt
@@ -48,6 +48,8 @@ vecmem_add_library( vecmem_core core SHARED
    "src/memory/contiguous_memory_resource.cpp"
    "include/vecmem/memory/contiguous_memory_resource.hpp"
    # Utilities.
+   "include/vecmem/containers/details/jagged_device_vector_iterator.hpp"
+   "include/vecmem/containers/impl/jagged_device_vector_iterator.ipp"
    "include/vecmem/utils/reverse_iterator.hpp"
    "include/vecmem/utils/reverse_iterator.ipp"
    "include/vecmem/utils/type_traits.hpp"
--- a/core/include/vecmem/containers/data/jagged_vector_view.hpp
+++ b/core/include/vecmem/containers/data/jagged_vector_view.hpp
@@ -10,6 +10,7 @@
 
 // Local include(s).
 #include "vecmem/containers/data/vector_view.hpp"
+#include "vecmem/utils/type_traits.hpp"
 #include "vecmem/utils/types.hpp"
 
 // System include(s).
@@ -71,9 +72,7 @@ namespace vecmem { namespace data {
          */
         template< typename OTHERTYPE,
                   std::enable_if_t<
-                     ( ! std::is_same< T, OTHERTYPE >::value ) &&
-                     std::is_same< T,
-                                   typename std::add_const< OTHERTYPE >::type >::value,
+                     details::is_same_nc< T, OTHERTYPE >::value,
                      bool > = true >
         VECMEM_HOST_AND_DEVICE
         jagged_vector_view( const jagged_vector_view< OTHERTYPE >& parent );
--- a/core/include/vecmem/containers/data/vector_view.hpp
+++ b/core/include/vecmem/containers/data/vector_view.hpp
@@ -7,6 +7,7 @@
 #pragma once
 
 // Local include(s).
+#include "vecmem/utils/type_traits.hpp"
 #include "vecmem/utils/types.hpp"
 
 // System include(s).
@@ -48,9 +49,7 @@ namespace vecmem { namespace data {
       ///
       template< typename OTHERTYPE,
                 std::enable_if_t<
-                   ( ! std::is_same< TYPE, OTHERTYPE >::value ) &&
-                   std::is_same< TYPE,
-                                 typename std::add_const< OTHERTYPE >::type >::value,
+                   details::is_same_nc< TYPE, OTHERTYPE >::value,
                    bool > = true >
       VECMEM_HOST_AND_DEVICE
       vector_view( const vector_view< OTHERTYPE >& parent );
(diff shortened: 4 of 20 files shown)


stephenswat:
Hmm, I'm afraid I really can't sign off on this one. The fact that this works on your compiler right now doesn't mean it works on other compilers, or that it will keep working in the future. `reinterpret_cast` provides essentially zero guarantees according to the C++ standard. The only thing that is guaranteed is that for all _T_, _U_, and _v_ is of type _T_: `reinterpret_cast<T *>(reinterpret_cast<U *>(&v)) == &v`. Nothing else. We're relying on undefined behavior here, and I don't think this has any place in a library.

What's the issue with creating `vector_view`s on the fly as we are doing right now? The overhead of that should already be quite low. Otherwise, we might be able to get by with giving `vector_view` and `device_vector` a shared mixin parent class.

krasznaa:
Creating the objects on the fly doesn't work with using iterators. At least not as long as we don't want to do something really expensive with the iterators. (Defining a custom iterator type, which would create `vecmem::device_vector` objects on the fly when de-referenced.)

As I began, I considered creating a `vecmem::device_vector` array on the heap in `vecmem::jagged_device_vector`'s constructor. But we would run out of resources **really fast** with that approach. Especially when debugging device code.

The custom iterator approach is not impossible, but I'm actually willing to introduce some amount of fragility if it comes with a significant performance gain. I mean, people take advantage of the physical memory layout all the time in high-performance code. (The xAOD EDM does this in a few select places as well.) As one cute example, that I only became aware of recently:

https://en.wikipedia.org/wiki/Fast_inverse_square_root

stephenswat:
The main difference between this and Quake's fast inverse square root is that the latter relies only on the memory layout of floating point numbers and integers, which are both fairly well defined by the IEEE 754 and C standards. The Quake implementation is not flawless, and also relies on undefined behaviour, but it is fairly stable because, firstly, it is compiled for the relatively limited range of hardware that was around back then and, secondly, it relies on those fairly well-defined standards.

What we're trying to do here has none of those advantages. C++'s guarantees about the memory layout of classes is extremely weak. Small changes in access modifiers can change the layout, and the layout is at the behest of the compiler in terms of alignment as well. We know that our compiler situation is already complicated because we are using at least four different ones, which just adds so much bug surface.

If we absolutely must go through with this, I think we should at least do the following things:

1. Make absolutely sure the types have external linkage.
2. Add compile-type checks that the types are layout-compatible using `std::is_layout_compatible`. Note that this does _not_ guarantee that `reinterpret_cast` between the types is safe, that is stated explicitly in the C++ standard, but it's better than nothing.

stephenswat:
> Creating the objects on the fly doesn't work with using iterators.

Thinking a little more about this, why not? This would only apply for the highest-level iterator, over the rows of the jagged vector. Those should be pretty infrequent (compared to the inner-most loop), so I don't think the overhead of this would be that problematic. What do you think?

core/include/vecmem/containers/details/jagged_device_vector_iterator.hpp
```diff
@@ -0,0 +1,184 @@
...
+namespace std {
+
+   /// Specialisation of @c std::iterator_traits
+   ///
+   /// This is necessary to make @c vecmem::reverse_iterator functional on top
+   /// of @c vecmem::details::jagged_device_vector_iterator.
+   ///
+   template< typename T >
+   struct
+   iterator_traits< vecmem::details::jagged_device_vector_iterator< T > > {
```
stephenswat, line 166: I'm not completely sure about this, but I think we should also declare `iterator_traits<vecmem::details::jagged_device_vector_iterator<T>>::iterator_category` to be `random_access_iterator_tag`.

krasznaa: It wouldn't hurt. It's not needed by the code, but having it there should not cause any problems either.

krasznaa:
The failure is fascinating... :confused:

Only in an optimised build the reverse iteration is not working correctly. And what really baffles me is that as soon as I add a printout to the code, the test starts working correctly.

```diff
diff --git a/core/include/vecmem/containers/impl/jagged_device_vector_iterator.ipp b/core/include/vecmem/containers/impl/jagged_device_vector_iterator.ipp
index 9e23f7e..317a516 100644
--- a/core/include/vecmem/containers/impl/jagged_device_vector_iterator.ipp
+++ b/core/include/vecmem/containers/impl/jagged_device_vector_iterator.ipp
@@ -7,6 +7,8 @@
  */
 #pragma once
 
+#include <iostream>
+
 namespace vecmem { namespace details {
 
    template< typename TYPE >
@@ -80,6 +82,7 @@ namespace vecmem { namespace details {
    jagged_device_vector_iterator< TYPE >::operator->() const {
 
       ensure_valid();
+      std::cout << "de-referencing m_ptr = " << m_ptr << ", m_value.size() = " << m_value.size() << std::endl;
       return &m_value;
    }
```

With this in place the behaviour "fixes itself". :confused: I don't want to jump the shark here, but could we have run into a bug in both Clang and GCC?

krasznaa:
I mean, obviously the compilers decide that `ensure_valid()` is a no-op, so they are not executing that function. Is this expected? :confused:

Since `ensure_valid()` is only called from 2 places in the end, I'll just get rid of that helper function. But it's still quite fascinating how optimisation would choose which functions have or do not have an effect on the outcome of the code. :thinking:

stephenswat:
I doubt it, the compilers compiling the failing version report the following warning:

```
/github/workspace/tests/core/test_core_jagged_vector_view.cpp: In member function 'virtual void core_jagged_vector_view_test_reverse_iterator_Test::TestBody()':
/github/workspace/tests/core/test_core_jagged_vector_view.cpp:117:23: warning: '*((void*)&<anonymous> +16)' may be used uninitialized in this function [-Wmaybe-uninitialized]
         i += itr->size();
```

This is almost certainly related. I'll dive into it a little more.

krasznaa:
Curious that it's only GCC that prints that warning though. Clang silently produces a non-functional binary.

Note that adding the following also fixes the code's behaviour:

```diff
diff --git a/core/include/vecmem/utils/reverse_iterator.ipp b/core/include/vecmem/utils/reverse_iterator.ipp
index 87c0124..4e4342e 100644
--- a/core/include/vecmem/utils/reverse_iterator.ipp
+++ b/core/include/vecmem/utils/reverse_iterator.ipp
@@ -9,6 +9,8 @@
 // Local include(s).
 #include "vecmem/utils/types.hpp"
 
+#include <iostream>
+
 namespace vecmem {
 
    template< typename Iterator >
@@ -169,6 +171,7 @@ namespace vecmem {
    typename reverse_iterator< Iterator >::pointer
    reverse_iterator< Iterator >::to_pointer( T itr ) {
 
+      std::cout << "to_pointer(...) called" << std::endl;
       return itr.operator->();
    }
```

Obviously I was trying to check that the expected call chain would be in effect. Which it seems to be, at least as long as I add these printouts.

It starts feeling like Schrodinger's cat. As long as I look closely, it's alive. But once it should do things without any checks, it fails.

stephenswat:
Yeah, to me that indicates either a race condition in multi-threaded code, or undefined behaviour somewhere. Obviously the second in this case. :frowning:

krasznaa:
I'm all open to suggestions. :confused: I tried to understand what the heck is happening by stepping through that test function in a "RelWithDebInfo" build line by line. But it doesn't make any sense...

```
[       OK ] core_jagged_vector_view_test.iterator (0 ms)
[ RUN      ] core_jagged_vector_view_test.reverse_iterator

Breakpoint 1, vecmem::jagged_device_vector<int>::end (this=<optimised out>, this=<optimised out>) at /data/ssd-1tb/projects/vecmem/vecmem/tests/core/test_core_jagged_vector_view.cpp:115
115         std::size_t i = 0;
(gdb) s
core_jagged_vector_view_test_reverse_iterator_Test::TestBody (this=0x555555641ab0) at /data/ssd-1tb/projects/vecmem/vecmem/tests/core/test_core_jagged_vector_view.cpp:116
116         for( auto itr = m_jag.rbegin(); itr != m_jag.rend(); ++itr ) {
(gdb) s
vecmem::jagged_device_vector<int>::rbegin (this=0x555555641b10) at /data/ssd-1tb/projects/vecmem/vecmem/tests/core/test_core_jagged_vector_view.cpp:116
116         for( auto itr = m_jag.rbegin(); itr != m_jag.rend(); ++itr ) {
(gdb) s
vecmem::jagged_device_vector<int>::end (this=0x555555641b10, this=0x555555641b10) at /data/ssd-1tb/projects/vecmem/vecmem/core/include/vecmem/containers/impl/jagged_device_vector.ipp:176
176         jagged_device_vector< T >::rbegin() {
(gdb) s
core_jagged_vector_view_test_reverse_iterator_Test::TestBody (this=0x555555641ab0) at /data/ssd-1tb/projects/vecmem/vecmem/core/include/vecmem/containers/impl/device_vector.ipp:16
16         device_vector< TYPE >::
(gdb) bt
#0  core_jagged_vector_view_test_reverse_iterator_Test::TestBody (this=0x555555641ab0) at /data/ssd-1tb/projects/vecmem/vecmem/core/include/vecmem/containers/impl/device_vector.ipp:16
#1  0x00005555555e7d21 in testing::internal::HandleSehExceptionsInMethodIfSupported<testing::Test, void> (location=0x5555555ef192 "the test body", method=<optimised out>, object=0x555555641ab0)
    at /data/ssd-1tb/projects/vecmem/vecmem-build-vanilla/_deps/googletest-src/googletest/src/gtest.cc:2414
#2  testing::internal::HandleExceptionsInMethodIfSupported<testing::Test, void> (object=object@entry=0x555555641ab0, method=<optimised out>, location=location@entry=0x5555555ef192 "the test body")
    at /data/ssd-1tb/projects/vecmem/vecmem-build-vanilla/_deps/googletest-src/googletest/src/gtest.cc:2469
#3  0x00005555555dac26 in testing::Test::Run (this=this@entry=0x555555641ab0) at /data/ssd-1tb/projects/vecmem/vecmem-build-vanilla/_deps/googletest-src/googletest/src/gtest.cc:2508
#4  0x00005555555daec2 in testing::Test::Run (this=0x555555641ab0) at /data/ssd-1tb/projects/vecmem/vecmem-build-vanilla/_deps/googletest-src/googletest/src/gtest.cc:2684
#5  testing::TestInfo::Run (this=0x55555563cf90) at /data/ssd-1tb/projects/vecmem/vecmem-build-vanilla/_deps/googletest-src/googletest/src/gtest.cc:2684
#6  0x00005555555db06c in testing::TestInfo::Run (this=<optimised out>) at /data/ssd-1tb/projects/vecmem/vecmem-build-vanilla/_deps/googletest-src/googletest/src/gtest.cc:2658
#7  testing::TestSuite::Run (this=0x55555563c4b0) at /data/ssd-1tb/projects/vecmem/vecmem-build-vanilla/_deps/googletest-src/googletest/src/gtest.cc:2816
#8  0x00005555555db843 in testing::TestSuite::Run (this=<optimised out>) at /data/ssd-1tb/projects/vecmem/vecmem-build-vanilla/_deps/googletest-src/googletest/src/gtest.cc:2796
#9  testing::internal::UnitTestImpl::RunAllTests (this=0x555555629170) at /data/ssd-1tb/projects/vecmem/vecmem-build-vanilla/_deps/googletest-src/googletest/src/gtest.cc:5338
#10 0x00005555555e8291 in testing::internal::HandleSehExceptionsInMethodIfSupported<testing::internal::UnitTestImpl, bool> (
    location=0x5555555f0578 "auxiliary test code (environments or event listeners)", method=<optimised out>, object=0x555555629170)
    at /data/ssd-1tb/projects/vecmem/vecmem-build-vanilla/_deps/googletest-src/googletest/src/gtest.cc:2414
#11 testing::internal::HandleExceptionsInMethodIfSupported<testing::internal::UnitTestImpl, bool> (object=0x555555629170, method=<optimised out>, 
    location=location@entry=0x5555555f0578 "auxiliary test code (environments or event listeners)")
    at /data/ssd-1tb/projects/vecmem/vecmem-build-vanilla/_deps/googletest-src/googletest/src/gtest.cc:2469
#12 0x00005555555db9d5 in testing::UnitTest::Run (this=0x5555556166a0 <testing::UnitTest::GetInstance()::instance>)
    at /data/ssd-1tb/projects/vecmem/vecmem-build-vanilla/_deps/googletest-src/googletest/include/gtest/gtest.h:1412
#13 0x000055555557a764 in RUN_ALL_TESTS () at /data/ssd-1tb/projects/vecmem/vecmem-build-vanilla/_deps/googletest-src/googletest/include/gtest/gtest.h:2473
#14 main (argc=<optimised out>, argv=0x7fffffffdde8) at /data/ssd-1tb/projects/vecmem/vecmem-build-vanilla/_deps/googletest-src/googletest/src/gtest_main.cc:45
(gdb) s
119         EXPECT_EQ( i, 16 );
(gdb) s
testing::internal::EqHelper::Compare<unsigned long, int, (void*)0> (rhs=<optimised out>, lhs=<optimised out>, rhs_expression=<optimised out>, lhs_expression=<optimised out>)
    at /data/ssd-1tb/projects/vecmem/vecmem/tests/core/test_core_jagged_vector_view.cpp:119
119         EXPECT_EQ( i, 16 );
(gdb)
```

From the `vecmem::jagged_device_vector::rbegin()` function it suddenly jumps into the constructor of `vecmem::device_vector`, and then right to the end of the loop. :confused: I'm just so very confused...

stephenswat:
I'm investigating... I'll keep you up to date if I find anything.

stephenswat:
Okay, so compiling with `-O1` works fine, and compiling with `-O1 -finline-small-functions` breaks the code.

stephenswat:
On a (probably) unrelated note: the `jagged_device_vector_iterator<T>::jagged_device_vector_iterator(data::vector_view<U>*)` constructor is broken.

krasznaa:
It **is** unrelated, but you're right. While `vecmem::data::vector_view<const T>` can be created from `vecmem::data::vector_view<T>`, pointers of these two types can not be created from each other without some forceful casting. I'll just remove that constructor.

I'm also trying out a few different incantations in the meanwhile. If I add `__attribute__((noinline))` to just the `vecmem::details::jagged_device_vector_iterator::ensure_valid` function:
  - The warning about using a potentially uninitialised variable is still printed;
  - But the results become correct.

If I add that attribute directly to `vecmem::details::jagged_device_vector_iterator::operator->`, then the warning goes away, and I get functional code.

At this point I do need to blame the compiler... As I really don't see why it wouldn't be valid to inline that code. As long as the inlining would would correctly...

krasznaa:
There is definitely something going wrong with the reverse iterator, I just still can't figure out what...

I was now playing with the iterators in CUDA device code. Adding the following behaves exactly as I would expect:

```diff
diff --git a/tests/cuda/test_cuda_jagged_vector_view_kernels.cu b/tests/cuda/test_cuda_jagged_vector_view_kernels.cu
index aa97c68..d701857 100644
--- a/tests/cuda/test_cuda_jagged_vector_view_kernels.cu
+++ b/tests/cuda/test_cuda_jagged_vector_view_kernels.cu
@@ -26,6 +26,12 @@ void doubleJaggedKernel(
     for (std::size_t i = 0; i < jag.at(t).size(); ++i) {
         jag.at(t).at(i) *= 2;
     }
+    __syncthreads();
+    for( auto itr = jag.begin(); itr != jag.end(); ++itr ) {
+        if( ( jag[ t ].size() > 0 ) && ( itr->size() > 0 ) ) {
+            jag[ t ].front() += itr->front();
+        }
+    }
 }
```

But if I try

```diff
diff --git a/tests/cuda/test_cuda_jagged_vector_view_kernels.cu b/tests/cuda/test_cuda_jagged_vector_view_kernels.cu
index aa97c68..8d5723f 100644
--- a/tests/cuda/test_cuda_jagged_vector_view_kernels.cu
+++ b/tests/cuda/test_cuda_jagged_vector_view_kernels.cu
@@ -26,6 +26,12 @@ void doubleJaggedKernel(
     for (std::size_t i = 0; i < jag.at(t).size(); ++i) {
         jag.at(t).at(i) *= 2;
     }
+    __syncthreads();
+    for( auto itr = jag.rbegin(); itr != jag.rend(); ++itr ) {
+        if( ( jag[ t ].size() > 0 ) && ( itr->size() > 0 ) ) {
+            jag[ t ].front() += itr->front();
+        }
+    }
 }
```

, CUDA just freaks out. It compiles the code just fine. And it even manages to run it without any error messages. But if I execute the test inside of [cuda-memcheck](https://docs.nvidia.com/cuda/cuda-memcheck/index.html), it makes it very clear that nothing is going right...

```
[bash][atlas]:build > cuda-memcheck ./test-bin/vecmem_test_cuda 
========= CUDA-MEMCHECK
Running main() from /data/projects/vecmem/build/_deps/googletest-src/googletest/src/gtest_main.cc
[==========] Running 12 tests from 4 test suites.
[----------] Global test environment set-up.
[----------] 2 tests from cuda_containers_test
[ RUN      ] cuda_containers_test.managed_memory
[       OK ] cuda_containers_test.managed_memory (44 ms)
[ RUN      ] cuda_containers_test.explicit_memory
[       OK ] cuda_containers_test.explicit_memory (2 ms)
[----------] 2 tests from cuda_containers_test (46 ms total)

[----------] 1 test from cuda_jagged_vector_view_test
[ RUN      ] cuda_jagged_vector_view_test.mutate_in_kernel
========= Invalid __global__ read of size 8
=========     at 0x00000108 in /data/projects/vecmem/vecmem/core/include/vecmem/containers/impl/device_vector.ipp:273:_ZNK6vecmem13device_vectorIiE4sizeEv
=========     by thread (5,0,0) in block (0,0,0)
=========     Address 0x03fffc20 is out of bounds
=========     Device Frame:/data/projects/vecmem/vecmem/tests/cuda/test_cuda_jagged_vector_view_kernels.cu:31:doubleJaggedKernel(vecmem::data::jagged_vector_view<int>) (doubleJaggedKernel(vecmem::data::jagged_vector_view<int>) : 0xbd0)
=========     Saved host backtrace up to driver entry point at kernel launch time
=========     Host Frame:/usr/lib64/libcuda.so.1 (cuLaunchKernel + 0x2b8) [0x222dd8]
=========     Host Frame:./test-bin/vecmem_test_cuda [0x86739]
=========     Host Frame:./test-bin/vecmem_test_cuda [0x867c7]
=========     Host Frame:./test-bin/vecmem_test_cuda [0xbcb25]
=========     Host Frame:./test-bin/vecmem_test_cuda [0x2f423]
=========     Host Frame:./test-bin/vecmem_test_cuda [0x2f2e9]
=========     Host Frame:./test-bin/vecmem_test_cuda [0x2f31e]
=========     Host Frame:./test-bin/vecmem_test_cuda [0x2f142]
=========     Host Frame:./test-bin/vecmem_test_cuda [0x2bc23]
=========     Host Frame:./test-bin/vecmem_test_cuda [0x7236b]
=========     Host Frame:./test-bin/vecmem_test_cuda [0x5b7da]
=========     Host Frame:./test-bin/vecmem_test_cuda [0x3be39]
=========     Host Frame:./test-bin/vecmem_test_cuda [0x3cb9d]
=========     Host Frame:./test-bin/vecmem_test_cuda [0x3d25b]
=========     Host Frame:./test-bin/vecmem_test_cuda [0x49fa1]
=========     Host Frame:./test-bin/vecmem_test_cuda [0x750fb]
=========     Host Frame:./test-bin/vecmem_test_cuda [0x5de6a]
=========     Host Frame:./test-bin/vecmem_test_cuda [0x49b0f]
=========     Host Frame:./test-bin/vecmem_test_cuda [0x2f491]
=========     Host Frame:./test-bin/vecmem_test_cuda [0x2f473]
=========     Host Frame:/usr/lib64/libc.so.6 (__libc_start_main + 0xf5) [0x22555]
=========     Host Frame:./test-bin/vecmem_test_cuda [0xa1b2]
...
```

Very, very curious...

stephenswat:
Interestingly, when I use `std::reverse_iterator` instead of `vecmem::reverse_iterator`, I get the exact same error in the exact same test. So the error is most likely _not_ in the reverse iterator.

Also, why do we have `vecmem::reverse_iterator` when `std::reverse_iterator` exists?

krasznaa:
Because `std::reverse_iterator` doesn't work in device code. :stuck_out_tongue: In fact I missed a few `VECMEM_HOST_AND_DEVICE` declarations from `vecmem::reverse_iterator` as well, which I had to add for my latest tests.

Yes, I tried this as well. It's not `vecmem::reverse_iterator` per se that makes a mistake. It also works just fine on top of simple pointers.

It's somehow the `vecmem::reverse_iterator< vecmem::details::jagged_device_vector_iterator< T > >` type that has problems with every single compiler. Including `nvcc`...

krasznaa:
Even in Debug mode, where the test succeeds, Valgrind still freaks out about the code.

```
...
[ RUN      ] core_jagged_vector_view_test.reverse_iterator
==347468== Conditional jump or move depends on uninitialised value(s)
==347468==    at 0x169344: testing::AssertionResult testing::internal::CmpHelperEQ<unsigned long, int>(char const*, char const*, unsigned long const&, int const&) (gtest.h:1527)
==347468==    by 0x16796F: testing::AssertionResult testing::internal::EqHelper::Compare<unsigned long, int, (void*)0>(char const*, char const*, unsigned long const&, int const&) (gtest.h:1554)
==347468==    by 0x19D9F2: core_jagged_vector_view_test_reverse_iterator_Test::TestBody() (test_core_jagged_vector_view.cpp:119)
==347468==    by 0x1D643E: void testing::internal::HandleSehExceptionsInMethodIfSupported<testing::Test, void>(testing::Test*, void (testing::Test::*)(), char const*) (gtest.cc:2433)
==347468==    by 0x1CF58E: void testing::internal::HandleExceptionsInMethodIfSupported<testing::Test, void>(testing::Test*, void (testing::Test::*)(), char const*) (gtest.cc:2469)
==347468==    by 0x1AACC9: testing::Test::Run() (gtest.cc:2508)
==347468==    by 0x1AB6B4: testing::TestInfo::Run() (gtest.cc:2684)
==347468==    by 0x1ABDEF: testing::TestSuite::Run() (gtest.cc:2816)
==347468==    by 0x1B7EAB: testing::internal::UnitTestImpl::RunAllTests() (gtest.cc:5338)
==347468==    by 0x1D7879: bool testing::internal::HandleSehExceptionsInMethodIfSupported<testing::internal::UnitTestImpl, bool>(testing::internal::UnitTestImpl*, bool (testing::internal::UnitTestImpl::*)(), char const*) (gtest.cc:2433)
==347468==    by 0x1D0664: bool testing::internal::HandleExceptionsInMethodIfSupported<testing::internal::UnitTestImpl, bool>(testing::internal::UnitTestImpl*, bool (testing::internal::UnitTestImpl::*)(), char const*) (gtest.cc:2469)
==347468==    by 0x1B66EB: testing::UnitTest::Run() (gtest.cc:4925)
==347468== 
[       OK ] core_jagged_vector_view_test.reverse_iterator (3 ms)
...
```

But I still can't decode from this what the heck is going wrong. :frowning:

krasznaa:
I think I (sort of) understand what's going wrong. But it's not looking good for this "type correct" implementation. :frowning:

The problem is that I tried to make (non-const) references work with `vecmem::details::jagged_device_vector_iterator`. For this I thought that I could have a `vecmem::device_vector` member in `vecmem::details::jagged_device_vector_iterator`, which would be lazily updated to represent the object that the iterator is at any point pointing at.

But I don't think anymore that that is a valid design. :frowning: Since if a client grabs a pointer/reference to the `vecmem::device_vector` object that the iterator touts as its value type, the grabbed object can be changed under the "grabber code's" feet.

I modified the iterator to return everything by value, on the stack. Having to introduce yet another adaptor class for it. (So that the iterator could return an object which would behave like a pointer.) With that implementation all of the memory errors seem to be gone.

I'll write this up into a clean-ish commit tomorrow. We can both ponder until then if we're willing to create this many layers of objects on the stack just to avoid "one" fragile type conversion...

krasznaa:
By now I believe I fully understood why the issue was showing up specifically for the reverse iteration. Since that was indeed the only test in which the fragile code ended up breaking.

You see, `vecmem::reverse_iterator` (just like [std::reverse_iterator](https://en.cppreference.com/w/cpp/iterator/reverse_iterator)) creates temporary forward iterator objects every time that it is de-referenced.

https://github.com/acts-project/vecmem/blob/main/core/include/vecmem/utils/reverse_iterator.ipp#L70-L86

These temporary `vecmem::details::jagged_device_vector_iterator` objects then returned a reference/pointer to their internal `vecmem::device_vector` object from these de-referencing functions. But as soon as the functions return, the temporary forward iterator object is already gone. With the assumption being that the returned reference/pointer would still be valid nevertheless. (Since the assumption is that the iterator visits persistent objects in memory, and doesn't just create objects on the fly.)

With all this nice explanation, I now see that the HIP build still died. :confused: It didn't do so with HIP 4.0.0 in my local tests. :confused:

krasznaa:
Ha!!!

```
In file included from /__w/vecmem/vecmem/tests/hip/test_hip_containers_kernels.hip:14:
In file included from /__w/vecmem/vecmem/core/include/vecmem/containers/const_device_array.hpp:10:
In file included from /__w/vecmem/vecmem/core/include/vecmem/containers/device_array.hpp:11:
In file included from /__w/vecmem/vecmem/core/include/vecmem/utils/reverse_iterator.hpp:143:
/__w/vecmem/vecmem/core/include/vecmem/utils/reverse_iterator.ipp:176:9: error: __host__ function 'operator==' cannot overload __host__ __device__ function 'operator=='
   bool operator==( const reverse_iterator< T >& itr1,
        ^
/__w/vecmem/vecmem/core/include/vecmem/utils/reverse_iterator.hpp:132:9: note: previous declaration is here
   bool operator==( const reverse_iterator< T >& itr1,
        ^
In file included from /__w/vecmem/vecmem/tests/hip/test_hip_containers_kernels.hip:14:
In file included from /__w/vecmem/vecmem/core/include/vecmem/containers/const_device_array.hpp:10:
In file included from /__w/vecmem/vecmem/core/include/vecmem/containers/device_array.hpp:11:
In file included from /__w/vecmem/vecmem/core/include/vecmem/utils/reverse_iterator.hpp:143:
/__w/vecmem/vecmem/core/include/vecmem/utils/reverse_iterator.ipp:183:9: error: __host__ function 'operator!=' cannot overload __host__ __device__ function 'operator!='
   bool operator!=( const reverse_iterator< T >& itr1,
        ^
/__w/vecmem/vecmem/core/include/vecmem/utils/reverse_iterator.hpp:137:9: note: previous declaration is here
   bool operator!=( const reverse_iterator< T >& itr1,
        ^
2 errors generated when compiling for gfx803.
```

It seems that for HIP one **does** need to put the `VECMEM_HOST_AND_DEVICE` declaration on the implementation of the functions as well! The reason that the code didn't freak out about `vecmem::jagged_device_vector` not doing it, is that we don't test that type with HIP yet... :stuck_out_tongue:

stephenswat:
The joys of using four different compilers for the same project... Good sleuthing!

krasznaa:
I think I made a mistake. :frowning: Just after pushing this latest change, I realised that the test code is not thread-safe. :frowning: (I.e. the results could vary run-by-run I think.)

But I guess this PR is anyway meant as a draft at this point...

krasznaa:
Replaced by #55.

PR #55, opened 2021-03-26
vecmem::jagged_device_vector Update, main branch (2021.03.26.)

krasznaa:
This is meant to replace #54.

After a long discussion in that PR, (for now) `vecmem::jagged_device_vector` is implemented with no unsafe type conversions. Returning `vecmem::device_vector` objects by value from many of its functions instead.

This was implemented through a number of technical changes:
  - Renamed `vecmem::reverse_iterator` to `vecmem::details::reverse_iterator`, and moved it under `vecmem/containers/details/`. This is so the 2 iterators of the project would end up side-by-side.
  - Introduced the `vecmem::details::is_same_nc` trait, and taught `vecmem::device_array` and `vecmem::device_vector` using it how to create "constant objects" on top of non-const data. Which was absolutely necessary in `vecmem::device_vector` for `vecmem::jagged_device_vector`. (And as long as I was doing it there, I also updated `vecmem::device_array` in the same way.)
  - Introduced a custom `vecmem::details::jagged_device_vector_iterator` type that provides a random access forward iterator for `vecmem::jagged_device_vector`.
 
With all these in place I largely re-wrote `vecmem::jagged_device_vector` to provide an interface similar to that of `vecmem::device_vector`.

Finally, I updated the relevant tests to use/test the new interface of `vecmem::jagged_device_vector`.

### DIFF
--- a/core/CMakeLists.txt
+++ b/core/CMakeLists.txt
@@ -33,6 +33,8 @@ vecmem_add_library( vecmem_core core SHARED
    "include/vecmem/containers/data/vector_view.hpp"
    "include/vecmem/containers/impl/vector_view.ipp"
    # Iterator types.
+   "include/vecmem/containers/details/jagged_device_vector_iterator.hpp"
+   "include/vecmem/containers/impl/jagged_device_vector_iterator.ipp"
    "include/vecmem/containers/details/reverse_iterator.hpp"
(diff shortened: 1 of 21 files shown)


stephenswat (approved):
This looks great, let's go through with it.

PR #56, opened 2021-03-27
vecmem::data::jagged_vector_buffer Introduction, main branch (2021.03.27.)

krasznaa:
In this I tried to figure out how to hold the data of a jagged vector in memory independent of the `vecmem::jagged_vector` that created/filled that memory in the first place.

After quite some deliberation I decided that it would have to be done something like this. That the `vecmem::data::jagged_vector_buffer` object would have to hold 1 or 2 `vecmem::data::vector_view` arrays, depending on what sort of memory resource(s) it gets.

Allocating the memory for holding the "inner vector elements" is easy. `vecmem::data::jagged_vector_buffer` just needs to allocate a 1 dimensional array large enough to hold all of those elements. Since we don't need to be able to read/write that array from host code, that's easy to do.

But the `vecmem::data::vector_view` array we absolutely need host access to. As it needs to be set up to properly point at points inside of the earlier mentioned array. In `vecmem::data::jagged_vector_data` we do this implicitly. Requiring the memory resource passed to `vecmem::data::jagged_vector_data` to **always** provide host-accessible memory. But with `vecmem::data::jagged_vector_buffer`, just like with `vecmem::data::vector_buffer`, I also want to be able to use non-host-accessible memory.

To resolve this I've set up `vecmem::data::jagged_vector_buffer` to allocate either 1 or 2 `vecmem::data::vector_view` arrays, based on how it is constructed. If it is constructed using a single `vecmem::memory_resource`, then it will assume that it is a host-accessible memory resource. If one wants to use a memory resource that's not host-accessible, then both that non-host-accessible, and a host-accessible memory resource needs to be provided to the constructor of `vecmem::data::jagged_vector_buffer`.

In this setup the underlying `vecmem::data::jagged_vector_view` member variables are set up to point to "stuff" that needs to be transferred to the device to use the jagged vector. While `vecmem::data::jagged_vector_buffer::host_ptr()` returns a pointer to the `vecmem::data::vector_view` array that's always (meant to be) host accessible.

The idea here is that the language specific memory copy functions will need to not only arrange for the copy of the array holding the "internal vector elements", but also the copy of the host-accessible `vecmem::data::vector_view` array into the non-host-accessible one.

A couple of things to take note of:
  - `vecmem::data::jagged_vector_buffer` does not inherit from `vecmem::data::jagged_vector_data`, but rather from `vecmem::data::jagged_vector_view`. I'm not too thrilled about this, but I couldn't find a setup where we wouldn't need 3 different classes. I'm ready to be proven wrong though.
  - `vemcem::data::jagged_vector_buffer` does not provide direct access to its "internal vector array". Clients of this object are expected to access that array through the `vecmem::data::vector_view` objects accessible through `vecmem::data::jagged_vector_buffer::host_ptr()`. Deciding at runtime how many separate copy operations they will need to perform the memory transfer. (This is where `vecmem::contiguous_memory_resource` should come into play as well. But I didn't try that in practice just yet.)

### DIFF
--- a/core/CMakeLists.txt
+++ b/core/CMakeLists.txt
@@ -24,6 +24,8 @@ vecmem_add_library( vecmem_core core SHARED
    "include/vecmem/containers/vector.hpp"
    "include/vecmem/containers/impl/vector.ipp"
    # Data holding/transporting types.
+   "include/vecmem/containers/data/jagged_vector_buffer.hpp"
+   "include/vecmem/containers/impl/jagged_vector_buffer.ipp"
    "include/vecmem/containers/data/jagged_vector_data.hpp"
    "include/vecmem/containers/impl/jagged_vector_data.ipp"
    "include/vecmem/containers/data/jagged_vector_view.hpp"
--- a/core/include/vecmem/containers/data/jagged_vector_buffer.hpp
+++ b/core/include/vecmem/containers/data/jagged_vector_buffer.hpp
@@ -60,7 +60,9 @@ namespace vecmem::data {
       ///        be host accessible.
       /// @param host_access_resource An optional host accessible memory
       ///        resource. Needed if @c resource is not host accessible.
-      template< typename OTHERTYPE >
+      template< typename OTHERTYPE,
+                std::enable_if_t< std::is_convertible< TYPE, OTHERTYPE >::value,
+                                  bool > = true >
       jagged_vector_buffer( const jagged_vector_view< OTHERTYPE >& other,
                             memory_resource& resource,
                             memory_resource* host_access_resource = nullptr );
--- a/core/include/vecmem/containers/data/jagged_vector_buffer.hpp
+++ b/core/include/vecmem/containers/data/jagged_vector_buffer.hpp
@@ -1,0 +1,105 @@
+/*
+ * VecMem project, part of the ACTS project (R&D line)
+ *
+ * (c) 2021 CERN for the benefit of the ACTS project
+ *
+ * Mozilla Public License Version 2.0
+ */
+#pragma once
+
+// Local include(s).
+#include "vecmem/containers/data/jagged_vector_view.hpp"
+#include "vecmem/memory/deallocator.hpp"
+#include "vecmem/memory/memory_resource.hpp"
+
+// System include(s).
+#include <cstddef>
+#include <memory>
+#include <type_traits>
+#include <vector>
+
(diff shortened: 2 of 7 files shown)


stephenswat:
I'm not sure I'm quite following here. You say that the `jagged_vector_buffer` has ownership of a single two-dimensional offset array that contains all the items in the jagged array; that implies that the construction of an object of this class requires us to copy all the elements over from the `std::vector` to our custom class, right? Isn't that pretty much exactly what we _didn't_ want to do? It's not really clear to me when you would ever want to use this type.

Other than that this class hierarchy is getting pretty confusing pretty quickly. How exactly does this class relate to `jagged_vector_view`? When do I use which one?

krasznaa:
:confused: The current implementation **only** works with shared/managed memory. It doesn't work with anything else. And shared/managed memory is always less performant than scheduling explicit memory copies by hand. This is why I introduced `vecmem::data::vector_buffer` towards the beginning as well.

I do want to be able to manage jagged vectors separately in host and device memory, without relying on the GPU runtime's implementation of automatic memory copies. So no, I definitely didn't want to avoid doing what this PR is doing. :confused:

For the class hierarchy: Any code that doesn't need to know about who owns a piece of memory, will continue using `vecmem::data::jagged_vector_view`, as it did so far. So all device code, and code "just around" device code will use this type. `vecmem::get_data` will continue using `vecmem::data::jagged_vector_data`, as that function is not meant to do any copies. It has to create an objects that's "host accessible".

Just like in the existing example for `vecmem::data::vector_buffer` (which I will clearly need to expand on, to make it clear what that type is meant for), `vecmem::data::jagged_vector_buffer` is something that users will be expected to create explicitly. When they want to manage the copies of their data explicitly. Which will be most of the time in any place where memory handling is performance critical.

Remember, the whole "efficient jagged vector copy" concept that we've been talking about before will only be possible to do by hand, when we write code that explicitly handles these data objects. (Like https://github.com/acts-project/vecmem/blob/main/cuda/include/vecmem/utils/cuda/copy.hpp does.) The automated memory copying of shared/managed memory will benefit very little from us efficiently laying out jagged vectors in host/device memory.

krasznaa:
We should really move forward with the discussion on this today...

stephenswat:
Let's move ahead with it then. I still have quite a few concerns, but we can always iron those out later... :slightly_smiling_face:

PR #59, opened 2021-03-31
VECMEM_DEBUG_MSG Update, main branch (2021.03.31.)

krasznaa:
Made `VECMEM_DEBUG_MSG` work without printing any variable values. Previously the macro would always need to receive at least one argument after the string literal that describes the format of the printout.

I tried to use the `##__VA_ARGS__` GNU extension at first, but that does not work with all compilers under all circumstances. :frowning: Not without making the code super compiler specific. This trick on the other hand should work everywhere by passing a useless empty string as an argument to `__VECMEM_DEBUG_MSG` even when the user didn't provide any arguments to `VECMEM_DEBUG_MSG`.

### DIFF
--- a/core/include/vecmem/utils/debug.hpp
+++ b/core/include/vecmem/utils/debug.hpp
@@ -81,6 +81,12 @@
 #   define __VECMEM_PRINT_5(MSG, ...)
 #endif
 
+/// Macro used for handling the case of printing a pure/simple character string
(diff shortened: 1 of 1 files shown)


krasznaa:
I'll just merge it as-is, since it's really just a minor technical update. But one that bothers my other developments quite a bit...

PR #61, opened 2021-04-01
HIP 4.1 Compatibility, main branch (2021.04.01.)

krasznaa:
Made the project's CMake code compatible with [ROCm/HIP](https://rocmdocs.amd.com) 4.1.

In ROCm/HIP 4.1 the previous `"hcc"` and `"nvcc"` values of the `HIP_PLATFORM` environment variable (for the `hipcc` compiler) were replaced with `"amd"` and `"nvidia"` respectively. With no backwards compatibility for `"hcc"` or `"nvcc"`. :frowning: (I just keep being surprised by AMD's software team...)

Seemingly this was the only UI change made. :confused: The macros in the HIP headers still use the `HCC` and `NVCC` naming internally.

For now I kept the default platform on `"hcc"`. So HIP 4.1 users will have to actively set their `HIP_PLATFORM` environment variable to something else.

Note that I switched to HIP 4.1 myself hoping that `printf` would start working better with it. But it didn't. :frowning: The only thing I got out of the update was that I needed to update this CMake code. :frowning:

### DIFF
--- a/cmake/FindHIPToolkit.cmake
+++ b/cmake/FindHIPToolkit.cmake
@@ -12,7 +12,7 @@ endif()
 set( CMAKE_HIP_PLATFORM "${CMAKE_HIP_PLATFORM_DEFAULT}" CACHE STRING
    "Platform to build the HIP code for" )
 set_property( CACHE CMAKE_HIP_PLATFORM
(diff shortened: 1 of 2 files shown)


krasznaa:
I don't expect any opposition to this, so just pinging you @stephenswat for info...

PR #63, opened 2021-04-01
CMake Updates, main branch (2021.04.01.)

krasznaa:
I finally figured out how to correctly link SYCL and HIP binaries with CMake. :partying_face: Had to look at the core CMake files for a bit, but I think I now have it...

The main outcome is the following:
  - SYCL binaries now finally receive correct RPATH settings;
    * In this case this means that binaries in the build directory get "full" RPATH-s set on them, while the installed libraries get nothing;
  - With this I was able to increase the priority of the "HIP linking", similar to the "SYCL linking", as well;
    * This shall make it possible to build HIP test applications that don't use `vecmem::hip`, only `vecmem::core`. So far those would've failed to link properly...
  - And finally this allowed me to remove the `LD_LIBRARY_PATH` tweaking from `vecmem_add_test(...)`. Which should make our code a lot more heterogeneous. :wink: (I'll do some tests on macOS and Windows next. :smile:)


krasznaa:
Well, I just tried compiling the code on Catalina. It didn't work. :stuck_out_tongue: So there will still be some work with making it be more robust...

krasznaa:
In the meanwhile I understood the issue on macOS. But I'll fix that in a separate PR...

PR #67, opened 2021-04-07
Copy Code Rewrite, main branch (2021.04.07.)

krasznaa:
While trying to add explicit memory handling functions for HIP, I think I managed to come up with a design for the memory copy code that should allow us some fair amount of flexibility going forward.

In all of the code that I wrote for CUDA earlier in #37 and #60, the only CUDA specific call was really a single [cudaMemcpy](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__MEMORY.html). So writing the same amount of code for HIP (and then later for SYCL) would've been silly.

After some pondering, I decided to introduce a `vecmem::copy` class that "language specific" classes could inherit from. Even though none of `vecmem::copy`, `vecmem::cuda::copy` or `vecmem::hip::copy` have a state at the moment, I thought that with this design we could implement asynchronous copies in the same way as well. By introducing for instance a `vecmem::cuda::async_copy` type later on, which would perform all of its copies using a `cudaStream_t` that it would get in its constructor. (So that type **would** have a state at that point.)

Please have a look. I myself like this setup quite a bit. (Though I'm not thrilled about some of the function names that I chose...) For now I'm making it a draft, as I'll want to implement the same for SYCL tomorrow as well.


krasznaa:
I'll leave out the SYCL updates from this PR after all. As it turns out, we'll need to tweak the buffer handling a bit with SYCL. As objects have a curious lifetime inside of SYCL kernels. But I'll elaborate on that more in a separate PR.

krasznaa:
Since nobody spoke up, I'll just add it like this. Since I'd like to upload my SYCL memory copy developments "in a nice way" as well, before turning to some other developments that will be necessary for traccc (as I had to realise today)...

PR #68, opened 2021-04-10
CMake Patches, main branch (2021.04.10.)

krasznaa:
Modified `vecmem_add_flag(...)` not to modify flags "globally", but only in a given context. So that when we build VecMem as part of another project (let's say [acts-project/traccc](https://github.com/acts-project/traccc) :wink:), it wouldn't pollute the other project's build options with its own. This required explicit include of `vecmem-compiler-options-cpp.cmake` in a number of additional places.

At the same time I made some additional changes as well:
  - Updated the project's "build version" to `0.2.0`, to be in sync with the previously created [v0.1.0 git tag](https://github.com/acts-project/vecmem/releases/tag/v0.1.0);
  - With the "bleeding" now stopped, I could remove the very specific C\+\+ build flag (`-Wno-gnu-zero-variadic-macro-arguments`) that I had to add earlier to be able to compile GTest with the `-pedantic` flag;
  - Explicitly turned off the build (and the configuration) of the GoogleTest components that VecMem doesn't need. Including the installation of the GoogleTest libraries.

Pinging @stephenswat and @paulgessinger. Though I doubt you guys will have strong opinions about the build system's configuration in such details...

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,7 +6,7 @@
 
 # Set up the project.
 cmake_minimum_required( VERSION 3.10 )
-project( vecmem VERSION 0.0.1 LANGUAGES CXX )
+project( vecmem VERSION 0.2.0 LANGUAGES CXX )
 
 # Standard CMake include(s).
 include( GNUInstallDirs )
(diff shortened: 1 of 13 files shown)


krasznaa:
I'll just go ahead with this one...

PR #69, opened 2021-04-12
SYCL Copy Code, main branch (2021.04.12.)

krasznaa:
As I said in #67, I wrote a `vecmem::sycl::copy` class as well. But to make it convenient to use in our tests, I added some more code as well.

Basically, I added `vecmem::get_data(...)` functions for some more types. To be able to express in a convenient way that I want to pass "view types" to the SYCL device code, and not host-only types. This could of course be done without the `vecmem::get_data(...)` functions as well, but I thought this would be a bit more elegant than having to create temporary reference variables in the code explicitly.

The `vecmem::sycl::copy` class is pretty much the same as `vecmem::cuda::copy` and `vecmem::hip::copy`. Though unlike those, it actually does have a state. As it needs to hold on to a `sycl::queue` object to be able to perform the memory copies.

And the thing that I learned during these developments, which may be obvious to everybody else, but it wasn't to me... :stuck_out_tongue:  At first I was trying to implement the test for the SYCL copy functionality like:

```c++
   queue.submit( [ &constants, &inputvec,
                   &outputvecdevice ]( cl::sycl::handler& h ) {

      // Create buffer objects for the input data.
      auto const_data = copy.to( vecmem::get_data( constants ),
                                 device_resource );
      auto input_data = copy.to( vecmem::get_data( inputvec ),
                                 device_resource );

      // Run the kernel.
      h.parallel_for< class LinearTransform2 >(
         cl::sycl::range< 1 >( input_data.m_size ),
         [ constants = vecmem::get_data( const_data ),
           input = vecmem::get_data( input_data ),
           output = vecmem::get_data( outputvecdevice ) ]( cl::sycl::id< 1 > id ) {
```

I.e. I tried to create the buffer objects inside the lambda that I would give to `sycl::queue::submit`. Not realising that in such a setup the buffers would be deleted already before the kernel code would start to run. Which lead to a lot of errors in the code of course. Which is why I ended up adding quite some debug output in this PR. :stuck_out_tongue:

Long story short, one must create "buffer objects" outside of `sycl::queue::submit`. :wink:

### DIFF
--- a/core/include/vecmem/containers/impl/jagged_vector_buffer.ipp
+++ b/core/include/vecmem/containers/impl/jagged_vector_buffer.ipp
@@ -63,7 +63,7 @@ namespace {
 
 } // private namespace
 
-namespace vecmem::data {
+namespace vecmem { namespace data {
 
    template< typename TYPE >
    template< typename OTHERTYPE,
@@ -119,4 +119,20 @@ namespace vecmem::data {
       return m_outer_host_memory.get();
    }
 
-} // namespace vecmem::data
+} // namespace data
+
+   template< typename TYPE >
+   data::jagged_vector_view< TYPE >&
+   get_data( data::jagged_vector_buffer< TYPE >& data ) {
+
+      return data;
+   }
+
+   template< typename TYPE >
+   const data::jagged_vector_view< TYPE >&
+   get_data( const data::jagged_vector_buffer< TYPE >& data ) {
+
+      return data;
+   }
+
+} // namespace vecmem
--- a/sycl/include/vecmem/utils/sycl/copy.hpp
+++ b/sycl/include/vecmem/utils/sycl/copy.hpp
@@ -27,7 +27,7 @@ namespace vecmem::sycl {
       copy( const queue_wrapper& queue = { "" } );
 
    protected:
-      /// Perform a memory copy using CUDA
+      /// Perform a memory copy using SYCL
       virtual void do_copy( std::size_t size, const void* from, void* to,
                             type::copy_type cptype ) override;
 
--- a/sycl/include/vecmem/utils/sycl/copy.hpp
+++ b/sycl/include/vecmem/utils/sycl/copy.hpp
@@ -1,0 +1,40 @@
+/*
+ * VecMem project, part of the ACTS project (R&D line)
+ *
+ * (c) 2021 CERN for the benefit of the ACTS project
+ *
+ * Mozilla Public License Version 2.0
+ */
+#pragma once
+
+// VecMem include(s).
+#include "vecmem/utils/copy.hpp"
+#include "vecmem/utils/sycl/queue_wrapper.hpp"
+
+namespace vecmem::sycl {
+
+   /// Specialisation of @c vecmem::copy for SYCL
+   ///
+   /// Unlike @c vecmem::cuda::copy and @c vecmem::hip::copy, this object does
+   /// have a state. As USM memory operations in SYCL happen through a
+   /// @c cl::sycl::queue object. So this object needs to point to a valid
+   /// queue object itself.
+   ///
+   class copy : public vecmem::copy {
+
+   public:
+      /// Constructor on top of a user-provided queue
+      copy( const queue_wrapper& queue = { "" } );
+
+   protected:
+      /// Perform a memory copy using CUDA
+      virtual void do_copy( std::size_t size, const void* from, void* to,
+                            type::copy_type cptype ) override;
+
+   private:
+      /// The queue that the copy operations are made with/for
+      queue_wrapper m_queue;
+
+   }; // class copy
+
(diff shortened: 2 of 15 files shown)


sycl/include/vecmem/utils/sycl/copy.hpp
```diff
@@ -0,0 +1,40 @@
...
+   /// queue object itself.
+   ///
+   class copy : public vecmem::copy {
+
+   public:
+      /// Constructor on top of a user-provided queue
+      copy( const queue_wrapper& queue = { "" } );
+
+   protected:
+      /// Perform a memory copy using CUDA
```
paulgessinger, line 30:
```suggestion
      /// Perform a memory copy using SYCL
```
?

core/include/vecmem/containers/impl/jagged_vector_buffer.ipp
```diff
@@ -119,4 +119,20 @@ namespace vecmem::data {
...
 
-} // namespace vecmem::data
+} // namespace data
+
+   template< typename TYPE >
+   data::jagged_vector_view< TYPE >&
+   get_data( data::jagged_vector_buffer< TYPE >& data ) {
+
+      return data;
+   }
```
paulgessinger, line 129: Ok, maybe I'm missing the point, but these functions exist so that you can explicitly write out the buffer -> view conversion (and use a little template deduction)?

paulgessinger (commented):
Looks reasonable to me. 

So the *main* thing here is that the copy object takes the queue, such that the actual copy function interface is the same as for the other backends?

tests/sycl/test_sycl_containers.sycl
```diff
@@ -79,3 +82,80 @@ TEST_F( sycl_containers_test, shared_memory ) {
...
+   // Perform a linear transformation using the vecmem vector helper types.
+   queue.submit( [ &const_data, &input_data,
+                   &outputvecdevice ]( cl::sycl::handler& h ) {
+
+      // Run the kernel.
+      h.parallel_for< class LinearTransform2 >(
+         cl::sycl::range< 1 >( input_data.m_size ),
+         [ constants = vecmem::get_data( const_data ),
+           input = vecmem::get_data( input_data ),
+           output = vecmem::get_data( outputvecdevice ) ]( cl::sycl::id< 1 > id ) {
```
krasznaa, line 132: The `vecmem::get_data(...)` functions were meant for this sort of usage. Without them we would have to do the following:

```c++
const vecmem::data::vector_view< int >& const_view = const_data;
const vecmem::data::vector_view< int >& input_view = input_data;
vecmem::data::vector_view< int >& output_view = outputvecdevice;

h.parallel_for< class LinearTransform >( cl::sycl::range< 1 >( input_data.m_size ),
                                                               [ const_view, input_view, output_view ]( cl::sycl::id< 1 > id ) {
```

I just wanted to avoid forcing users to have to write out the exact types in their code, if at all possible...

tests/sycl/test_sycl_containers.sycl
```diff
@@ -79,3 +82,80 @@ TEST_F( sycl_containers_test, shared_memory ) {
...
+            const vecmem::const_device_vector< int > inputvec( input );
+            vecmem::device_vector< int > outputvec( output );
+
+            // Perform the linear transformation.
+            outputvec.at( id ) = inputvec.at( id ) * constantarray.at( 0 ) +
+                                 constantarray.at( 1 );
+            return;
+         } );
+   } );
+   queue.wait(); // This is needed just for the OpenCL CPU driver...
```
krasznaa, line 151: It turns out I was incorrect here. :confused: I don't know why this test was/is working correctly for me from inside of Docker with a GPU backend. I suspect that the newer version of oneAPI that I'm using, already supports "asynchronous" copies. But when I use the last oneAPI production release, this `wait()` call must be here for all backends...

sycl/include/vecmem/utils/sycl/copy.hpp
krasznaa: :stuck_out_tongue:

krasznaa (commented):
Thanks for the comments!

tests/sycl/test_sycl_containers.sycl
paulgessinger: Yeah, exactly, that's what I was thinking of.

krasznaa:
> So the _main_ thing here is that the copy object takes the queue, such that the actual copy function interface is the same as for the other backends?

This was my idea, yes. As mentioned in #67, I also plan to introduce `vecmem::cuda::async_copy` and similar types as well later on. That one for instance will hold on to a `cudaStream_t` object (pointer) for setting up the copies.

Since these "copy objects" are super lightweight, I thought this would be a good design for abstracting some of the algorithmic code from the way that the "actual copies" are made.

PR #70, opened 2021-04-13
Extendable Device Vectors - Part 1, main branch (2021.04.13.)

krasznaa:
Made `vecmem::data::vector_view` hold a separate size and capacity variable. This is meant as a first step towards "extendable" device vectors.

The intended logic is the following:
  - For fixed size device vectors the code's logic pretty much stays as it is now. `vecmem::data::vector_view` holds on to an `std::size_t` variable that describes both the capacity and the size of the array, and a `T*` pointer that points at the array.
  - For "resizable" device vectors `vecmem::data::vector_view` continues to hold the maximum capacity of the vector as a regular `std::size_t` member variable, since the vector capacities inside of the device code will be fixed still. However the actual size of the vector is held in global memory by `vecmem::data::vector_view` in such a case. Opening the possibility for `vecmem::device_vector` to modify the vector size from device code.

This is just the very first step in adding this non-trivial new functionality. We'll need to:
  - Update the code to correctly create "resizable versions" of the data objects;
  - Teach `vecmem::device_vector` how to extend existing vectors in a safe way;
  - Teach the memory copy code some new tricks. (This feature will of course only work with separate host and device memory allocations. We can't have device code try to resize a vector whose memory is owned by a host vector...)

At the same time found a bug in the `vecmem::data::jagged_vector_buffer` code, and cleaned up the `vecmem::copy` implementation a little. Other than those, most of the updates are just to switch from accessing the `vecmem::data::vector_view` variables directly, to using the new accessor functions of the class.

P.S. I only really foresee adding `push_back(...)` and `emplace_back(...)` functions to `vecmem::device_vector`. Removing elements from the vector, or inserting new elements in the middle of it, is off the table. (That could just not be done efficiently in a thread-safe way. But for `push_back(...)` and `emplace_back(...)` we "just" need a few atomic operations...)

### DIFF
--- a/core/include/vecmem/containers/data/vector_buffer.hpp
+++ b/core/include/vecmem/containers/data/vector_buffer.hpp
@@ -30,6 +30,12 @@ namespace vecmem::data {
    public:
       /// The base type used by this class
       typedef vector_view< TYPE > base_type;
+      /// Size type definition coming from the base class
+      typedef typename base_type::size_type size_type;
+      /// Size pointer type definition coming from the base class
+      typedef typename base_type::size_pointer size_pointer;
+      /// Pointer type definition coming from the base class
+      typedef typename base_type::pointer pointer;
 
       /// @name Checks on the type of the array element
       /// @{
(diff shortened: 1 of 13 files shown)


stephenswat:
Looks good for first steps, but what I am curious about is how we will propagate changes in vector size back to the CPU? In the end, the CPU code still owns the memory in most cases, so how will the `std::vector` be informed of the changes made by kernel code?

krasznaa:
This will only be usable on device vectors instantiated on top of "`vecmem::data::vector_buffer` owned memory". If you check, the `vecmem::get_data(...)` functions are set up to produce `vecmem::data::vector_view` objects that are **not** resizable. Only the `vecmem::data::vector_buffer` objects produce "a view" that allows to be resized.

So I plan to extend the `vecmem::copy` interface a bit to handle the copy back from these "separate buffers" into `vecmem::vector` objects correctly.

I was considering for a while how this could be done on "`vecmem::vector` owned" memory. but decided that that would be super fragile...

PR #71, opened 2021-04-14
Extendable Device Vectors - Part 2, Atomics, main branch (2021.04.14.)

krasznaa:
This is the next step after #71. In order to be able to extend device vectors in device code, we need to be able to execute atomic operations in a "nicely organised" way inside of `vecmem::device_vector`.

To do this, I introduced `vecmem::atomic`. It's a class strongly influenced by [sycl::atomic](https://intel.github.io/llvm-docs/doxygen/classcl_1_1sycl_1_1atomic.html) and (of course) [std::atomic](https://en.cppreference.com/w/cpp/atomic/atomic).

At first I just wanted to make `vecmem::atomic` a typedef of `sycl::atomic` when building SYCL code. But unfortunately the UI of `sycl::atomic` is **very** jenky at the moment. :frowning: (I complained to the Intel developers about that already.) So finally I decided to hide the SYCL atomic calls behind the `vecmem::atomic` interface as well.

I wasted a good amount of time yesterday trying to make this interface able to work in host code as well, on bare pointers. Trying to make use of the C atomic programming interface. While I could (sort of) make it work with Clang, it does not work with GCC. So after a good number of wasted hours I just gave up, and declared `vecmem::atomic` a "device only" type. This is pretty bad, as we can't use it in tests run on the host. :frowning: So I may just add non-atomic operations in the end when using the class in host code. But that will then introduce the possibility of performing non-atomic operations in device code as well. (If/when we make a mistake.) So for now I just went with this setup.

I also ran into a lot of bugs in HIP. :frowning:
  - Trying to execute `atomicSub(...)` on my AMD card just makes my system hang. :scream: I had to hard-reset my desktop yesterday many times because of this. (I don't know why this is, as I didn't find any similar reports with Google. :confused:)
  - Atomic operations and shared host/device memory do not mix on AMD cards. :dizzy_face: That also took me a long time to figure out. (The HIP runtime does not fetch the latest version of the device memory correctly after the kernel finished running.) So one must use explicit device -> host copies to get the correct memory values back to the device.

To make those device -> host copies more convenient, I also extended `vecmem::copy` in this PR in the way that I hinted at in https://github.com/acts-project/vecmem/pull/70#issuecomment-819359315.


krasznaa:
Something is unfortunately royally messed up. :frowning: The changes from #70 are not mixing well with these updates. :confused:

What the problem comes from is that `vecmem::data::vector_view<T>::size()` is returning invalid values inside the SYCL kernel. But I'm still trying to understand why...

krasznaa:
I'm not sure what to do now. :frowning: See https://github.com/intel/llvm/issues/3549 for a description of why the CI test fails. (The CI test actually uses the CPU OpenCL driver and **not** the SYCL host device interestingly enough.)

krasznaa:
I did have to make all of the functions `VECMEM_HOST_AND_DEVICE` after all. I wrote a first implementation for a set of "modifier functions" in `vecmem::device_vector`, and those can really only work if the `vecmem::atomic` functions are declared `VECMEM_HOST_AND_DEVICE`. Since all of the functions that I need to use the `vecmem::atomic<...>` objects from are also declared `VECMEM_HOST_AND_DEVICE`...

I had to play a bit with choosing the right pre-processor definitions to use in the implementation. I think I managed to find them by now, but you guys should also have a look.

krasznaa:
With nobody reacting, I'll just merge it as-is. Since by now I'm trying to debug issues with the device code making use of the `vecmem::atomic` type, and that will be a lot easier to do with this already in the `main` branch...

paulgessinger:
Do I understand correctly that, as the name implies, this is **only** atomic on device, and non-atomic on the host?

PR #72, opened 2021-04-20
CMake GPU Language Fixes, main branch (2021.04.20.)

krasznaa:
While debugging some new developments in the project, I had to realise that compilation flags were not being propagated to all build steps correctly. This is meant to fix that.

Some of the problem comes from #68, as that changed the behaviour of `vecmem_add_flag(...)` to only modify the flag given to it in the current scope. But the include guards on the `vecmem-compiler-options-<lang>.cmake` files then lead to the flags not being set everywhere that they were meant to be. This could simply be solved by removing those include guards.

But as it turns out the optimisation / debug symbol generation flags were just never set up correctly for HIP or SYCL. (For CUDA they were not okay "just" because of the include guards.) So I tried to set up the default compilation flags for the 3 main ("Debug", "Release" and "RelWithDebInfo") build modes correctly for those two languages as well. Even going a step further, and making the code respect the `SYCLFLAGS` and `HIPFLAGS` environment variables set before/during the CMake configuration. Just how CMake respects the `CXXFLAGS` and `CUDAFLAGS` environment variables out of the box. (These are meant to help build systems tweak the builds of a project "in a GNU way".)


krasznaa:
I'll just fast-track this one @stephenswat  and @paulgessinger, as it comes in very handy for any other development...

PR #73, opened 2021-04-20
vecmem::atomic Fixes, main branch (2021.04.20.)

krasznaa:
Fixed the host implementation of `vecmem::atomic`, as I had to find while starting to use it in `vemcem::device_vector`, that it was not correct. :frowning:

At the same time simplified the incantations for the SYCL atomic calls a little bit. (Now that I understand a bit better how the oneAPI atomic implementation is designed.)

Note that  the current scope of the SYCL atomic calls is wider than those used in the CUDA and HIP code. The CUDA and HIP atomic calls are "device specific". While the SYCL calls, in their current form, are "global". Since this is the default for SYCL atomics, I didn't want to change it for now. But we'll probably want to think about using `sycl::device_ptr` instead of the current `sycl::global_ptr` later on, once we start any performance measurements.

### DIFF
--- a/core/include/vecmem/memory/impl/atomic.ipp
+++ b/core/include/vecmem/memory/impl/atomic.ipp
@@ -12,27 +12,17 @@
 #   include <CL/sycl/atomic.hpp>
 #endif
 
-/// Helpers for super explicit calls to the SYCL atomic functions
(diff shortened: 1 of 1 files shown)


krasznaa:
Let me just put in the "slightly more trivial" of these. You can have a bit longer to look at #76...

PR #75, opened 2021-04-20
Extendable Device Vectors - Part 3, vecmem::copy Updates, main branch (2021.04.20.)

krasznaa:
I taught the copy classes how to deal with resizable vector buffers. Since the size of such buffers is not necessarily host accessible, one needs to be careful with both their initial setup, and copying data to/from them.

This meant adding a new `setup(...)` function to `vecmem::copy`, and introducing the `do_memset(...)` virtual function for all of the copy classes, as a small optimisation over only using `do_copy(...)`.

Note that my idea here is that the "inner vectors" of jagged vectors could become resizable eventually. Not the outer vectors, as doing that in a thread-safe way would be pretty impossible. But the inner vectors should be possible to handle like this. (In the Acts seed finding code such output arrays are used in a few places...)

### DIFF
--- a/core/include/vecmem/utils/copy.hpp
+++ b/core/include/vecmem/utils/copy.hpp
@@ -60,6 +60,10 @@ namespace vecmem {
       /// @name 1-dimensional vector data handling functions
       /// @{
 
+      /// Set up the internal state of a vector buffer correctly on a device
+      template< typename TYPE >
+      void setup( data::vector_view< TYPE >& data );
+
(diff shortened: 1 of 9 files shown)


krasznaa:
Forgot to say: I also updated `vecmem::sycl::copy` to **always** perform synchronous memory operations from now on. Since by default SYCL memory copies are asynchronous. (If/when we'll need asynchronous memory copies, we'll introduce a `vecmem::sycl::async_copy` type, along with similar types for CUDA and HIP.)

PR #76, opened 2021-04-20
Extendable Device Vectors - Part 4, Updates to vecmem::device_vector, main branch (2021.04.20.)

krasznaa:
With all the previous changes made, now I made it possible to resize `vecmem::device_vector`-s (in some situations). It can only be done on top of "resizable buffers". And even in that case it's only the `vecmem::device_vector::push_back(...)` and `vecmem::device_vector::emplace_back(...)` functions can may be used without some form of global synchronisation.

Changed the size type variables in 1D vectors to `unsigned int` (from `std::size_t`), as that works much better with atomic operations on GPUs. Also updated all of the relevant unit tests to exercise the new functionality of `vecmem::device_vector`.

I did not implement any functions in `vecmem::device_vector` that would've required a [memmove](https://en.cppreference.com/w/cpp/string/byte/memmove) operation. Since none of the GPU languages provide such a function for the device code. That functionality **could** be added to our own code using multiple `memcpy` operations of course. But unless somebody really needs such operations (like inserting or removing elements in/from the middle of a vector), we should just not bother.

Note that for these changes to work, #73, #74 and #75 all need to be merged in first. Still, I wanted to open this PR already now, so that the changes in the previous PRs could be looked at "in context". :wink:


krasznaa:
Let me increase the pressure a bit. :stuck_out_tongue: I now have a functional test locally in which I fill elements into a jagged vector on my AMD GPU. :partying_face: (In a kernel that is executed on a 2D grid. :smile:) So you should really have a look at this PR, as tomorrow at one point I'll want to merge it in...

core/include/vecmem/containers/device_vector.hpp
```diff
@@ -118,6 +124,60 @@ namespace vecmem {
...
+         for( InputIt itr = other_begin; itr != other_end; ++itr ) {
+            construct( asize.fetch_add( 1 ), *itr );
+         }
+      }
+
+      /// Add a new element at the end of the vector ("kind of" thread-safe)
+      template< typename... Args >
+      VECMEM_HOST_AND_DEVICE
+      reference emplace_back( Args&&... args );
+      /// Add a new element at the end of the vector ("kind of" thread-safe)
```
paulgessinger, line 160: Is it obvious what "kind of" means? Thread-safe unless capacity needs increased?

core/include/vecmem/containers/impl/device_vector.ipp
```diff
@@ -74,7 +81,7 @@ namespace vecmem {
    device_vector< TYPE >::at( size_type pos ) {
```
paulgessinger, line 81: Nitpick: isn't the usual assumption that `at()` methods always perform bounds checks, not only in DEBUG?

core/include/vecmem/containers/device_vector.hpp
```diff
@@ -35,10 +36,15 @@ namespace vecmem {
...
-      typedef std::size_t       size_type;
+      typedef unsigned int      size_type;
       /// Pointer difference type
       typedef std::ptrdiff_t    difference_type;
 
+      /// Pointer type to the size of the array
+      typedef typename std::conditional< std::is_const< TYPE >::value,
+                                         const size_type*,
+                                         size_type* >::type size_pointer;
+
```
paulgessinger, line 47: I don't really understand what the constness of `size_pointer` is relevant for. Could you elaborate?

core/include/vecmem/containers/impl/vector.ipp
```diff
@@ -13,15 +13,21 @@ namespace vecmem {
...
 
    template< typename TYPE, typename ALLOC >
    VECMEM_HOST
    data::vector_view< const TYPE >
    get_data( const std::vector< TYPE, ALLOC >& vec ) {
 
-      return { vec.size(), vec.data() };
+      return {
+         static_cast<
+            typename data::vector_view< const TYPE >::size_type >( vec.size() ),
```
paulgessinger, line 29: This is because the size_types can be different?

paulgessinger (commented):
Cool, this looks like the payoff of 1 through 3.

core/include/vecmem/containers/device_vector.hpp
krasznaa: As long as every thread is only making use of `push_back(...)` and/or `emplace_back(...)`, that should be thread-safe. But if any other thread is trying to use any of the non-thread-safe functions at the same time, then these will also not work correctly.

I guess that does fall under thread safe still. :thinking: Since on any class thread-safe and non-thread-safe functions would have that behaviour... So the wording here could indeed be changed. Do you have a proposal?

paulgessinger: Yeah, probably just "thread safe".

But what about the capacity? Is it fixed in the context in which we consider this thread safe?

core/include/vecmem/containers/impl/device_vector.ipp
krasznaa: This is something that we pondered about with @stephenswat as well early on.

In device code the only way in which a function like `at(...)` can fail is either through an `assert(...)`, or by performing some invalid memory operation explicitly. It does not have the option of throwing an exception, as those are "not a thing" in device code.

So... I definitely wanted to keep both the `at(...)` functions and the `[]` operators, to be as source-compatible with `std::vector` as possible. But I just couldn't come up with a way of properly mimicking the "safety" of `std::vector::at(...)` in `vecmem::device_vector::at(...)`. We should indeed reflect this in our documentation...

core/include/vecmem/containers/device_vector.hpp
krasznaa: This is a tricky little thing... With the device code the const-ness of something can not be enforced through the const-ness of the "high level" objects. Since we always pass `vecmem::data::vector_view` types by value to kernels (that's the only way of doing it really...), you can't express that a vector should be const, through an "external" `const` declaration.

What you can do instead is to use types like `vecmem::data::vector_view<const int>` to show that this is a view of a constant `int` array. (This is why `vecmem::const_device_vector` "is a thing".)

Now... Here I was thinking that the const-ness of the elements should also extend to the const-ness of the size of the vector. I.e. if you pass a `vecmem::data::vector_view<const T>` object to your kernel, that should mean that not only do you not want to allow the kernel to modify the values of the elements in the vector, but also that the kernel should not be allowed to resize the vector.

And that's what I tried to express with this piece of code. :wink:

core/include/vecmem/containers/impl/vector.ipp
krasznaa: Yes. :smile: With this PR I change `vecmem::data::vector_view<T>::size_type` and `vecmem::device_vector<T>::size_type` to `unsigned int`. (From the previous `std::size_t`.) The conversion from `std::size_t` to `unsigned int` happens in these functions "canonically"... (And the compiler would give us warnings about loss of precision if we didn't use an explicit cast.)

core/include/vecmem/containers/impl/device_vector.ipp
paulgessinger: Fair enough.

core/include/vecmem/containers/device_vector.hpp
krasznaa: The only way in which I imagine that we can have resizable vectors is by constructing "vector buffers" in the right way.

The memory allocated by `vecmem::vector` or `vecmem::jagged_vector` (so `std::vector`...) can not be allowed to be "resized" by device code. As propagating that information back to the host vector object would be a **very** fragile operation. (You can't just call `std::vector::resize(...)` on the host object after the kernel finished running, since that would overwrite the memory that the kernel filled up just previously.) It wouldn't be impossible, but it would require very fragile code...

So you would create, in this PR's example, a `vecmem::data::vector_buffer<T>` with a separate "size" and "capacity" parameter. The `vecmem::device_vector<T>` created on top of that buffer will now know that it can change the "size" of the vector, but only up to the buffer's "capacity".

So yes, the capacity is decided in device code, ahead of running the kernel. This is just how device code works. :frowning: You must put restrictions / assumptions around your memory management, the algorithms are not free to use however much memory they want. (If you ever tried programming in Fortran, this is pretty similar to how memory management there works / worked.)

Now... One thing that I really hate in the current code, and which I was very curious if either of you would notice, is how `vecmem::data::vector_buffer<T>` is set up to be resizable. It's a very misleading interface in this PR. :frowning: You should have a look, and then we can discuss afterwards how it could be made better...

core/include/vecmem/containers/device_vector.hpp
paulgessinger: I see your point. Just from looking at it, I would expect `vecmem::data::vector_view<const T>` to let me change the size though. But ok.

core/include/vecmem/containers/impl/vector.ipp
paulgessinger: Ok I see.

core/include/vecmem/containers/device_vector.hpp
krasznaa: Okay, removed the `"kind of"`. :wink:

PR #78, opened 2021-04-22
Additional Jagged Vector Tests, main branch (2021.04.22.)

krasznaa:
In preparation for adding resizable jagged vectors, I first wanted to extend the "regular" testing of jagged vectors for HIP and SYCL devices as well. This is what this PR is doing.

The HIP code is pretty much a carbon-copy of the CUDA tests, but the SYCL one did need to be modified a bit. Since the CUDA and HIP tests are set up to use threads from just one execution block, with an internal synchronisation point inside of the kernels.

But with SYCL that can't be done. :frowning: Not if you want to be able to run the test on a host device as well. Since host devices do not support [sycl::nd_item::barrier](https://intel.github.io/llvm-docs/doxygen/classcl_1_1sycl_1_1nd__item.html#a25470d2b92e6e8df3d83709969ff876b) calls. Mainly because there's no logical way in which they could. :frowning: (On a host device every work item is executed one by one, in the same CPU thread. So the kernel waiting for what "other threads" are doing is nonsensical in such a setup.) So long story short, instead of using a single kernel in the SYCL tests, I've broken up the code into 2 kernels.

What's worrisome is that running the SYCL test on bare metal using oneAPI 2021.2.0, the test fails. :frowning: Even after I updated to the currently latest version of the Intel GPU driver. However the test runs successfully when I run it from inside the [gitlab-registry.cern.ch/akraszna/atlas-gpu-devel-env:centos7-1-4-0](https://gitlab.cern.ch/akraszna/atlas-gpu-devel-env/container_registry/5827) image. (That image **can** run kernels on my GPU as well, it's a matter of starting it with "just the right" flags.)

So I'm now in the process of building the same version of the compiler "on bare metal" that the Docker image holds, to see if it's indeed the compiler that causes a difference. Of course I'll hold off on merging this while I'm still debugging that issue.

### DIFF
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -109,5 +109,5 @@ jobs:
       run: |
         cd build
         source /opt/intel/oneapi/setvars.sh
-        export SYCL_DEVICE_FILTER=gpu,host
+        export SYCL_DEVICE_FILTER=level_zero:gpu,host
         ctest --output-on-failure
--- a/tests/hip/CMakeLists.txt
+++ b/tests/hip/CMakeLists.txt
@@ -14,7 +14,7 @@ include( vecmem-compiler-options-hip )
 # Test all of the HIP library's features.
 vecmem_add_test( hip
    "test_hip_memory_resources.cpp"
-   "test_hip_containers.cpp" "test_hip_containers_kernels.hpp"
-   "test_hip_containers_kernels.hip"
+   "test_hip_containers.cpp" "test_hip_jagged_containers.cpp"
+   "test_hip_containers_kernels.hpp" "test_hip_containers_kernels.hip"
    LINK_LIBRARIES vecmem::core vecmem::hip GTest::gtest_main
                   vecmem_testing_common )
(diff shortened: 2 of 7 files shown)


krasznaa:
:frowning: I'll have to create yet another bug report as it seems.

I can make use of the Level-0 GPU device successfully on the bare metal. It's the OpenCL GPU driver that does not work as expected. :frowning: On either the bare metal, or when running inside the Docker container. (With oneAPI 2021.2 I did not have access to a Level-0 device previously, that's why I saw the issue showing up there...)

krasznaa:
So... I spent some time looking at these new failures. Unfortunately it's a really ugly thing to debug. Since problematic jobs (in this case the newly introduced SYCL tests) "break" the machine that they were executed on. One has to restart the machine to get the Intel iGPU back to working order using the Level-0 backend.

Let me tag @ivorobts. Igor, this is the sort of code that my e-mail was about.

https://github.com/acts-project/vecmem/blob/JaggedVectorTests-main-20210422/tests/sycl/test_sycl_jagged_containers.sycl#L152-L195

And what is able to break my iGPU until a reboot...

With all this said, since the code works fine with the Level-0 driver, and also with CUDA and HIP, I'll just go ahead with this PR...

PR #80, opened 2021-04-23
Extendable Device Vectors - Part 5, Jagged Vectors (2021.04.23.)

krasznaa:
With 1D vectors being resizable, the next step is to allow jagged vectors to be resizable inside of device code as well.

For now I only allow the "inner vectors" of a jagged vector to be resized from device code. It would not be impossible to extend this to the "outer vector" later on as well. But most algorithms that I used so far can live with just this functionality already.

The logic in `vecmem::data::jagged_vector_buffer` is pretty much the same as in `vecmem::data::vector_buffer` for making the managed arrays resizable. Just like in #75, I had to teach `vecmem::copy` how to handle resizable jagged vector buffers as well. But notice that `vecmem::jagged_device_vector` did not actually need to be touched in this PR! :smile:

I still have to write the test for CUDA, and test the SYCL code on actual accelerated hardware. So for now this is just a draft.

### DIFF
--- a/core/include/vecmem/containers/data/jagged_vector_buffer.hpp
+++ b/core/include/vecmem/containers/data/jagged_vector_buffer.hpp
@@ -79,6 +79,21 @@ namespace vecmem { namespace data {
                             memory_resource& resource,
                             memory_resource* host_access_resource = nullptr );
 
+      /// Constructor from a vector of ("inner vector") sizes and capacities
+      ///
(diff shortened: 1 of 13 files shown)


krasznaa:
Surprisingly enough the SYCL code worked on my iGPU as well, and the CUDA code could be written as a carbon-copy of the HIP one. So this is ready at this point...

krasznaa:
I'll just go ahead with this unless somebody stops me in the next hour or so...

PR #83, opened 2021-05-12
Implement a static array type

stephenswat:
This new type represents an array with length known at compile time, very similar to how `std::array` works. The difference between this and `std::array` is that this is designed to be device-code friendly, and the functions have been marked as device compatible where applicable.

Continuation of https://github.com/acts-project/traccc/pull/34.

### DIFF
--- a/core/include/vecmem/containers/impl/static_array.ipp
+++ b/core/include/vecmem/containers/impl/static_array.ipp
@@ -1,0 +1,237 @@
+/**
+ * VecMem project, part of the ACTS project (R&D line)
+ *
+ * (c) 2021 CERN for the benefit of the ACTS project
+ *
+ * Mozilla Public License Version 2.0
+ */
+
+#pragma once
+
+#include "vecmem/utils/types.hpp"
+
+#include <cstddef>
+#include <type_traits>
+#include <stdexcept>
+
+namespace vecmem {
+    template<typename T, std::size_t N>
+    VECMEM_HOST_AND_DEVICE
+    static_array<T, N>::static_array(
+        void
+    ) {
+        /*
+         * This function does quite literally nothing, and leaves the array's
+         * contents uninitialized.
+         */
+    }
+
+    template<typename T, std::size_t N>
+    VECMEM_HOST
+    constexpr
+    typename static_array<T, N>::reference
+    static_array<T, N>::at(
+        size_type i
+    ) {
+        /*
+         * The at function is bounds-checking in the standard library, so we
+         * do a boundary check in our code, too. This makes this method
+         * incompatible with device code.
+         */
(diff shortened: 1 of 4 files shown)


stephenswat:
Wonderful, the SYCL compiler doesn't like my code for some reason. :tada:

paulgessinger:
Other than the build failure, this looks to be in pretty good shape.

stephenswat:
Alright, fixed it. Very strange, smells a little like a compiler bug.

paulgessinger:
I love the smell of compiler bug in the morning.

core/include/vecmem/containers/impl/static_array.ipp
```diff
@@ -0,0 +1,237 @@
...
+    template<typename T, std::size_t N>
+    VECMEM_HOST_AND_DEVICE
+    static_array<T, N>::static_array(
+        void
+    ) {
+        /*
+         * This function does quite literally nothing, and leaves the array's
+         * contents uninitialized.
+         */
+    }
```
krasznaa, line 27: I think we'll need to replace it with a declaration that the default constructor should be `= default`. Otherwise SYCL will not accept such objects as "input variables" to SYCL kernels.

But we can sort that out once we get there...

core/include/vecmem/containers/impl/static_array.ipp
```diff
@@ -0,0 +1,237 @@
...
+        size_type i
+    ) {
+        /*
+         * The at function is bounds-checking in the standard library, so we
+         * do a boundary check in our code, too. This makes this method
+         * incompatible with device code.
+         */
+        if (i >= N) {
+            throw std::out_of_range("Index greater than size of static array.");
+        }
```
krasznaa, line 43: I'm not a fan of this. In all the other classes I forewent using exceptions, and instead just used assertions. Exactly so that they could be used in device code.

Let's keep it like this for now, but I believe we'll want to change this later on. As those assertions, I believe, will come in handy in device code once we start debugging more complicated algorithms. (While the advantage of exceptions will be pretty marginal in host code during those debugging sessions.)

krasznaa (approved):
Let's go with it like this for now, to allow you to go forward with your traccc developments. But I have a few things that we'll need to add afterwards:
  - Iterator support;
  - Device code tests.

Note that the build issue that you ran into, is a Clang issue rather than a SYCL one. I think... I've seen the same previously with `vecmem::device_vector` as well. This is why `vecmem::device_vector::assign` had to be implemented inline. As Clang refused to accept it with that function implemented out-of-line...

https://github.com/acts-project/vecmem/blob/main/core/include/vecmem/containers/device_vector.hpp#L133-L154

PR #89, opened 2021-07-08
vecmem::static_array Improvements, main branch (2021.07.08.)

krasznaa:
As I noted back in #83, the current implementation of `vecmem::static_array` is missing some things.

First off, I implemented the missing iterator returning and size checking functions on the class. Plus I also added `vecmem::static_array::fill(...)`, as it seemed like it could be useful. (And since there's an [std::array::fill(...)](https://en.cppreference.com/w/cpp/container/array/fill) function as well.)

Then I tried to set up some additional tests for the type, including some minimal tests in device code. And I quickly had to realise that the implementation of the "main constructor" of the class required C\+\+17. Which CUDA and HIP were non too happy with. So after some thinking, I came up with the included re-write of the constructor. I'm also not super happy about it, but I couldn't come up with anything much better with just the C\+\+14 syntax. (The current code is definitely nicer. But if we can't use it in device code in all conditions, then it's no good...)

I also renamed `vecmem::static_array::v` to `vecmem::static_array::m_array`. Just to keep with the ATLAS coding conventions. (Which we are partially following with VecMem, even if not in all aspects.)

I also updated the class to use `vecmem::details::array_type` for the type of `vecmem::static_array::m_array`. Even though I just couldn't really trigger the same issue with zero sized arrays that I wrote that class for in the first place. (In `vecmem::static_vector`...) But I think that's just because I didn't try hard enough...

P.S. The code formatter also touched a couple of lines that I thought it shouldn't have. Hopefully it won't have some weird hysteresis...

### DIFF
--- a/core/include/vecmem/containers/impl/static_array.ipp
+++ b/core/include/vecmem/containers/impl/static_array.ipp
@@ -59,7 +59,7 @@ static_array<T, N>::operator[](size_type i) {
      * Non-bounds checking access, which could cause a segmentation
      * violation.
      */
-    return v[i];
+    return m_array[i];
 }
 
 template <typename T, std::size_t N>
@@ -68,7 +68,7 @@ static_array<T, N>::operator[](size_type i) const {
     /*
      * Return an element as constant.
      */
-    return v[i];
+    return m_array[i];
 }
 
 template <typename T, std::size_t N>
@@ -77,7 +77,7 @@ static_array<T, N>::front(void) {
     /*
      * Return the first element.
      */
-    return v[0];
+    return m_array[0];
 }
 
 template <typename T, std::size_t N>
@@ -86,7 +86,7 @@ static_array<T, N>::front(void) const {
     /*
      * Return the first element, but it's const.
      */
-    return v[0];
+    return m_array[0];
 }
 
 template <typename T, std::size_t N>
@@ -95,7 +95,7 @@ static_array<T, N>::back(void) {
     /*
      * Return the last element.
      */
-    return v[N - 1];
+    return m_array[N - 1];
 }
 
 template <typename T, std::size_t N>
@@ -104,7 +104,7 @@ static_array<T, N>::back(void) const {
     /*
      * Return the last element, but it's const.
      */
-    return v[N - 1];
+    return m_array[N - 1];
 }
 
 template <typename T, std::size_t N>
@@ -113,7 +113,7 @@ static_array<T, N>::data(void) {
     /*
      * Return a pointer to the underlying data.
      */
-    return v;
+    return m_array;
 }
 
 template <typename T, std::size_t N>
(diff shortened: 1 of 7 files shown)


core/include/vecmem/containers/impl/static_array.ipp
```diff
@@ -122,20 +122,141 @@ static_array<T, N>::data(void) const {
...
+VECMEM_HOST_AND_DEVICE constexpr typename static_array<T, N>::reverse_iterator
+static_array<T, N>::rend() {
+
+    return reverse_iterator(begin());
+}
+
+template <typename T, std::size_t N>
+VECMEM_HOST_AND_DEVICE constexpr
+    typename static_array<T, N>::const_reverse_iterator
+    static_array<T, N>::rend() const {
```
krasznaa, line 203: I just noticed this now... We'll definitely need to tweak `clang-format` a little further... :stuck_out_tongue:

core/include/vecmem/containers/impl/static_array.ipp
```diff
@@ -122,20 +122,141 @@ static_array<T, N>::data(void) const {
...
-template <typename std::size_t... Is, typename... Tp>
-VECMEM_HOST_AND_DEVICE static_array<T, N>::static_array(
-    std::index_sequence<Is...>, Tp&&... a) {
-    /*
-     * Construct the array from an enumerated list of elements of arbitrary
-     * size. This is a fold expression which iterates over the elements of
-     * Is (the indices) and Tp (the values), and inserts them into the inner
-     * array using a comma expression.
-     */
-    (static_cast<void>(v[Is] = a), ...);
```
stephenswat, line 138: What's wrong with this constructor? I don't see why we would need to get rid of it, I think it is even used in traccc.

core/include/vecmem/containers/static_array.hpp
```diff
@@ -62,12 +92,18 @@ class static_array {
...
      */
-    template <
-        typename... Tp, typename = std::enable_if_t<sizeof...(Tp) == N>,
-        typename = std::enable_if_t<(std::is_convertible_v<Tp, T> && ...)> >
-    VECMEM_HOST_AND_DEVICE static_array(Tp &&... a)
-        : static_array(std::index_sequence_for<Tp...>(),
-                       std::forward<Tp>(a)...) {}
+    template <typename... Tp, typename = std::enable_if_t<sizeof...(Tp) == N>,
+              typename = std::enable_if_t<
+                  details::conjunction_v<std::is_convertible<Tp, T>...> > >
```
stephenswat, line 97: What is the advantage over using this conjunction over fold syntax?

core/include/vecmem/containers/impl/static_array.ipp
krasznaa: [Fold expressions](https://en.cppreference.com/w/cpp/language/fold) were only introduced with C\+\+17. When I tried to compile the HIP code using the compiler that we use in the CI, I was getting:

```
In file included from /data/ssd-1tb/projects/vecmem/vecmem/tests/hip/test_hip_containers_kernels.hip:19:
In file included from /data/ssd-1tb/projects/vecmem/vecmem/core/include/vecmem/containers/static_array.hpp:305:
/data/ssd-1tb/projects/vecmem/vecmem/core/include/vecmem/containers/impl/static_array.ipp:254:42: warning: pack fold expression is a C++17 extension [-Wc++17-extensions]
    (static_cast<void>(m_array[Is] = a), ...);
                                         ^
```

As I wrote in the description, I had to come up with a C\+\+14 alternative. I was thinking for a while how to do that, but I could only come up with a recursive function call...

core/include/vecmem/containers/static_array.hpp
stephenswat: Oh, is this the C++17 issue?

stephenswat:
Sorry, I was pressing the "Add single comment" button instead of the "Start review" button. What a mess! :cry:

core/include/vecmem/containers/impl/static_array.ipp
```diff
@@ -122,20 +122,141 @@ static_array<T, N>::data(void) const {
...
+}
+
+template <typename T, std::size_t N>
+template <typename Tp1, typename... Tp>
+VECMEM_HOST_AND_DEVICE void static_array<T, N>::static_array_impl(size_type i,
+                                                                  Tp1&& a1,
+                                                                  Tp&&... a) {
+
+    m_array[i] = a1;
+    static_array_impl(i + 1, std::forward<Tp>(a)...);
```
stephenswat, line 251: Honestly I don't think this is too big of a problem. This method is most likely used only for relatively small arrays, so that in combination with tail-call elimination will hopefully make this perform quite well. Hopefully.

krasznaa:
> Sorry, I was pressing the "Add single comment" button instead of the "Start review" button. What a mess!

I **was** wondering why I'm getting all of these individual comments. :stuck_out_tongue_closed_eyes:

core/include/vecmem/containers/static_array.hpp
```diff
@@ -167,17 +203,96 @@ class static_array {
...
+    void fill(const_reference value);
+
+    /// @}
+
     private:
     /**
      * @brief Private helper-constructor for the parameter pack constructor.
      *
      * HACK: This template pack is defined as std::size_t instead of
      * size_type because the SYCL compiler refuses to accept it otherwise.
```
stephenswat, line 286: This little hack warning no longer applies!

stephenswat (approved):
It all looks fairly sensible to me. Of course it is a shame to lose all the fancy C++17 goodness, but indeed we have to support C++14 as well... Have we tested if traccc still compiled with these changes?

krasznaa:
> Have we tested if traccc still compiled with these changes?

Let me give that a try. Though I'd be very surprised if it didn't. (We have a reasonable set of tests for the class by now.)

core/include/vecmem/containers/impl/static_array.ipp
paulgessinger: Sure. In all honesty though, this is an absolute beast of a function signature to format. Just changing the return type to `auto` probably already helps a lot.

paulgessinger: Or even going with 

```cpp
template <typename T, std::size_t N>
VECMEM_HOST_AND_DEVICE constexpr auto
    static_array<T, N>::rend() const -> const_reverse_iterator {
```

PR #90, opened 2021-07-09
Use `auto func() -> return_t` signature to improve legibility

paulgessinger:
As I mentioned in a PR comment, while clang-format sometimes struggles with formatting complex function signatures, there is a way to reduce the signature complexity and improve legibility by using `auto func() -> return_t` signature. This is possible because the type behind `->` is evaluated **in the scope of the method**, as opposed to the classical return type.

This PR does this for all cases where it's possible (and that I saw), mostly in `ipp` files, as in class signature declaration the scope is obvious anyway.

Let me know what you think.

### DIFF
--- a/core/include/vecmem/containers/impl/array.ipp
+++ b/core/include/vecmem/containers/impl/array.ipp
@@ -34,9 +34,7 @@ allocate_array_memory(vecmem::memory_resource& resource,
 
 /// Helper function used in the @c vecmem::array constructors
 template <typename T, std::size_t N>
-std::unique_ptr<typename vecmem::array<T, N>::value_type,
-                typename vecmem::array<T, N>::deleter>
-initialize_array_memory(
+auto initialize_array_memory(
     std::unique_ptr<typename vecmem::array<T, N>::value_type,
                     typename vecmem::array<T, N>::deleter>
         memory,
@@ -93,7 +91,7 @@ array<T, N>::array(memory_resource& resource, size_type size)
 }
 
 template <typename T, std::size_t N>
-typename array<T, N>::reference array<T, N>::at(size_type pos) {
+auto array<T, N>::at(size_type pos) -> reference {
 
     if (pos >= m_size) {
         throw std::out_of_range("Requested element " + std::to_string(pos) +
@@ -104,7 +102,7 @@ typename array<T, N>::reference array<T, N>::at(size_type pos) {
 }
 
 template <typename T, std::size_t N>
-typename array<T, N>::const_reference array<T, N>::at(size_type pos) const {
+auto array<T, N>::at(size_type pos) const -> const_reference {
 
     if (pos >= m_size) {
         throw std::out_of_range("Requested element " + std::to_string(pos) +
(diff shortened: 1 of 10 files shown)


krasznaa (commented):
This is super subjective, but I'm not a fan of this formalism. :frowning: I see what you're saying with it requiring fewer characters, and that's true. Still, I'm not a huge fan of it...

But I'm curious to know what @stephenswat thinks.

paulgessinger:
It's requiring fewer characters, while being safer, and avoiding the clang-format issues, largely.

stephenswat:
I'm very much in two minds about this. From my point of view, the benefits of this approach are:

1. Requires fewer characters.
2. Improves the formatting through `clang-format`.
3. Closer to Haskell-style function signatures that I prefer.
4. Type names are arguably clearer since they are resolved in the scope of the class.
5. Trailing return types are strictly more powerful, so it makes sense from a consistency point of view to use them everywhere, even if not necessary.

However, there are also some very clear downsides:

1. Trailing return type syntax is much less common, meaning people reading our code might not understand it and development environments might have a harder time supporting it.
2. It's a slippery slope towards `auto` return types (without trailing specification), which I consider to be evil and would like to avoid wherever possible.
3. Unless we convert all our other code, this will leave us with a rather inconsistent mix of function signature styles.

In the end, none of these are really game-changers, as far as I am concerned. I am generally a fan of trailing return syntax, and I would argue in favour of them in general due to their increased expressive power. But I am a little more skeptical about introducing them just to appease `clang-format`. As far as I am concerned, while this change makes the signatures vastly more readable, the signatures are still horribly formatted. I would rather tweak the formatting configuration to ensure it outputs readable function signatures.

paulgessinger:
Tweaking formatting configuration inevitably leads to pain and suffering, and you make hysteresis much much more likely (been there).

But ok, it's 2:1 even though I think it both fixes the formatting issues here and is much easier to read and write. I'll close this then.

stephenswat:
No, no, hold on. I would still be in favour of pulling this in for readability!

paulgessinger:
Just to clarify, this usage of `auto` with a trailing `->` is actually safer then the previous version, because there's no chance of mismatch between the types. Apart from this, I've observed in ACTS that this style improves compilation memory usage and speed, because the compiler only has to instantiate the template once for each definition.

stephenswat:
> But are you now saying Stephen that we should update every single source file to this coding style?

Not quite - I just think it's a funny thought that we would be making our code more inconsistent by mixing and matching signature styles, in order to make the code more legible after formatting, while we introduced the formatter to make our code more consistent. :smile: 

But I should come down from my consistency high horse - sometimes you need to settle for a bit of inconsistency for the good of the project. In my opinion the advantage of this MR is clear: the signatures of these methods are a lot more readable. Since Paul has already done the work I personally don't see a reason not to go ahead with this, unless you really dislike the syntax, Attila.

krasznaa:
Fine... Let's go ahead like this.

PR #92, opened 2021-07-15
Clang Format Update, main branch (2021.07.15.)

krasznaa:
Tweaked the indentation of the class access modifiers as I hinted at earlier.

Now... Normally my preference would've been to use `AccessModifierOffset: -4`, and have classes look like:

```c++
class MyClass {
public:
    MyClass();
};
```

But as long as we want to stay close to Google's coding format, I thought this would be more appropriate. Since by default Google uses a layout like:

```c++
class MyClass {
 public:
  MyClass();
};
```

I thought that since we use a tab width of 4 spaces instead of Google's default of 2, it would only be fair to double the indentation on the access modifiers as well.

Thoughts?

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -9,5 +9,5 @@ AllowShortIfStatementsOnASingleLine: false
 AllowShortLoopsOnASingleLine: false
 PointerAlignment: Left
 ColumnLimit: 80
-AccessModifierOffset: 0
+AccessModifierOffset: -4
 KeepEmptyLinesAtTheStartOfBlocks: true
--- a/core/include/vecmem/containers/array.hpp
+++ b/core/include/vecmem/containers/array.hpp
@@ -39,7 +39,7 @@ static constexpr std::size_t array_invalid_size = static_cast<std::size_t>(-1);
 template <typename T, std::size_t N = details::array_invalid_size>
 class array {
 
-    public:
+public:
     /// @name Type definitions, mimicking @c std::array
     /// @{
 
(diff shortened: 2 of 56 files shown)


paulgessinger:
I'm ok with this. @stephenswat ?

stephenswat:
I'm getting a little worried that I'm going to drive this whole project apart by disagreeing with these formatting changes, but I must admit I personally don't like this change. :sweat_smile: 

In my opinion, if the indent is set to 4, having lines start at column 2 is really confusing, not just for humans, but also for editors. I'm pretty sure the editors I use would have a hard time compensating for this, and they would either assume the document was indented with 2 spaces, or I would have to manually add or remove spaces against the wishes of my editor's automatic indenting.

With that said, it's not a big deal to me. If this change makes the two of you happy then I am more than willing to put aside my objections and go ahead with this. :smile:

krasznaa:
As I wrote, my first choice would've been `AccessModifierOffset: -4`. Should we then switch to that instead...? :thinking:

stephenswat:
I would be in complete agreement with that!

paulgessinger:
Still don't mind, so let's do it.

stephenswat:
Looking at the documentation, shouldn't we do this via `IndentAccessModifiers=false`?

krasznaa:
> Looking at the documentation, shouldn't we do this via `IndentAccessModifiers=false`?

Ha! That was my first idea as well. :stuck_out_tongue: But that flag does not exist in clang-format-11 yet... :frowning_face: 

https://releases.llvm.org/11.0.0/tools/clang/docs/ClangFormatStyleOptions.html

stephenswat (approved):
Good to go as far as I am concerned! :+1:

PR #93, opened 2021-08-06
Add More Warning Checks, main branch (2021.08.06.)

krasznaa:
I have been getting a very annoying compilation warning from @konradkusiak97's Acts code, as it is building VecMem. This is because Acts uses `-Wshadow` in its build (https://github.com/acts-project/acts/blob/main/cmake/ActsCompilerOptions.cmake#L7), which triggered a warning in our code.

So in the end I decided to adopt 2 flags from Acts (`-Wshadow` and `-Wunused-local-typedefs`) into the VecMem build. And then to fix the warning brought forward by `-Wshadow`.

### DIFF
--- a/cmake/vecmem-compiler-options-cpp.cmake
+++ b/cmake/vecmem-compiler-options-cpp.cmake
@@ -18,6 +18,8 @@ if( ( "${CMAKE_CXX_COMPILER_ID}" MATCHES "GNU" ) OR
    foreach( mode RELEASE RELWITHDEBINFO MINSIZEREL DEBUG )
       vecmem_add_flag( CMAKE_CXX_FLAGS_${mode} "-Wall" )
       vecmem_add_flag( CMAKE_CXX_FLAGS_${mode} "-Wextra" )
+      vecmem_add_flag( CMAKE_CXX_FLAGS_${mode} "-Wshadow" )
+      vecmem_add_flag( CMAKE_CXX_FLAGS_${mode} "-Wunused-local-typedefs" )
    endforeach()
 
    # More rigorous tests for the Debug builds.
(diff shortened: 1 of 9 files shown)


stephenswat (approved):
I'm all for more warnings!

krasznaa:
**Very** interesting. Apparently GCC 7 considers variable names to shadow function names. And complains that in the code `to` is being used for both in a given function. :confused: 

Fine, changing the internal variable name is no biggie. Still, I was not expecting this... (I only checked in my local build that the SYCL, HIP and CUDA compilers would be fine with these changes. I didn't test the whole suite of C\+\+ compilers that we use.)

krasznaa:
I take it back. Making the changes in the "copy classes" was a biggie after all... :angry:

stephenswat:
So - just out of curiosity - is it only GCC 7 that generates this warning? Later versions do not?

krasznaa:
> So - just out of curiosity - is it only GCC 7 that generates this warning? Later versions do not?

It seems so. If you check, the "CUDA build" failed in that attempt while building a source file from the core library. I fully believe that that's purely because of the C\+\+ compiler version.

Locally I was updating the code on Ubuntu 20.04, and on that (with GCC 9) this warning did not show up. Neither did it with the [gitlab-registry.cern.ch/akraszna/atlas-gpu-devel-env:centos7-1-5-0](https://gitlab.cern.ch/akraszna/atlas-gpu-devel-env/container_registry/5827) image, which comes with GCC 8.

But other than these, I did not do any thorough checks. I just went ahead and renamed a whole bunch of variables... :stuck_out_tongue:

krasznaa:
I was waiting a bit, in case Paul had any comments. But I can't imagine that he would object anyway...

paulgessinger:
Yeah no, more warnings is more better :)

PR #94, opened 2021-08-25
Jagged Vector Improvements, main branch (2021.08.25.)

krasznaa:
As @konradkusiak97 found during his updates in the Acts SYCL seedfinder code, the current implementation of resizable `vecmem::data::jagged_vector_buffer` objects is far from ideal. This PR aims to improve upon that...

First off, the internal memory layout of these buffers is updated like the following:

image

Instead of scattering the "size variables" of the resizable jagged vectors across the allocated blob, now these size variables are put at the beginning of the blob, before all of the payloads of the "internal vectors".

With that update in place, I taught `vecmem::copy` to be able to grab the sizes of resizable jagged vectors with a single operation, minimising the number of copy operations needed to copy a jagged vector. The number of `memset` operations was also reduced to just 1 per jagged vector from previously executing N. (Where N is the size of the "outer vector".)

We managed to learn a bit about GCC and "generic lambdas" with @stephenswat along the way. :smile:

### DIFF
--- a/core/include/vecmem/utils/impl/copy.ipp
+++ b/core/include/vecmem/utils/impl/copy.ipp
@@ -31,8 +31,8 @@ void copy::setup(data::vector_view<TYPE>& data) {
               data.size_ptr(), 0);
     VECMEM_DEBUG_MSG(2,
                      "Prepared a device vector buffer of capacity %u "
-                     "for use on a device",
-                     data.capacity());
+                     "for use on a device (ptr: %p)",
+                     data.capacity(), static_cast<void*>(data.size_ptr()));
 }
 
 template <typename TYPE>
@@ -98,6 +98,25 @@ void copy::operator()(const data::vector_view<TYPE1>& from_view,
 }
 
 template <typename TYPE>
+typename data::vector_view<TYPE>::size_type copy::get_size(
+    const data::vector_view<TYPE>& data) {
+
+    // Handle the simple case, when the view/buffer is not resizable.
+    if (data.size_ptr() == nullptr) {
+        return data.capacity();
+    }
+
+    // If it *is* resizable, don't assume that the size is host-accessible.
+    // Explicitly copy it for access.
+    typename data::vector_view<TYPE>::size_type result = 0;
+    do_copy(sizeof(typename data::vector_view<TYPE>::size_type),
+            data.size_ptr(), &result, type::unknown);
+
+    // Return what we got.
+    return result;
+}
+
+template <typename TYPE>
 void copy::setup(data::jagged_vector_buffer<TYPE>& data) {
 
     // Check if anything needs to be done.
@@ -108,7 +127,18 @@ void copy::setup(data::jagged_vector_buffer<TYPE>& data) {
     // "Set up" the inner vector descriptors, using the host-accessible data.
     for (typename data::jagged_vector_buffer<TYPE>::size_type i = 0;
          i < data.m_size; ++i) {
-        setup(data.host_ptr()[i]);
+        // Find the first "inner vector" that has a non-zero capacity, and is
+        // resizable.
+        if ((data.host_ptr()[i].capacity() != 0) &&
+            (data.host_ptr()[i].size_ptr() != nullptr)) {
+            // Initialise the size values using this "inner vector's" pointer.
+            do_memset(sizeof(typename data::vector_buffer<TYPE>::size_type) *
+                          (data.m_size - i),
+                      data.host_ptr()[i].size_ptr(), 0);
+            // This would have initialised all of the "inner vectors" correctly,
+            // we can end the loop.
+            break;
+        }
     }
 
     // Check if anything else needs to be done.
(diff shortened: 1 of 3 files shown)


core/include/vecmem/utils/impl/copy.ipp
```diff
@@ -108,7 +127,18 @@ void copy::setup(data::jagged_vector_buffer<TYPE>& data) {
...
          i < data.m_size; ++i) {
-        setup(data.host_ptr()[i]);
+        // Find the first "inner vector" that has a non-zero capacity, and is
+        // resizable.
+        if ((data.host_ptr()[i].capacity() != 0) &&
+            (data.host_ptr()[i].size_ptr() != nullptr)) {
+            // Initialise the size values using this "inner vector's" pointer.
+            do_memset(sizeof(typename data::vector_buffer<TYPE>::size_type) *
+                          (data.m_size - i),
+                      data.host_ptr()[i].size_ptr(), 0);
```
paulgessinger, line 137: So this way it just sets the entire size block to 0 with a single memset, right?

core/include/vecmem/utils/impl/copy.ipp
```diff
@@ -347,21 +401,32 @@ void copy::copy_views(std::size_t size,
 }
 
 template <typename TYPE>
-typename data::vector_view<TYPE>::size_type copy::get_size(
-    const data::vector_view<TYPE>& data) {
+std::vector<typename data::vector_view<TYPE>::size_type> copy::get_sizes(
+    const data::vector_view<TYPE>* data, std::size_t size) {
```
paulgessinger, line 405: Is `size` here the *outer size*?

paulgessinger (commented):
I'll have to go through in some more detail I suppose. But here are two comments right away?

core/include/vecmem/utils/impl/copy.ipp
krasznaa: Yes.

I may have overcomplicated the code here a bit, but I wasn't entirely sure whether empty (zero capacity) inner vectors would have a valid "size pointer" or not. Since we do need to check for that in `vecmem::copy::get_size(...)`. So to be safe, I added this for-loop to find the first inner vector that would have a valid "size pointer", and then perform the initialisation of the sizes based on that.

core/include/vecmem/utils/impl/copy.ipp
krasznaa: It is.

I introduced this **private** function mainly because in the `vecmem::copy::copy_views(...)` function we are receiving such `vecmem::data::vector_view` pointers already. And since I wanted to use `get_sizes(...)` there, it had to provide an interface that I could use.

The public functions just receive `vecmem::data::jagged_vector_view` and `vecmem::data::jagged_vector_buffer` references. So I thought that for the internal function I could go without writing too much documentation. :wink:

core/include/vecmem/utils/impl/copy.ipp
paulgessinger: Hm, I'm not sure I follow yet. Is this because the size of the size block is implicitly given as the location of the first non-empty vector? Or why to you need the vector's size pointers for this?

core/include/vecmem/utils/impl/copy.ipp
paulgessinger: Right, just trying to follow what's going on. I agree that private functionality doesn't have to be commented, although it would help me with reviewing of course 😅

core/include/vecmem/utils/impl/copy.ipp
krasznaa: In the current implementation resizable (jagged) vectors all start from zero sizes. Which is very misleading, I very much agree. Since the UI of these classes suggests that we can create the buffers with a non-zero size, and a capacity larger/equal to that size. But in fact, at the moment the starting sizes of resizable (jagged) vectors are always zero. (We can talk about the reasons for this in more detail in let's say an "issue".)

Since the resizable (jagged) vectors are assumed to be empty at the start, we really only need to make sure that in the allocated (device) memory the "size variables" would be initialised to zero. It doesn't matter how the memory which will hold the payload, is initialised. Since we'll write some values explicitly to that area as we fill the vector.

`size_ptr()` is what returns the pointer to the size variable in "global memory" on the device. This is what we can use to read/write the variable from the host code, and this is also what the device code uses to atomically interact with that variable.

Previously the code didn't assume anything about the internal memory layout of `vecmem::data::jagged_vector_buffer`. It initialised the sizes of the inner vectors one-by-one. Leading to a tremendous number of `memset(...)` operations, as @konradkusiak97 observed. Now I decided to hardcode knowledge into `vecmem::copy` about what the memory layout of the buffer would be. With this slightly non-trivial code... 😛

krasznaa:
It would be nice to move ahead on this one finally. I'm very sure about myself with the changes, but you guys should still have a quick look at least.

There will be some further quality-of-life changes coming for the benefit of the Acts seedfinder code, but I don't want to add them to this PR, complicating it even further...

stephenswat (approved):
I don't know enough about this part of the code to provide much meaningful commentary but it seems free of any obvious problems. :+1:

PR #97, opened 2021-09-06
Fix inclusion of PMR headers on C++14

stephenswat:
It appears that the logic we use to determine whether we can import the "real" or the experimental version of the standard PMR library is a little broken. It does not seem to work on C++14, probably because `__has_include` is a C++17 feature. However, I am not one hundred percent sure what the exact problem is here. Adding these additional checks on the C++ version seems to resolve the problem.

This commit fixes the issues outlined in acts-project/traccc#78.

### DIFF
--- a/core/include/vecmem/memory/memory_resource.hpp
+++ b/core/include/vecmem/memory/memory_resource.hpp
@@ -14,13 +14,13 @@
  * depending on the GCC version used, so we try to unify them by aliassing
  * depending on the compiler feature flags.
  */
-#if __has_include(<memory_resource>)
+#if __cplusplus >= 201703L && __has_include(<memory_resource>)
 #include <memory_resource>
 
 namespace vecmem {
 using memory_resource = std::pmr::memory_resource;
 }
-#elif __has_include(<experimental/memory_resource>)
+#elif __cplusplus == 201402L || (__cplusplus >= 201703L && __has_include(<experimental/memory_resource>))
 #include <experimental/memory_resource>
 
 namespace vecmem {
--- a/core/include/vecmem/memory/polymorphic_allocator.hpp
+++ b/core/include/vecmem/memory/polymorphic_allocator.hpp
@@ -14,14 +14,14 @@
  * depending on the GCC version used, so we try to unify them by aliassing
  * depending on the compiler feature flags.
  */
-#if __has_include(<memory_resource>)
+#if __cplusplus >= 201703L && __has_include(<memory_resource>)
 #include <memory_resource>
 
 namespace vecmem {
 template <typename T>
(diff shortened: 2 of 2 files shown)


stephenswat:
It must be noted that this is not a very good solution. The import errors disappear, but it does not stop `std::pmr::memory_resource` and `std::experimental::pmr::memory_resource` from being different resources, so linker errors are likely.

krasznaa (commented):
Let me do some tests with the Acts Docker image(s).

I've been long unhappy with these `__has_include` statements to be honest. I think I'll try to set up something using [try_compile(...)](https://cmake.org/cmake/help/latest/command/try_compile.html) instead. Though that will also be a bit tricky, so I'll have to see how feasible that is. (The problem is that we'd need to decide during the configuration of the core library which include / namespace to use. Taking into account all of the languages and C\+\+ standards that we'll be using.)

stephenswat:
I think the ugly problem is that the result of `__has_include` can differ between the compiler used to build vecmem (especially if you have the binaries installed, and you are not building it from source) and the compiler used to build the downstream project, which includes these headers.

If one of them imports the experimental version and the other one doesn't your entire compilation breaks down because the types don't match.

krasznaa:
Hence my idea to use `try_compile(...)`. As that could take into account the "downstream project's" C\+\+ settings correctly, to find out if it should use the C\+\+14 or C\+\+17 classes in all situations from there on.

But let me actually do some tests...

stephenswat:
Right, but that would still break down in cases where vecmem binaries are pre-installed, and not built as part of another project. Ideally we would build vecmem twice, once with experimental PMR and one with non-experimental PMR, and expose one of the two built versions through CMake magic.

krasznaa:
:confused:  No... We can expose with CMake which C\+\+ version was used to build the vecmem libraries. We should add that to their exported configuration. So that clients would report problems if they tried to use a different standard than what vecmem was built with.

Let's not go crazy over this issue. We won't be building multiple versions of vecmem just because we can't make sure that clients would compile their code consistently with how vecmem was compiled.

This is the exact issue why we have LCG releases. Why we decided that externals would be compiled with a consistent set of settings, so that clients (like ATLAS TDAQ and offline) would not need to figure out these settings external-by-external.

stephenswat:
The problem then becomes: are later versions of the C++ standard guaranteed to provide the `<experimental/memory_resource>` header even if a non-experimental `<memory_resource>` is available? I don't really know the answer to that question, but if they are not then none of these solutions really bring us any further... :disappointed:

At this point I am not sure supporting C++14 is a sustainable development model for this at all.

paulgessinger:
> Edit: I don't actually think they do at all.

Also pretty sure they don't/won't.

PR #98, opened 2021-09-06
Memory Resource Source Detection Update, main branch (2021.09.06.)

krasznaa:
This is my take on what we should do in response to https://github.com/acts-project/traccc/pull/83, which @stephenswat opened #97 about.

As we discussed with @stephenswat, the `vecmem/memory/memory_resource.hpp` and `vecmem/memory/polymorphic_allocator.hpp` headers can really only be used by a compiler that knows how to handle C\+\+17. CUDA in C\+\+14 mode must not see any of these headers. (In principle these headers should only be used by the "host compiler" in any project anyway.)

To make sure that the build of this project, and any client project, would always use the same source for picking up the memory resource types from, I added some logic to the CMake code to determine at configuration time where these types would be available from. And then the build just forces every compiler to use the types from there.

It's interesting to note that even in our CI, we pick up these types from both of those locations, since `<memory_resource>` only became available in GCC 9. The configuration now provides this type of output as feedback about these choices:

```
-- Performing Test VECMEM_HAVE_PMR_MEMORY_RESOURCE
-- Performing Test VECMEM_HAVE_PMR_MEMORY_RESOURCE - Success
-- Using memory resource types from the std::pmr namespace
```

```
-- Performing Test VECMEM_HAVE_PMR_MEMORY_RESOURCE
-- Performing Test VECMEM_HAVE_PMR_MEMORY_RESOURCE - Failed
-- Performing Test VECMEM_HAVE_EXPERIMENTAL_PMR_MEMORY_RESOURCE
-- Performing Test VECMEM_HAVE_EXPERIMENTAL_PMR_MEMORY_RESOURCE - Success
-- Using memory resource types from the std::experimental::pmr namespace
```

### DIFF
--- a/core/include/vecmem/memory/memory_resource.hpp
+++ b/core/include/vecmem/memory/memory_resource.hpp
@@ -7,6 +7,15 @@
 
 #pragma once
 
+// This header can only be used in C++17 mode. So "device compilers" that
+// can't use C++17, must not see it. Which in practically all cases should
+// "just" be a question of code organisation.
+#if __cplusplus < 201700L
+#error \
+    "This header can only be used in C++17 mode. " \
+         "Ideally it should only be used by the \"host compiler\"."
+#endif  // < C++17
+
 /*
  * The purpose of this file is to provide uniform access (on a source-code
  * level) to the memory_resource type from the standard library. These are
@@ -14,18 +23,18 @@
  * depending on the GCC version used, so we try to unify them by aliassing
  * depending on the compiler feature flags.
  */
-#if __has_include(<memory_resource>)
+#if defined(VECMEM_HAVE_PMR_MEMORY_RESOURCE)
 #include <memory_resource>
 
 namespace vecmem {
 using memory_resource = std::pmr::memory_resource;
 }
-#elif __has_include(<experimental/memory_resource>)
+#elif defined(VECMEM_HAVE_EXPERIMENTAL_PMR_MEMORY_RESOURCE)
 #include <experimental/memory_resource>
 
 namespace vecmem {
 using memory_resource = std::experimental::pmr::memory_resource;
 }
 #else
-#error "vecmem requires C++17 LFTS V1 (P0220R1) component memory_resource!"
+#error "C++17 LFTS V1 (P0220R1) component memory_resource not found!?!"
 #endif
--- a/core/CMakeLists.txt
+++ b/core/CMakeLists.txt
@@ -72,3 +72,37 @@ string( LENGTH "${CMAKE_SOURCE_DIR}/" VECMEM_SOURCE_DIR_LENGTH )
 target_compile_definitions( vecmem_core PUBLIC
    $<BUILD_INTERFACE:VECMEM_DEBUG_MSG_LVL=${VECMEM_DEBUG_MSG_LVL}>
    $<BUILD_INTERFACE:VECMEM_SOURCE_DIR_LENGTH=${VECMEM_SOURCE_DIR_LENGTH}> )
+
+# Figure out where to get <memory_resource> from.
+include( CheckCXXSourceCompiles )
+check_cxx_source_compiles( "
+   #include <memory_resource>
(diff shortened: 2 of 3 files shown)


core/include/vecmem/memory/memory_resource.hpp
```diff
@@ -7,25 +7,34 @@
...
 
+// This header can only be used in C++17 mode. So "device compilers" that
+// can't use C++17, must not see it. Which in practically all cases should
+// "just" be a question of code organisation.
+#if __cplusplus < 201700L
+#error \
+    "This header can only be used in C++17 mode. " \
+         "Ideally it should only be used by the \"host compiler\"."
+#endif  // < C++17
+
```
stephenswat, line 18: I still don't really understand why we have this. If a compiler supports memory resources for C++14 then why shouldn't we allow that to work? It seems a little ham-fisted to me to check for this explicitly, but in the end it's not a big deal.

stephenswat (approved):
Looks fine to me.

core/include/vecmem/memory/memory_resource.hpp
krasznaa: What we check during the CMake configuration in the new version is that the host compiler, with whatever c\+\+ standard it uses, would find these headers/types.

But at that point we may still mistakenly include this header in a CUDA source file, with CUDA set to cpp14. In which case, without this explicit check, they will just get an error that the header is missing. This message is meant to give users a hint, that they were not supposed to include this file in the place where they did.

At least this was my thinking with this...

stephenswat (commented):
Actually, thinking about it a little more, I am not convinced this check is sufficiently powerful. If we want to make this robust I think we should check at _include time_ whether these headers are available and proceed if and only if the value for the include time matches exactly the value that was generated at _compile time_. Otherwise you will still get rather hard-to-read errors.

krasznaa:
> Actually, thinking about it a little more, I am not convinced this check is sufficiently powerful. If we want to make this robust I think we should check at _include time_ (or at compile time in the dependent project) whether these headers are available and proceed if and only if the value for the include time matches exactly the value that was generated at _compile time_. Otherwise you will still get rather hard-to-read errors.

This is just not a setup that we should worry about at the moment.

If somebody compiled vecmem with GCC 7 and then proceeds to compile another project against it with GCC 9, they deserve to get some not completely trivial error messages.

Note that they will still get compilation, and not linking (or heaver forbid runtime) errors with this code. So I think the level of robustness like this is just fine...

stephenswat:
But isn't that exactly the setup and scenario that caused us all of this hassle today? An experimental version was available in one TU, and a non-experimental version was available in another TU, causing a bunch of linker errors. If we could protect against that at compile time, wouldn't that be a valuable check?

core/include/vecmem/memory/memory_resource.hpp
stephenswat: That makes sense, but wouldn't it be more robust, in that case, to either check only against `__CUDACC__`, or check against both `__CUDACC__` _and_ `__cplusplus`?

krasznaa: I definitely do not want to check for `__CUDACC__` here. What would be the logic anyway?

If CUDA with C\+\+17 doesn't manage to find the same version of these types as the host compiler (which would be very weird btw, telling us that CUDA is not being used correctly), then getting the error message about a missing header file should be "good enough".

stephenswat: Okay, works for me. Let's leave it as it is then. :slightly_smiling_face:

krasznaa:
With this code the compilation will fail just fine. It will say that the appropriate header is not found. This is indeed not spoon-feeding the issue to the user, but I fear that if we try to provide error messages for every erroneous scenario, we will just be wasting our time. This is the sort of thing that could instead be added to some documentation later on, in the form of "what to do if X happens". 🤔

Issue #96, opened 2021-08-26
Size type of `vecmem::jagged_device_vector` should be `unsigned int`


stephenswat:
This is to bring it in sync with the jagged device vector. Having these two classes have different size types is unlikely to cause any real problems, but it is a little bit of sloppy design that I think we should fix.

PR #101, opened 2021-09-14
Resolve MSVC Warnings, main branch (2021.09.13.)

krasznaa:
As @stephenswat could see it ahead of time, #100 set some things in motion... (I'll be setting up a CI for Windows next. :stuck_out_tongue:)

Here I enabled roughly the same level of warnings for MSVC that we use with GCC and Clang as well. I **did not** use `/Wall` as the warning level, as that seemed a bit ridiculous. Even the Microsoft documentation recommends against using that. (With that enabled, even the test code from https://github.com/acts-project/vecmem/blob/main/core/CMakeLists.txt#L79-L84 would not compile. So at that point I have up on that warning level.)

Then I set out to fix all of the warnings that MSVC would print. Which are mostly integer type issues. MSVC really doesn't like it when you implicitly convert between different integer types... Since, as is discussed in #96, we use `std::size_t` as the size type in some places, and `unsigned int` in others, I had to make the code very explicit about type conversions in a good number of places. I'm not super happy about this either, but at the same time it did make the code a lot more explicit about what it expects the compiler to do...

I was pondering about https://github.com/acts-project/vecmem/blob/main/tests/core/test_core_contiguous_memory_resource.cpp for a bit. Since this test fails with MSVC in Debug builds. For multiple reasons actually. First off, in a Debug build it's a runtime error to de-reference `std::vector<T>::end()`. But even if I don't do that in the code, the test still fails. As in Debug mode MSVC pads all STL containers by some amount. To make it easier to detect out-of-range accesses in the code.

I was wondering about replacing `std::vector` with simple allocation/de-allocation calls in the tests, but in the end rather decided to just disable it for Debug builds with MSVC. But I'm not completely happy with this setup either...

I still maintain that listening to warnings from yet another compiler is a good thing! So I still very much want to go forward with this, even with all the necessary changes. (Note that the library code itself had to be changed very little. It was mostly the tests that needed adjustments.)

### DIFF
--- a/tests/core/test_core_memory_resources.cpp
+++ b/tests/core/test_core_memory_resources.cpp
@@ -78,8 +78,9 @@ protected:
 
         // Remove a couple of elements from the vectors.
         for (int i : {26, 38, 25}) {
-            std::remove(reference_vector.begin(), reference_vector.end(), i);
-            std::remove(test_vector.begin(), test_vector.end(), i);
+            (void)std::remove(reference_vector.begin(), reference_vector.end(),
+                              i);
+            (void)std::remove(test_vector.begin(), test_vector.end(), i);
         }
         // Make sure that they are still the same.
         EXPECT_EQ(reference_vector.size(), test_vector.size());
--- a/tests/cuda/test_cuda_memory_resources.cpp
+++ b/tests/cuda/test_cuda_memory_resources.cpp
@@ -109,8 +109,9 @@ protected:
 
         // Remove a couple of elements from the vectors.
         for (int i : {26, 38, 25}) {
-            std::remove(reference_vector.begin(), reference_vector.end(), i);
-            std::remove(test_vector.begin(), test_vector.end(), i);
+            (void)std::remove(reference_vector.begin(), reference_vector.end(),
+                              i);
+            (void)std::remove(test_vector.begin(), test_vector.end(), i);
         }
         // Make sure that they are still the same.
         EXPECT_EQ(reference_vector.size(), test_vector.size());
--- a/cmake/vecmem-compiler-options-cpp.cmake
+++ b/cmake/vecmem-compiler-options-cpp.cmake
@@ -30,4 +30,15 @@ if( ( "${CMAKE_CXX_COMPILER_ID}" MATCHES "GNU" ) OR
    # More rigorous tests for the Debug builds.
    vecmem_add_flag( CMAKE_CXX_FLAGS_DEBUG "-Werror" )
    vecmem_add_flag( CMAKE_CXX_FLAGS_DEBUG "-pedantic" )
+
+elseif( "${CMAKE_CXX_COMPILER_ID}" MATCHES "MSVC" )
+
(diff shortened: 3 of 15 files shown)


tests/core/test_core_memory_resources.cpp
```diff
@@ -78,8 +78,9 @@ class core_memory_resource_test
 
         // Remove a couple of elements from the vectors.
         for (int i : {26, 38, 25}) {
-            std::remove(reference_vector.begin(), reference_vector.end(), i);
-            std::remove(test_vector.begin(), test_vector.end(), i);
+            (void)std::remove(reference_vector.begin(), reference_vector.end(),
+                              i);
+            (void)std::remove(test_vector.begin(), test_vector.end(), i);
```
stephenswat, line 83: I think this is the only part I don't like, I'd go for `static_cast<void>(...)` in this case.

tests/cuda/test_cuda_memory_resources.cpp
```diff
@@ -109,8 +109,9 @@ class cuda_host_accessible_memory_resource_test
 
         // Remove a couple of elements from the vectors.
         for (int i : {26, 38, 25}) {
-            std::remove(reference_vector.begin(), reference_vector.end(), i);
-            std::remove(test_vector.begin(), test_vector.end(), i);
+            (void)std::remove(reference_vector.begin(), reference_vector.end(),
+                              i);
+            (void)std::remove(test_vector.begin(), test_vector.end(), i);
```
stephenswat, line 114: Same here.

stephenswat (approved):
I think this is great, the mixed integers are a real potential problem, and it's good to make these conversions explicit. In the long time as you say you would want to change the underlying types so they become compatible, but for now it's fine.

Regarding the contiguous tests, yes they are indeed fragile, and perhaps the tests are fundamentally badly designed. We should revisit them sometimes. I like the idea of moving to "raw" allocations, rather than using the resource through a vector.

tests/core/test_core_memory_resources.cpp
krasznaa: I don't know... I'm not trying to cast the results into anything. I'm just trying to explicitly state that I don't care about the return value. Since MSVC is very vocal about these functions being `[[nodiscard]]`... Which, by the way, is not standard as far as I can see... 😦 

https://en.cppreference.com/w/cpp/algorithm/remove

From all the examples I've seen so far, everybody uses this `(void)` notation in such cases. I haven't seen anybody using `static_cast<void>(...)` for this... 😕

stephenswat: It really makes no sense to me why MSVC would mark this `[[nodiscard]]` in their standard library. The return value of this function seems... Not particularly useful.

Anyway, that doesn't matter. In the end, `(void)` and `static_cast<void>` are equivalent, and both will do the job. In my opinion, the `static_cast` version is more C++-like, although I must admit that the C version is more readable.

To be honest I am not sure it matters. Let's merge it in as is.

PR #104, opened 2021-09-19
CMake 3.21+ Compatibility, main branch (2021.09.19.)

krasznaa:
As it turns out, the names of some of the internal variables that the SYCL code interacts with, have changed. Now the code should work correctly with CMake 3.21+ as well.

At the same time I also had to realise that the Visual Studio pre-processor is not able to deal with the `VECMEM_DEBUG_MSG` macros. 😦 I currently blame the compiler, and not our code over this, but this will need some further checks later on...

### DIFF
--- a/cmake/sycl/CMakeSYCLInformation.cmake
+++ b/cmake/sycl/CMakeSYCLInformation.cmake
@@ -6,8 +6,13 @@
 
 # Tweak the linker flags on Windows, to make them compatible with DPC++.
 if( WIN32 )
-   set( CMAKE_CREATE_WIN32_EXE "-Xlinker /subsystem:windows" )
-   set( CMAKE_CREATE_CONSOLE_EXE "-Xlinker /subsystem:console" )
+   if( ${CMAKE_VERSION} VERSION_LESS 3.21 )
+      set( CMAKE_CREATE_WIN32_EXE "-Xlinker /subsystem:windows" )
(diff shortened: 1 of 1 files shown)


paulgessinger:
What's the problem you're observing with `VECMEM_DEBUG_MSG`?

krasznaa:
> What's the problem you're observing with `VECMEM_DEBUG_MSG`?

MSVC does not understand the macros that I implemented. :frowning: It generates invalid code with them. I checked this in a simple Visual Studio project, by copy-pasting the macros into it, and trying to use them. In the GUI I could get the compiler to tell me exactly what code it put together based on the macros. Which was just invalid. :frowning:

Based on https://docs.microsoft.com/en-us/cpp/preprocessor/variadic-macros I want to try what happens when I add `/Zc:preprocessor` as a compiler option. I just didn't get around to trying it yet...

stephenswat:
You can halt most compilers after the preprocessing step to view the preprocessed code only. For MSVC that appears to be the `/P` flag. Might be useful in debugging and macro errors.

krasznaa:
> You can halt most compilers after the preprocessing step to view the preprocessed code only. For MSVC that appears to be the `/P` flag. Might be useful in debugging and macro errors.

I was lazy. Instead of learning the compiler option, and then executing the correct command in the correct subdirectory, just copy-pasting the code into a Visual Studio window seemed a lot simpler in this case... :stuck_out_tongue: But yeah, with any GCC/Clang preprocessor issue the solution does have to be a command-line one usually...

PR #105, opened 2021-09-23
Add a polyfill for `get_default_resource`

paulgessinger:
As observed in https://github.com/acts-project/detray/issues/102 and elsewhere, if no explicit memory resource is given to e.g. the vector types, compilation will fail on libc++ because `get_default_resource` and related are not implemented. While this is not an encouraged usage pattern for vecmem, it still makes sense to technically allow these constructs.

This PR adds an additional check if these parts of LFTS are available, and if not adds a minimal polyfill that adds implementations for

- `set_default_resource`
- `new_delete_resource` (almost identical to `vecmem::host_memory_resource`, but reimplemented to match the spec and avoid a dependency cyclce)
- `get_default_resource`

I tried to make sure `set_default_resource` and `get_default_resource` are thread-safe, as is demanded in the spec.

I also added a test to ensure this (albeit discouraged usage pattern) compiles.

Let me know what you think.


core/include/vecmem/memory/memory_resource.hpp
```diff
@@ -38,3 +38,54 @@ using memory_resource = std::experimental::pmr::memory_resource;
...
+}
+
+}  // namespace detail
+
+inline memory_resource* new_delete_resource() noexcept {
+    static detail::new_delete_resource_impl res{};
+    return &res;
+}
+
+inline memory_resource* get_default_resource() noexcept {
```
krasznaa, line 80: :confused: Does this really work? `libc++` does define this function.

https://github.com/llvm/llvm-project/blob/main/libcxx/include/experimental/memory_resource#L147

I know that technically this definition does not clash with it, since the actual function in `libc++` is **not** in the `std::experimental::pmr` namespace, but rather in `std::experimental::fundamentals_v1::pmr`. (It is only pulled into the `std::pmr::experimental` namespace by a `typedef`/`using` statement.)

All in all, I have a lot of worries about this implementation...

krasznaa (changes requested):
I am definitely not in favour of making our `memory_resource.hpp` header use even more macros. Implementing these missing symbols should really be hidden behind the scenes, in a `.cpp` file.

The way that you look for the problem in CMake is fine. But when the error is detected, CMake should add one more source file to the core library, and not put yet another public definition on the library.

core/include/vecmem/memory/memory_resource.hpp
```diff
@@ -38,3 +38,54 @@ using memory_resource = std::experimental::pmr::memory_resource;
...
+
+    virtual void do_deallocate(void* p, std::size_t, std::size_t) override {
+        free(p);
+    }
+
+    virtual bool do_is_equal(
+        const memory_resource& other) const noexcept override {
+        return &other == this;
+    }
+};
```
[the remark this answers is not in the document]
paulgessinger: `host_memory_resource` depends on that header, also the spec says that `new_delete_resource` is singleton and `p->is_equal(r)` returns `&r == p`, rather than the fancy stuff we do in `host_memory_resource`.

core/include/vecmem/memory/memory_resource.hpp
paulgessinger: It works on macOS for me with Clang12. What's the alternative?

core/CMakeLists.txt
```diff
@@ -101,8 +103,25 @@ else()
...
+         ${pmr_ns}::set_default_resource(nullptr);
+         ${pmr_ns}::memory_resource* mr = ${pmr_ns}::get_default_resource();
+         return 0;
+   }
+  " VECMEM_HAVE_DEFAULT_RESOURCE)
+
+if(NOT VECMEM_HAVE_DEFAULT_RESOURCE)
+    target_compile_definitions( vecmem_core PUBLIC
+        VECMEM_NOT_HAVE_DEFAULT_RESOURCE )
+endif()
```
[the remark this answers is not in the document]
paulgessinger: I don't know maybe. I can add the EOLs for you.

paulgessinger:
I can do this in a separate `cpp` file, sure.

core/include/vecmem/memory/memory_resource.hpp
paulgessinger: Honestly, the underlying problem is that we're using `std::experimental::pmr` at all, and begin to polyfill by doing the typedef in `memory_resource.hpp`. Once you start doing that, all bets are off. 

Overall we'd be better off not doing that, and simply not suporting libc++.

paulgessinger: I think the reason why this combination of symbols works is because `fundamentals_v1` is an `inline namespace`.

I changed the polyfill to put the symbols in `inline namespace fundamentals_v1`, which is more consistent, maybe.

core/src/memory/default_resource_polyfill.cpp
```diff
@@ -0,0 +1,57 @@
...
+
+    virtual void do_deallocate(void* p, std::size_t, std::size_t) override {
+        free(p);
+    }
+
+    virtual bool do_is_equal(
+        const memory_resource& other) const noexcept override {
+        return &other == this;
+    }
+};
```
krasznaa, line 27: Now that it's in a `.cpp` file, there's nothing preventing you from just using `vecmem::host_memory_resource`, right?

core/src/memory/default_resource_polyfill.cpp
```diff
@@ -0,0 +1,57 @@
+#include <atomic>
```
krasznaa, line 1: You didn't put a copyright statement into the file...

core/src/memory/default_resource_polyfill.cpp
```diff
@@ -0,0 +1,57 @@
...
+    static std::atomic<memory_resource*> res{new_delete_resource()};
+    return res;
+}
+
+}  // namespace
+
+memory_resource* new_delete_resource() noexcept {
+    static new_delete_resource_impl res{};
+    return &res;
+}
```
krasznaa, line 39: Is this symbol also needed to be able to default-construct vectors? Or why did you implement it? (While leaving out some other functions, that are also not implemented in `libc++` I imagine.)

krasznaa (commented):
I was expecting you to argue that implementing the functions in the header would allow the compiler to do additional optimisations while constructing vectors without an explicit memory resource... :thinking: Since I'm pretty sure that GCC implements it this way because of that. (Or didn't you take the implementation from some existing place?)

But I'm happier with having these implemented in a `.cpp` file, so definitely let's not reverse course now. :stuck_out_tongue:

core/include/vecmem/memory/memory_resource.hpp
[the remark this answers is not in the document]
krasznaa: We can update `vecmem::host_memory_resource`. Though the singleton stuff I'm not too sure about... :frowning:

core/src/memory/default_resource_polyfill.cpp
paulgessinger: Only `vecmem::host_memory_resource::do_equal`, which the spec says should return `this == &other`. `vecmem::host_memory_resource::do_equal` does something different, which I believe ultimately results in the same result boolean. If I supply this as `new_delete_resource`, I'd probably want it to match `this == &other`, up to you if you want me to modify `vecmem::host_memory_resource::do_equal` to do that, in which case I can remove `new_delete_resource_impl`.

core/src/memory/default_resource_polyfill.cpp
paulgessinger: Ah yeah, sure.

core/include/vecmem/memory/memory_resource.hpp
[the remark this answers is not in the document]
paulgessinger: As said in the other comment, up to you if you want me to update `vecmem::host_memory_resource::do_equal` and get rid of `new_delete_resource_impl`.

paulgessinger:
> I was expecting you to argue that implementing the functions in the header would allow the compiler to do additional optimisations while constructing vectors without an explicit memory resource... 🤔 Since I'm pretty sure that GCC implements it this way because of that. (Or didn't you take the implementation from some existing place?)

I mean, we discourage this pattern anyway because that's not what the library is about, so I largely don't care about the performance of this 😄

core/src/memory/default_resource_polyfill.cpp
krasznaa: :thinking: I don't know... Yes, the documentation does say that the standard implementation of the object returned by [new_delete_resource()](https://en.cppreference.com/w/cpp/memory/new_delete_resource) should implement its `is_equal(...)` check through pointer comparison. But I fail to see that as a good design. :confused:

This comparison function to me meant that the runtime could decide with it whether some memory resources could cross-manage each others' resources. And two different instances of `vecmem::host_memory_resource` can very much do that. While two different instances of (for instance) `vecmem::cuda::device_memory_resource` not necessarily. So it makes (in my mind) perfect sense to allow non-trivial implementations for this comparison function.

I see the implementation logic in what the standard says. Since the standard does not define a public type that would manage memory through `new` / `delete`, it only provides the `new_delete_resource()` function, the pointer comparison is "fine". Since you are not supposed to be able to instantiate the underlying resource multiple times.

But I just hate that part of the design... :frowning: These helper functions make sense to have, to make sure that default constructed containers could "do something". But I don't agree with the part of the design that disallows creating "host memory resources" on the stack...

So long story short, I would really like to keep using `vecmem::host_memory_resource` as is, and provide a static instance of it through `new_delete_resource()`. (Is this latter also not provided by `libc++`? Did you check? Should we maybe provide these functions one-by-one, after having checked for their existence one-by-one?)

paulgessinger: I checked on macOS at least, and `new_delete_resource()` seems not to be supplied, so needs to be implemented.

paulgessinger: Done.

krasznaa (approved):
Once the CI finishes, let's finally push it in...

PR #108, opened 2021-09-24
MSVC Fixes and CI, main branch (2021.09.24.)

krasznaa:
With a lot of trial-and-error, but finally I managed to set up a CI for Windows with MSVC. Since #105 made the Windows build fail, it was quite timely to finally have this set up...

Let me go through the modifications:
  - The `VECMEM_HAVE_DEFAULT_RESOURCE` test kept failing on Windows. I was already suspicious previously of the unused variable in the test code possibly giving us problems down the line (I bumped into the same sort of issue in https://github.com/root-project/root/pull/4561 back in the day.), I just didn't realise that it would be with the Windows build. :stuck_out_tongue: After eliminating the unused variable warning/error, the test now works on all platforms.
  - The test introduced by #105 performed an implicit `int` -> `float` conversion. Which MSVC really doesn't like...
  - As discussed in #104, MSVC was not happy with the setup of the variadic macros in `debug.hpp`. I played around with different pre-processor behaviour setting flags, but those would only make things worse. :frowning: (At one point making the compiler issue warnings about system headers as well. :scream:) So I decided to bite the bullet, and provide a different implementation for MSVC. Which is the same implementation that I made initially btw. (Before turning on warnings from GCC about it not liking that implementation.)
  - I was trying my best to add the Windows CI build with the least amount of code lines. But unfortunately I couldn't avoid defining a different CMake configuration command than what we use for the other platforms. On Windows we just have to use the `"NMake Makefiles"` generator, as the SYCL configuration doesn't work with MSBuild. :frowning: (Since MSBuild doesn't do anything with `.sycl` files, the compilation test that we perform to check whether the C\+\+ compiler understands SYCL, is no-op. So with MSBuild the MSVC compiler is always accepted as a valid SYCL compiler... :facepalm:)
    - If you know how to define something like an environment variable differently for different platforms, then this part could be simplified a bit. For now I just defined two separate CMake configuration steps.
    - Note that I also replaced `$GITHUB_WORKSPACE` with `${{ github.workspace }}`. Even though in the end this was not necessary, once it became clear that I'd have to implement the CMake configuration command separately for the platforms. Still, using this formalism is a more platform independent way of referring to the workspace directory. :wink:

### DIFF
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -21,7 +21,7 @@ jobs:
       matrix:
         BUILD_TYPE: [ "Release", "Debug" ]
         DEBUG_MSG_LVL: [ 0, 5 ]
-        OS: ["ubuntu-latest", "macos-10.15"]
+        OS: ["ubuntu-latest", "macos-10.15", "windows-latest"]
 
     # The system to run on.
     runs-on: ${{ matrix.OS }}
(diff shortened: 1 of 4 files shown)


krasznaa:
Let's push this in, as I don't like being blind to possible issues coming from MSVC...

PR #110, opened 2021-09-24
Copy Improvements, main branch (2021.09.24.)

krasznaa:
We bumped into this issue with @konradkusiak97 in https://github.com/acts-project/acts/pull/998. The current implementation of `vecmem::copy` is not able to copy from a constant vector into a non-constant one.

Initially I thought that I may be able to get away with writing this type of code:

```c++
    template <typename TYPE>
    void operator()(data::vector_view<const TYPE> from,
                    data::vector_view<TYPE> to,
                    type::copy_type cptype = type::unknown);
```

Since a view of a non-const vector can be translated into a view of a constant one. But switching to passing the vector views by value in these copy functions would've made their usage **very** inconvenient as it turns out. :frowning: As for instance `vecmem::data::vector_buffer` objects would not be auto-converted into `vecmem::data::vector_view` objects.

So I went with the brute-force method, making the functions use two template types, and only making sure with a `static_cast` that these types would be compatible with each other. Not the absolute best, but that's life.

I also made sure that the `vecmem::copy::to(...)` functions could be conveniently called with constant vectors. Which was also not the case previously...

### DIFF
--- a/core/include/vecmem/utils/copy.hpp
+++ b/core/include/vecmem/utils/copy.hpp
@@ -16,6 +16,7 @@
 
 // System include(s).
 #include <cstddef>
+#include <type_traits>
 #include <vector>
 
 namespace vecmem {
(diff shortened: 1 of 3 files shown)


krasznaa:
With no comments coming in, I'm leaning towards merging this in by the end of the day...

Issue #107, opened 2021-09-24
const experssion for `vecmem::static_array`


beomki-yeo:
While integrating vecmem functionality into detray, I've found that vecmem::static_array doesn't support the const expression - is it intended or just missing it?

paulgessinger:
Hmm, other than `fill`, I believe all other methods are at least marked `constexpr`. Where do you encounter problems with this?

beomki-yeo:
One of the cases is at the mask types:

https://github.com/acts-project/detray/blob/b6426b91a663e923bf33056acbdfecebafe0aeb8/core/include/masks/annulus2.hpp#L70-L73

krasznaa:
Hi Beomki,

Please be a lot more verbose. What is the exact issue that you're running into? What is the exact compiler error that you get?

We did not really test how well `vecmem::array` behaves as a `constexpr` object. (Which is what you seem to be doing there.) So yes, there could very well be things that we should improve. But we'd need a much more exact description of what is not working. Mainly because I myself do not have a lot of experience with writing classes that provide good `constexpr` behaviour.

Attila

beomki-yeo:
Sorry for insufficient explanation.
Following is the compiler error message when I change the data type of `darray` from `std::array` to `vecmem::static_array` in https://github.com/acts-project/detray/blob/b6426b91a663e923bf33056acbdfecebafe0aeb8/plugins/algebra/include/detray/array/include/plugins/algebra/array_definitions.hpp#L11

```
/home/beomki/projects/detray/detray/core/include/masks/annulus2.hpp:70:37: error: the type ‘const mask_tolerance’ {aka ‘const vecmem::static_array<double, 2>’} of ‘constexpr’ variable ‘detray::annulus2<>::within_epsilon’ is not literal
     static constexpr mask_tolerance within_epsilon = {
                                     ^~~~~~~~~~~~~~
In file included from /home/beomki/projects/detray/detray/extern/algebra-plugins/core/common/include/algebra/common/types.hpp:17,
                 from /home/beomki/projects/detray/detray/extern/algebra-plugins/core/include/array/include/algebra/definitions/std_array.hpp:10,
                 from /home/beomki/projects/detray/detray/plugins/algebra/include/detray/array/include/plugins/algebra/array_definitions.hpp:2,
                 from /home/beomki/projects/detray/detray/tests/benchmarks/array/array_intersect_surfaces.cpp:8:
[...]

PR #111, opened 2021-09-26
Constexpr Capability For vecmem::static_array, main branch (2021.09.26.)

krasznaa:
Made it possible to create `constexpr` static_array objects, by declaring the existing constructors as `constexpr`. The only tricky part was that GCC needed the `m_array` member to be explicitly initialised in this case. Luckily an empty initialisation was enough, as it's not possible to meaningfully initialise that variable in the constructor itself.

I think it's interesting to note (just because I myself didn't even think about it beforehand) that declaring any `constexpr` functions on a class is really only meaningful if it has `constexpr` constructors. Otherwise those function can never be used as `constexpr` anyway...

At the same time added some basic tests for this functionality as well. Though we'll probably want to re-organise the tests a bit at one point, as at the moment they are a bit too all over the place...

Of course this is all to address the issue brought up by @beomki-yeo in #107.

### DIFF
--- a/tests/core/test_core_containers.cpp
+++ b/tests/core/test_core_containers.cpp
@@ -156,4 +156,21 @@ TEST_F(core_container_test, static_array) {
     EXPECT_EQ(test_array4.size(), 0);
     EXPECT_EQ(test_array4.max_size(), 0);
     EXPECT_TRUE(test_array4.empty());
+
+    constexpr vecmem::static_array<int, 0> test_array5;
+    constexpr auto test_array5_size = test_array5.size();
+    EXPECT_EQ(test_array5_size, 0);
+    constexpr auto test_array5_max_size = test_array5.max_size();
+    EXPECT_EQ(test_array5_max_size, 0);
+    constexpr bool test_array5_empty = test_array5.empty();
+    EXPECT_TRUE(test_array5_empty);
+
+    constexpr vecmem::static_array<int, ARRAY_SIZE> test_array6 = {
+        0, 1, 2, 3, 4, 5, 6, 7, 8, 9};
+    constexpr auto test_array6_size = test_array6.size();
+    EXPECT_EQ(test_array6_size, ARRAY_SIZE);
+    constexpr auto test_array6_max_size = test_array6.max_size();
+    EXPECT_EQ(test_array6_max_size, ARRAY_SIZE);
+    constexpr bool test_array6_empty = test_array6.empty();
+    EXPECT_FALSE(test_array6_empty);
 }
--- a/core/include/vecmem/containers/impl/static_array.ipp
+++ b/core/include/vecmem/containers/impl/static_array.ipp
@@ -15,13 +15,10 @@
 #include "vecmem/utils/types.hpp"
 
 namespace vecmem {
+
 template <typename T, std::size_t N>
-VECMEM_HOST_AND_DEVICE static_array<T, N>::static_array(void) {
-    /*
-     * This function does quite literally nothing, and leaves the array's
-     * contents uninitialized.
-     */
-}
+VECMEM_HOST_AND_DEVICE constexpr static_array<T, N>::static_array(void)
+    : m_array() {}
 
 template <typename T, std::size_t N>
 VECMEM_HOST constexpr auto static_array<T, N>::at(size_type i) -> reference {
@@ -236,9 +233,8 @@ VECMEM_HOST_AND_DEVICE void static_array<T, N>::fill(const_reference value) {
 
 template <typename T, std::size_t N>
 template <typename Tp1, typename... Tp>
-VECMEM_HOST_AND_DEVICE void static_array<T, N>::static_array_impl(size_type i,
-                                                                  Tp1&& a1,
-                                                                  Tp&&... a) {
+VECMEM_HOST_AND_DEVICE constexpr void static_array<T, N>::static_array_impl(
+    size_type i, Tp1&& a1, Tp&&... a) {
 
     m_array[i] = a1;
     static_array_impl(i + 1, std::forward<Tp>(a)...);
@@ -246,8 +242,8 @@ VECMEM_HOST_AND_DEVICE void static_array<T, N>::static_array_impl(size_type i,
 
 template <typename T, std::size_t N>
 template <typename Tp1>
-VECMEM_HOST_AND_DEVICE void static_array<T, N>::static_array_impl(size_type i,
-                                                                  Tp1&& a1) {
+VECMEM_HOST_AND_DEVICE constexpr void static_array<T, N>::static_array_impl(
+    size_type i, Tp1&& a1) {
 
     m_array[i] = a1;
 }
--- a/core/include/vecmem/containers/static_array.hpp
+++ b/core/include/vecmem/containers/static_array.hpp
@@ -73,7 +73,7 @@ public:
      * uninitialized.
      */
     VECMEM_HOST_AND_DEVICE
-    static_array(void);
+    constexpr static_array(void);
 
     /**
      * @brief Construct an array from a parameter pack of arbitrary size.
@@ -95,7 +95,7 @@ public:
     template <typename... Tp, typename = std::enable_if_t<sizeof...(Tp) == N>,
               typename = std::enable_if_t<
                   details::conjunction_v<std::is_convertible<Tp, T>...> > >
-    VECMEM_HOST_AND_DEVICE static_array(Tp &&... a) {
+    VECMEM_HOST_AND_DEVICE constexpr static_array(Tp &&... a) : m_array() {
 
         static_array_impl(0, std::forward<Tp>(a)...);
     }
@@ -283,13 +283,15 @@ private:
      * @brief Private helper-constructor for the parameter pack constructor.
      */
     template <typename Tp1, typename... Tp>
-    VECMEM_HOST_AND_DEVICE void static_array_impl(size_type i, Tp1 &&a1,
-                                                  Tp &&... a);
+    VECMEM_HOST_AND_DEVICE constexpr void static_array_impl(size_type i,
+                                                            Tp1 &&a1,
+                                                            Tp &&... a);
     /**
      * @brief Private helper-constructor for the parameter pack constructor.
      */
     template <typename Tp1>
-    VECMEM_HOST_AND_DEVICE void static_array_impl(size_type i, Tp1 &&a1);
+    VECMEM_HOST_AND_DEVICE constexpr void static_array_impl(size_type i,
+                                                            Tp1 &&a1);
 
     /// Array holding the container's data
     typename details::array_type<T, N>::type m_array;


tests/core/test_core_containers.cpp
```diff
@@ -156,4 +156,21 @@ TEST_F(core_container_test, static_array) {
     EXPECT_EQ(test_array4.size(), 0);
     EXPECT_EQ(test_array4.max_size(), 0);
     EXPECT_TRUE(test_array4.empty());
+
+    constexpr vecmem::static_array<int, 0> test_array5;
+    constexpr auto test_array5_size = test_array5.size();
```
paulgessinger, line 161: Maybe 

```cpp
static_assert(test_array5_size == 0)
```

to make sure it's actually executed at compile time?

Same below.

beomki-yeo:
Looks very  good to me! Glad that it didn't require many changes in source codes

tests/core/test_core_containers.cpp
krasznaa: :confused: I added these `test_array5_size` variables exactly to force the evaluation of these functions at compile time. The only question here is whether we prefer to see errors show up during the running of the unit test like:

```
[ RUN      ] core_container_test.static_array
/home/krasznaa/ATLAS/projects/vecmem/vecmem/tests/core/test_core_containers.cpp:162: Failure
Expected equality of these values:
  test_array5_size
    Which is: 0
  1
[  FAILED  ] core_container_test.static_array (0 ms)
```

, or during the compilation, like:

```
[ 53%] Building CXX object tests/core/CMakeFiles/vecmem_test_core.dir/test_core_containers.cpp.o
/home/krasznaa/ATLAS/projects/vecmem/vecmem/tests/core/test_core_containers.cpp: In member function ‘virtual void core_container_test_static_array_Test::TestBody()’:
/home/krasznaa/ATLAS/projects/vecmem/vecmem/tests/core/test_core_containers.cpp:162:5: error: static assertion failed
     static_assert(test_array5_size == 1);
     ^~~~~~~~~~~~~
tests/core/CMakeFiles/vecmem_test_core.dir/build.make:103: recipe for target 'tests/core/CMakeFiles/vecmem_test_core.dir/test_core_containers.cpp.o' failed
```

I'm not sure that the latter is that much more descriptive. Plus, it prevents us from seeing any other test failures. So all in all, I'm leaning towards keeping the test as-is... :thinking:

paulgessinger: I'm not sure assigning to a `constexpr` variable actually forces the compiler to execute the expression at compile time, or if it only suggests it. 

If you're certain it does, than let's keep it like it is, if not let's use `static_assert`.

krasznaa: I'm 99.9% sure that the compiler would let us know if it's not possible to evaluate the

```c++
constexpr auto test_array5_size = test_array5.size();
```

line at compile time. (Unless there's a bug in the compiler.) I do not believe that `constexpr` is a "suggestion" in that sense. If you declared something as a `constexpr`, the compiler better check that it can in principle comply.

I don't even see a situation when a compiler should meaningfully decide that evaluating a `constexpr` variable would be better at runtime for the application's performance. :confused:

Having said this, my main motivation here is consistency. I'm not 1000% sure of how GCC or Clang work internally with `constexpr`. But I don't want to differentiate our tests on just a hunch.

paulgessinger: 99.9% is good enough for me 😋

paulgessinger (approved):
👍

tests/core/test_core_containers.cpp
HadrienG2: > I don't even see a situation when a compiler should meaningfully decide that evaluating a constexpr variable would be better at runtime for the application's performance.

As we learned the hard way in ACTS, compile-time resources are every bit as finite as run-time ones. If a compiler gets the impression that evaluating something at compile time will consume unreasonable CPU time or RAM, and deferring the evaluation at runtime is possible, it is a good thing that it has the option of doing so.

This is doubly true considering that some compilers' constexpr evaluators are _much_ slower than run-time expression evaluation, as they often go for an interpreter-like strategy with lots of instrumentation. The reason is that non-reproducibility and undefined behavior have worse consequences at compile-time than they do at runtime.

krasznaa: Do you know any examples of this? On a theoretical level this is all fine, but given that GCC will happily run out of memory on some of the ATLAS source files, I just fail to see how they would meaningfully make a decision about this particular thing during compilation. :confused:

Producing some simple piece of code that compiles slowly is not much of a challenge. As I learned already during LHC's Run-1, just having a large (like... really large) number of `std::map` operations in a single function can result in a ridiculous memory/CPU usage during the build. So we had to re-design a piece of code back in the day to do certain things at runtime rather than at compile time, just because the price during compilation was enormous. (And the performance hit during runtime was minimal after the re-write.)

But none of that was ever automatic. The only "automatic thing" that landed on my radar not too long ago are these types of things from GCC:

```
/build/atnight/localbuilds/nightlies/Athena/master/athena/PhysicsAnalysis/JetTagging/JetTagValidation/JetTagDQA/src/BTaggingValidationPlots.cxx: In member function 'virtual void JetTagDQA::BTaggingValidationPlots::initializePlots()':
/build/atnight/localbuilds/nightlies/Athena/master/athena/PhysicsAnalysis/JetTagging/JetTagValidation/JetTagDQA/src/BTaggingValidationPlots.cxx:208:8: note: variable tracking size limit exceeded with -fvar-tracking-assignments, retrying without
   void BTaggingValidationPlots::initializePlots(){
        ^~~~~~~~~~~~~~~~~~~~~~~
```

However an example of GCC actually giving up on `constexpr` would be pretty interesting to see.

HadrienG2: [Here's a simple example](https://godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAMzwBtMA7AQwFtMQByARg9KtQYEAysib0QXACx8BBAKoBnTAAUAHpwAMvAFYTStJg1DIApACYAQuYukl9ZATwDKjdAGFUtAK4sGe1wAyeAyYAHI%2BAEaYxHoADqgKhE4MHt6%2BcQlJAkEh4SxRMVy2mPaOAkIETMQEqT5%2BRXaYDskVVQQ5YZHRegqV1bXpDX3twZ353VwAlLaoXsTI7BxoDL2YqrHEANTBBJu0nhCTmyYA7FYaAIKb15vEmARzDJuSZiYAzOcXpwAiJpd/Fx2mxYTGCh2OZwBN1u90eewOk3enx%2BHGmtE4AFZeH4OFpSKhOG5rNZNgpZvNMMczG8eKQCJpUdMANYgDEaAB0ZhObxOAA42XyTlw3mYzPpOJJsQz8ZxeAoQBo6QzpnBYDBEChUCxYnRouRKGhtbqYsAuFwxTRaARovKIBFpRFglUAJ6cWmGtiCADyDFortxvCwIKM4gDpHwd2aADdMPKw2sml5rdKdiVpbQ8BFiC6PFg3bwCMQ8Cx89MqAZgAoAGp4TAAdy9sUY%2BZkghEYnYUlb8iUamlujFBiMKGJln0mflkGmqFiZRWnAAtF68Y1ms4IK4BvVSIFRnkChlEnOt4esgwOvuJsVSi1hieGiUmnPWtUL11CrY7546j1hm/xoU0xknMCx6IWmCLDwaKYlKYYEhwqi8gAbAuSGSJswDIMgmxmpymwQESljWKQmy4IQJBUjSJEeEa9BbOYIqTAWyrMiAkgAJzsicZhIWYUjsXxZjIXy4ocJKpAlmypA4ni8FygqSoBiq6oQEgho6nR%2BoQOpxooEOppmBoRSWtaxC2vaYaOswxD%2Bu6WqegQPp%2BtKQaGMAoZ4hGT54DGcZ4gmyBJoseKpuiYYZlmOYYJBBZFiW3ConwFbVrWDZNjitL8G2ojiF2WU9io6hhroRT6SORFjhFk6HPis7JHGS4ro%2Ba5%2BBuDDuN%2BgxiruuTvhIGKkPER7JPeYpDWe/4HlwA2rs%2BX5pPUYqzbebSTRMM3zT%2B5qfqte59dNQHkqBRTgTF0EcFi0nSvBiEoWhmzIPpOFmOyHJcPhhFWGOpH4EQ9HUkUmw0Rp0SUVMzFKdMCCYEwWAxDVYXiZJioybwcm2Ap9KQ6QLIcVxPF8RxgnCbyolvLBsmyopWiTOdZgU2jVNYzT0wxmZyRsUAA%3D%3D%3D) where both GCC and clang will refrain from evaluating a constexpr at compile time when told that compilation speed is all that matters via `-O0`.

Beyond this, GCC has a notion of "too complicated to constexpr" via this set of flags...

> **-fconstexpr-depth=n**
>
> Set the maximum nested evaluation depth for C++11 constexpr functions to n. A limit is needed to detect endless recursion during constant expression evaluation. The minimum specified by the standard is 512.
>
> **-fconstexpr-loop-limit=n**
>
> Set the maximum number of iterations for a loop in C++14 constexpr functions to n. A limit is needed to detect infinite loops during constant expression evaluation. The default is 262144 (1<<18).
>
> **-fconstexpr-ops-limit=n**
>
> Set the maximum number of operations during a single constexpr evaluation. Even when number of iterations of a single loop is limited with the above limit, if there are several nested loops and each of them has many iterations but still smaller than the above limit, or if in a body of some loop or even outside of a loop too many expressions need to be evaluated, the resulting constexpr evaluation might take too long. The default is 33554432 (1<<25).

...but I don't know what GCC actually does when the limit is reached. Does it simply error out, or does it have the logic in place to back out of const evaluation in that scenario? I personally don't have time to experiment here, but feel free to do so.

In the interest of the GCC developers' sanity, I personally hope that they somehow handle "unnecessary constexpr" as a special case of the general optimizer logic that is used when the compiler attempts to evaluate expressions at compile time : inlining, CSE, constprop, etc... They did so a long time before `constexpr` was a thing, it just happens that `constexpr` is designed to guarantee that if they attempt this optimization, then given enough time it will eventually succeed.

HadrienG2: Oh, well, it wasn't so hard to try, so I did it anyway.

Both GCC and Clang stop trying to evaluate complicated constexpr and revert to runtime evaluation after a while.

- [Clang doesn't even try very hard before giving up](https://godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAMzwBtMA7AQwFtMQByARg9KtQYEAysib0QXACx8BBAKoBnTAAUAHpwAMvAFYTStJg1DIApACYAQuYukl9ZATwDKjdAGFUtAK4sGIAMxmpK4AMngMmAByPgBGmMR6AA6oCoRODB7evgFByamOAmER0SxxCVy2mPYFDEIETMQEmT5%2BgZXV6XUNBEVRsfF6CvWNzdltQ929JWUSAJS2qF7EyOwcaAxDmKqJxADU4QS7MUzoEAe7iQ2suyb%2BACK7ZrM3AOxWGgCCu9/7VLsQl2I11ubluDy4zxMbxMnx%2BcN2xEwBCWDAuVxYt3eXx%2BUIeVSUryx8J%2BiORxFRx1OgNYAFoITdLEcTgD0bsaY9ZpiYdjvrjuXzPtzziwmOEIJDobCSUiUUzTpz/FjcRx5rROABWXh%2BDhaUioTigyzWXYKRbLTAM/w8UgETQq%2BYAaxA6o0ADozC9/C8ABwun0vLiBIJqjiSLV2vWcXgKEAaG12%2BZwWAwRAoVAsRJ0eLkShoDNZhLALhcII0WgEeIxiAxCMxcINACenGtebYggA8gxaE2dbwsCKjOJe6R8IiHHgAG6YGPDraYZBeCsRg5VCO0PAxIHEBseLDN3gEYh4Fj7%2BZUAzABQANTwmAA7u3Eox9zJBCIxOwpK/5Eo1BHdMGhjGNY1j6BuMaQPMqCJDUM40u2/jRlU841C4DDuJ4LR6KE4R9KUAwVHkaQCKMfiESkxEMFM/TlO0KGdMMTSYdkFR2PRAhdI01H4bREwjMxZG2Ix3EzBCCxLCseiHpgqw8KqGrhsO%2BocKo3oAGw0mpki7MAyDILsxbuv8hpWJYpC7LghAkJaFS7B4%2Bb0Hs5iBLMB4Jo6ICSAAnK6LxmGpZhSF5gVmOpPr6JwYakCeLqkNqurKdGsbxr2iYphASB5pmjk5hAWUFigBhGCWGgVGWFbEFWNbDnWzDbi%2BraMAQnbdhG/ZAUOuqjvRU4zrqc4Lkuw4riGurrpuja7rJB5Hie3AqnwF7XreD5Ptq1r8G%2BojiF%2Bm0/io6jDroFRFcBRpmeNEHinqMHpHBCFIR0zgQK4pHYehIkEaQRE1G95H5Okn28ch44cYxf10aDtTCbh0xfXxTFZIJCNA3M4nmlJiLTfJHCanFEbKapGlabsyCnQZZium6XDGSBZkWfgRBOWYVrmfZ2XxDZrkpVoszzAgmAnAM10hlFMVxvFvCJbYyW2qlHneb5/mBd5IVhd6EUcIh%2BNKVGPP2prZiKQlety7z8xTpV6SeUAA%3D%3D)
- GCC tries harder, but [will eventually give up as well](https://godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAMzwBtMA7AQwFtMQByARg9KtQYEAysib0QXACx8BBAKoBnTAAUAHpwAMvAFYTStJg1DIApACYAQuYukl9ZATwDKjdAGFUtAK4sGIAMxmpK4AMngMmAByPgBGmMR6AA6oCoRODB7evgFByamOAmER0SxxCVy2mPYFDEIETMQEmT5%2BgZXV6XUNBEVRsfF6CvWNzdltQ929JWUSAJS2qF7EyOwcaAxDmKqJxADU4QS7MUzoEAe7iQ2suyb%2BACK7ZrM3AOxWGgCCu9/7VLsQl2I11ubluDy4zxMbxMnx%2BcN2xEwBCWDAuVxYt3eXx%2BUIeVSUryx8J%2BiORxFRx1OgNYAFoITdLEcTgD0bsaY9ZpiYdjvrjuXzPtzziwmOEIJDobCSUiUUzTpz/FjcRx5rROABWXh%2BDhaUioTigyzWXYKRbLTAM/w8UgETQq%2BYAaxA6o0ADozC9/C8ABwun0vLiBIJqjiSLV2vWcXgKEAaG12%2BZwWAwRAoVAsRJ0eLkShoDNZhLALhcII0WgEeIxiAxCMxcINACenGtebYggA8gxaE2dbwsCKjOJe6R8IiHHgAG6YGPDraYZBeCsRg5VCO0PAxIHEBseLDN3gEYh4Fj7%2BZUAzABQANTwmAA7u3Eox9zJBCIxOwpK/5Eo1BHdMGhjGNY1j6BuMaQPMqCJDUM40u2/jRlU841C4DDuJ4LR6KE4R9KUAwVHkaQCKMfiESkxEMFM/TlO0KGdMMTSYdkFR2PRAhdI01H4bREwjMxZG2Ix3EzBCCxLCseiHpgqw8KqGrhsO%2BocKo3oAGw0mpki7MAyDILsxbuv8hpWJYpC7LghAkJaFS7B4%2Bb0Hs5iBLMB4Jo6ICSAAnK6LxmGpZhSF5gVmOpPr6JwYakCeLqkNqurKdGsbxr2iYphASB5pmjk5hAWUFigBhGCWGgVGWFbEFWNbDnWzDbi%2BraMAQnbdhG/ZAUOuqjvRU4zrqc4Lkuw4riGurrpuja7rJB5Hie3AqnwF7XreD5Ptq1r8G%2BojiF%2Bm0/io6jDroFRFcBRpmeNEHinqMHpHBCFIR0zgQK4pHYehIkEaQRE1G95H5Okn28ch44cYxf10aDtTCbh0xfXxTFZIJCNA3M4nmlJiLTfJHCanFEbKapGlabsyCnQZZium6XDGSBZkWfgRBOWYVrmfZ2XxDZrkpVoszzAgmAnAM10hlFMVxvFvCJbYyW2qlHneb5/mBd5IVhd6EUcIh%2BNKVGPP2prZiKQlety7z8xTpV6SeUAA%3D%3D).

Anyway, for those who think they know better than the compiler whether something needs to be evaluated at compile time or not, there is always C++20's `consteval`...

HadrienG2: By the way, in case you're curious, the single top reason why compilers go crazy when asked to compile a single very complicated function is that compilers operate at function granularity, and the resource usage of some very important compilation algorithms like register allocation grows super-linearly with the complexity of their input.

For "sane" functions that do little more than calling into other functions that live into separate cpp files, this is not usually a big problem, because compilers can just guesstimate the complexity of compiling the function using heuristics, and stop inlining calees as soon as the extimated complexity grows too large.

The problem, however, is what happens when the function you're calling cannot be "outlined". What if it's just some random inline declaration in a header without an associated "home" cpp flag? Like, say, most class/function template declarations?

In principle, in that event, I imagine the compiler should be able to do the moral equivalent of "outlining" that function anyway, using some sort of static linkage trick so that the build doesn't later fail due to all the duplicate declarations that will inevitably result across multiple compilation units. But in practice, I don't know of a single compiler that does this.

Maybe it's considered bad style to have multiple clones of the same symbol in the output file even if you can hack the linker into accepting to live with it (I can certainly imagine that it would make the work of tools like debuggers and profilers a lot more "interesting"). Or maybe I'm overestimating the capabilities of modern linkers, and it's just plain verboten to link together multiple instances of the same symbol even if they "hidden" from each other via static linkage.

PR #113, opened 2021-09-27
Memory Resource Stress Test, main branch (2021.09.27.)

krasznaa:
Added a stress test for the core memory resources. One that `vecmem::binary_page_memory_resource` is failing at the moment. :frowning:

@stephenswat, this is just meant as a simple demonstrator for the problem that I wrote you about earlier. Hopefully this will allow you/us to understand what is the bug in that memory resource...

There are a few things to point out:
  - If I add an explicit `vecmem::vector<int>::reserve(...)` statement before the filling of the inner vectors (thereby eliminating all the re-allocations in the inner loop), the test succeeds.
  - If the outermost loop is only executed once, even if the inner loops do a lot of (re-)allocations, the test succeeds.
    * So something seems to be going wrong when all memory is de-allocated, and then we start asking for memory again. Just how we observe it in @konradkusiak97's code as well.

### DIFF
--- a/core/src/memory/binary_page_memory_resource.cpp
+++ b/core/src/memory/binary_page_memory_resource.cpp
@@ -13,6 +13,7 @@
 #include <stack>
 
 #include "vecmem/memory/memory_resource.hpp"
+#include "vecmem/utils/debug.hpp"
 
 namespace {
 /**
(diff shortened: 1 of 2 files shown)


stephenswat:
Superb, I'll finish off a few fancy things I am working on now and then get onto debugging this!

stephenswat (approved):
Looks good to me, clearly this is a useful test to have. :smile:

PR #116, opened 2021-09-30
Test Logic Updates, main branch (2021.09.30.)

krasznaa:
Having learned from https://github.com/acts-project/acts/pull/998, this PR tries to make it easier to integrate VecMem into downstream projects.

First off, I introduced the `VECMEM_BUILD_TESTING` flag. Since I had to realise that only depending on `BUILD_TESTING` is not a good idea after all. That one is declared as a CACHE variable in [CTest](https://cmake.org/cmake/help/latest/module/CTest.html). Which means that you can't easily turn it on/off for individual subdirectories in a project. So to allow client projects, that do want to build their own tests, to turn off the VecMem tests, this flag should come in handy.

At the same time, in order to be in line with how [Acts](https://github.com/acts-project/acts) handles its externals, I replaced `vecmem-googletest.cmake` with the `tests/googletest` subdirectory. Whose usage is governed by the `VECMEM_USE_SYSTEM_GOOGLETEST` flag. (To follow the naming used in Acts.)

Note that the logic is not entirely bullet-proof, even so. :frowning: For instance if we try to build VecMem as part of a project that itself builds GoogleTest, and the operating system also provides GoogleTest, then in this setup we can still not easily force VecMem to pick up the GoogleTest build from the "parent" project. Should we maybe add a flag like `VECMEM_SETUP_GOOGLETEST` to control whether VecMem would do anything to set up GoogleTest? (With it being set to `FALSE`, it would just expect that "somebody" would take care of setting up the `GTest::gtest` and `GTest::gtest_main` targets for the build.) Any thoughts? :thinking:


paulgessinger:
Wouldn't you be able to steer that via `CMAKE_PREFIX_PATH`?

tests/CMakeLists.txt
```diff
@@ -5,9 +5,22 @@
...
+# Set up GoogleTest.
+option( VECMEM_USE_SYSTEM_GOOGLETEST
+   "Pick up an existing installation of GoogleTest from the build environment"
+   FALSE )
+if( VECMEM_USE_SYSTEM_GOOGLETEST )
+   find_package( GTest )
+   # If GoogleTest is not found, don't fail right away. It might be that we are
+   # building VecMem as part of another project, which takes care of building
+   # GoogleTest itself. CMake will fail anyway if the GTest:: targets are not
+   # found by the end of the project's configuration.
```
paulgessinger, line 19: I'm not too sure about this. There is always a dependency chain, and some node has no dependencies (correct me if I'm wrong). So in my opinion the parent project just has to make sure the dependencies are set up in a way that the chain works. In that case, we should just fail here.

krasznaa:
> Wouldn't you be able to steer that via `CMAKE_PREFIX_PATH`?

If let's say [traccc](https://github.com/acts-project/traccc) is set up to build GoogleTest, then the GoogleTest headers/libraries would not be available anywhere at the moment that VecMem calls `find_package(GTest)`. No setting of `CMAKE_PREFIX_PATH` would be able to help with that. On the other hand if it **is** available from let's say `/usr` (but not yet built as part of `traccc`), then I don't think we could force that `find_package` call to not find it under `/usr`. At least not in any easy way.

So I'm really just talking myself more and more into introducing one more cache variable...

tests/CMakeLists.txt
krasznaa: Consider a directory layout like:

```
parent_project
|- vecmem
|- project_providing_googletest
```

In this setup the `GTest::gtest` and `GTest::gtest_main` targets will only be defined in `project_providing_googletest`, and are not yet visible in vecmem. So we can't even check with `if( TARGET "GTest::gtest" )` for this.

No, this logic that I wrote down does not work for all situations after all... :frowning:

paulgessinger: And

```
parent_project
|- project_providing_googletest
|- vecmem
``` 

doesn't work either because of the directory scoping?

krasznaa: No, that could potentially work. Since if `project_providing_googletest` builds GoogleTest, it has to create "global" targets for it.

But having to carefully manage include orders like this does not sound like fun. And in our code we would still need to use `if( TARGET "GTest::gtest" )`. Which is also pretty fragile.

BTW this is the first time that I saw discussions showing up in VSCode. :stuck_out_tongue:

Screenshot 2021-09-30 10:52:02

krasznaa:
I guess we concluded on this Paul, right?

paulgessinger:
Yes, let's go.

paulgessinger (approved):

PR #117, opened 2021-09-30
Explicit Symbol Export, main branch (2021.09.30.)

krasznaa:
As we discussed with @stephenswat on a few occasions, we were both interested in making the project's main libraries explicitly declare which symbols they want to export for the clients. From my side I'm a little interested in it for my continued play-sessions on Windows, but this can be a healthy thing on other platforms as well.

The PR first of all makes use of [GenerateExportHeader](https://cmake.org/cmake/help/latest/module/GenerateExportHeader.html) to generate header files with the macros necessary to nicely declare which classes/functions provide the public interface of the libraries. Then I went ahead and set the `CXX_VISIBILITY_PRESET`, `CUDA_VISIBILITY_PRESET`, `HIP_VISIBILITY_PRESET` and `SYCL_VISIBILITY_PRESET` target properties to just the main libraries of the project. (Note that `CUDA_VISIBILITY_PRESET` and `HIP_VISIBILITY_PRESET` don't actually play any role, since `vecmem_cuda` and `vecmem_hip` don't contain any CUDA/HIP code at the moment. But should that change, the target property will already be there. :wink:) And finally I updated the code of the libraries to explicitly declare every class/function that is meant to provide the public interface of the libraries.

For SYCL I had to improve our CMake code yet a little further to make this work. For HIP I didn't bother, since at one point we'll completely retire the HIP CMake code from the repository anyway. (Now that CMake 3.21 comes with a first-class support for HIP itself. See #112.)

The updates did remind me about some of the shenanigans that we do in the unit tests. As the tests make use of some of the private headers of the libraries as well. Since using those private headers does make sense for the tests, but I didn't want to make the symbols in question visible on the libraries, I now added the appropriate source files to the test executables as well.

### DIFF
--- a/cmake/sycl/CMakeSYCLCompiler.cmake.in
+++ b/cmake/sycl/CMakeSYCLCompiler.cmake.in
@@ -27,5 +27,9 @@ set( CMAKE_SYCL_FLAGS_RELEASE_INIT "@CMAKE_SYCL_FLAGS_RELEASE_INIT@" )
 set( CMAKE_SYCL_FLAGS_RELWITHDEBINFO_INIT
    "@CMAKE_SYCL_FLAGS_RELWITHDEBINFO_INIT@" )
 
+# Save the C++ compiler version. This is needed for Platform/Windows-MSVC to
+# work correctly. (As that one knows nothing about SYCL.)
+set( CMAKE_CXX_COMPILER_VERSION "@CMAKE_CXX_COMPILER_VERSION@" )
+
 # Mark the SYCL compiler as "loaded".
 set( CMAKE_SYCL_COMPILER_LOADED TRUE )
--- a/cmake/sycl/CMakeSYCLInformation.cmake
+++ b/cmake/sycl/CMakeSYCLInformation.cmake
@@ -22,6 +22,9 @@ endif()
 # Common CMake include(s).
 include( CMakeCommonLanguageInclude )
 
+# Set up platform specific flags.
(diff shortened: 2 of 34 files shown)


krasznaa:
I started to play a bit with turning the libraries into DLLs on Windows. I believe (with this foundation in place) it will be doable, but it will require some additional amount of work. I'll probably play with that during the weekend. Fixing the things that MSVC complains about will, I believe, make the code more robust...

paulgessinger:
Can you remind me why we need dynamic libraries on windows?

krasznaa:
> Can you remind me why we need dynamic libraries on windows?

:confused: For the same reason as anywhere else...? We don't **need** it. But trying to build DLLs did point out weaknesses in some of our classes. (Where it would be possible to hide more of their implementation behind the scenes, instead of making it all public.) So at this point I view it more as a debugging tool for our code.

krasznaa:
I'm not hearing a lot of objections. So I would be keen on getting this in... :wink:

paulgessinger:
If you say we need it, let's do it. I remain unconvinced, but won't try to block this.

krasznaa:
Abandoning it will be a lot easier than it is to set it up. So if it doesn't work out, we can always just remove the `<LANG>_VISIBILITY_PRESET` settings, and stop the DLL building on Windows. I do hope that the code will become better as a result of all of this... :crossed_fingers:

PR #120, opened 2021-10-06
"Private" CMake 3.21+ & CMake 4.3+ Support, main branch (2021.10.06.)

krasznaa:
Undoing #112, I now switched back to always using our own "HIP CMake code" in the build after all. This was necessary because the HIP support introduced in CMake 3.21 does not seem to be able to build NVidia binaries. Or at least I was not able to find how to force it to do so.

For some reason `hipcc` with ROCm/HIP 4.3 does not add the HIP include directory to the build commands out of the box anymore. So I now add it in the build of `.hip` files by hand all the time.

I also had to add an explicit setting of `CMAKE_HIP_ARCHITECTURE`. The `HIP_ARCHITECTURE` target property is something that CMake 3.21+ expects to have on all "HIP targets". (Since CMake 3.21+ now "knows about" the HIP language at the C\+\+ level.) So even when we set up these targets with our own code, that target property still has to be defined. Luckily setting it to `FALSE` is fine. (In this case CMake just doesn't try to use it.) It's interesting to note that this code is still so new that there's literally zero documentation about this anywhere on the web. I had to look at CMake's C\+\+ code to find what to do about this `HIP_ARCHITECTURE` property. :frowning:


krasznaa:
Let me fast-track this, as it has been blocking me from building the project using the latest version of [gitlab-registry.cern.ch/akraszna/atlas-gpu-devel-env](https://gitlab.cern.ch/akraszna/atlas-gpu-devel-env/container_registry)...

PR #99, opened 2021-09-07
arena memory resource

Yhatoh:
I made the port of the arena memory resource to vecmem,

I created an interface called: arena_interface_memory_resource that will help us to manage the allocations and deallocations without using directly the format asked by the arena_memory resource.

details is a file that get all classes that will help to manage the arena_memory_resource. 

Cuda stream view is implemented to know is which stream we are during the code, so i had to implement a class to make this code having not a dependency of cuda library

The testing part, i tried to added, but didn't work for some reason, i will add it in a future commit.

PR #121, opened 2021-10-06
Allocator Test Fixes, main branch (2021.10.06.)

krasznaa:
While looking at the issues in #99 with [valgrind](https://www.valgrind.org/), I had to realise that our existing core tests were leaking memory. :frowning: This is meant to fix that.

At the same time I updated the tests in `test_core_allocator.cpp`  to use GTest's (non-)equality checking macros, instead of `EXPECT_TRUE(...)` / `ASSERT_TRUE(...)`. (Just because those give a nicer output upon failure.)

### DIFF
--- a/tests/core/test_core_allocator.cpp
+++ b/tests/core/test_core_allocator.cpp
@@ -31,75 +31,78 @@ public:
 class core_allocator_test : public testing::Test {
 protected:
     vecmem::host_memory_resource m_upstream;
-    vecmem::allocator* m_alloc;
-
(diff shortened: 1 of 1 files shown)


stephenswat (approved):
Hmm, so the memory leak was that our tests were not properly deallocating allocator _itself_? That makes sense, as confusing as it is on the surface. Looks good to me. :+1:

krasznaa:
It was 2 things. :stuck_out_tongue:
  - The `vecmem::allocator` object was not deleted;
  - The objects created in the `core_allocator_test_constructor` test were not deleted.

PR #122, opened 2021-10-06
CI (and SYCL) Updates, main branch (2021.10.06.)

krasznaa:
After some extended discussions with @stephenswat today about GitHub Actions, this is what I came up with, what a streamlined CI configuration could look like for the project.

First off, as I was aghast to find out, the SYCL tests were actually failing since a while. The CMake version used in `ghcr.io/acts-project/ubuntu2004_oneapi:v12` was too old for the recently updated CMake SYCL code to work with. Which resulted in the SYCL builds silently only building the core library and tests. :frowning: In this new configuration the "GPU builds" are forced to turn on the build of their respective libraries/tests. So similar issues should be caught by the system in the future.

To overcome the "environment setup conundrum" that we discussed with @stephenswat, I ended up introducing a helper script into the `.github` directory. This script is meant to set up "platform specific" flags for the container based builds. For now it only does anything for the SYCL builds, but later on we may want to add specific stuff for CUDA/HIP as well.

The main feature of this setup from the outside is that the number of tests was halved. In order to reduce the load on the GitHub CI system, "Release" builds are now only made with "debug message level" 0, and "Debug" builds are only set up with "debug message level" 5. This should still give a reasonable coverage for the CI, while halving the number of build jobs.

Opinions?

### DIFF
--- a/.github/ci_setup.sh
+++ b/.github/ci_setup.sh
@@ -1,0 +1,22 @@
+# VecMem project, part of the ACTS project (R&D line)
+#
+# (c) 2021 CERN for the benefit of the ACTS project
+#
+# Mozilla Public License Version 2.0
+#
+# This script is meant to configure the build/runtime environment of the
+# Docker contaners that are used in the project's CI configuration.
+#
+# Usage: source .github/ci_setup.sh <platform name>
+#
+
+# The platform name.
+PLATFORM_NAME=$1
+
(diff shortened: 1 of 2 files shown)


krasznaa:
Let's start with this one. As this very crucially restores our ability to spot SYCL code errors in the CI...

PR #125, opened 2021-10-08
Implement an instrumenting memory resource

stephenswat:
This memory resource is a little different from the others, in the sense that its purpose is not to control how memory is allocated, but rather to monitor and instrument that process. This memory resource is mainly designed for debugging, profiling, and testing other resources, both in our code and in user code.

The instrumenting memory resource provides support for arbitrary user-provided pre-allocation, post-allocation, and pre-deallocation hooks. It keeps track of a chronological series of allocation and deallocation requests, and it knows about the side of its own outstanding allocation pool.

### DIFF
--- a/core/include/vecmem/memory/instrumenting_memory_resource.hpp
+++ b/core/include/vecmem/memory/instrumenting_memory_resource.hpp
@@ -1,0 +1,168 @@
+/**
+ * VecMem project, part of the ACTS project (R&D line)
+ *
+ * (c) 2021 CERN for the benefit of the ACTS project
+ *
+ * Mozilla Public License Version 2.0
+ */
+
+#pragma once
+
+#include <cstddef>
+#include <functional>
+#include <vector>
+
+#include "vecmem/memory/details/memory_resource_base.hpp"
+#include "vecmem/memory/memory_resource.hpp"
+#include "vecmem/vecmem_core_export.hpp"
+
+#ifdef _MSC_VER
+#pragma warning(push)
+#pragma warning(disable : 4251)
+#endif
+
+namespace vecmem {
+/**
+ * @brief This memory resource forwards allocation and deallocation requests to
+ * the upstream resource while recording useful statistics and information
+ * about these events.
+ *
+ * This allocator is here to allow us to debug, to profile, to test, but also
+ * to instrument user code.
+ */
+class VECMEM_CORE_EXPORT instrumenting_memory_resource final
+    : public details::memory_resource_base {
+public:
+    /**
+     * @brief Structure describing a memory resource event.
+     */
+    struct VECMEM_CORE_EXPORT memory_event {
+        /**
+         * @brief Classify an event as an alloction or a deallocation.
+         */
+        enum class type { ALLOCATION, DEALLOCATION };
+
+        /**
+         * @brief Construct an allocation/deallocation event.
+         *
+         * @param[in] t The type of event (allocation or deallocation).
+         * @param[in] s The size of the request.
+         * @param[in] a The alignment of the request.
+         * @param[in] p The pointer that was returned or deallocated.
+         * @param[in] ns The time taken to perform the request in nanoseconds.
+         */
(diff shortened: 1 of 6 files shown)


tests/core/test_core_memory_resources.cpp
```diff
@@ -165,13 +168,15 @@ INSTANTIATE_TEST_SUITE_P(core_memory_resource_tests, core_memory_resource_test,
...
+                               "instrumenting_resource"}}));
+
+INSTANTIATE_TEST_SUITE_P(
+    core_memory_resource_stress_tests, core_memory_resource_stress_test,
+    testing::Values(&host_resource, &binary_resource, &arena_resource),
+    vecmem::testing::memory_resource_name_gen(
+        {{&host_resource, "host_resource"},
+         {&binary_resource, "binary_resource"},
+         {&arena_resource, "arena_resource"},
+         {&instrumenting_resource, "instrumenting_resource"}}));
```
krasznaa, line 182: I made this same mistake in the past myself as well. :stuck_out_tongue:

You need to list the new memory resource not just to `vecmem::testing::memory_resource_name_gen`, but more importantly to the `testing::Values` object. In the current setup the test would know how to name the resource, but it is not set up to actually use it. :smile:

core/src/memory/instrumenting_memory_resource.cpp
```diff
@@ -0,0 +1,173 @@
...
+bool instrumenting_memory_resource::do_is_equal(
+    const memory_resource &other) const noexcept {
+    /*
+     * An instrumenting memory resource is only ever equal to itself.
+     * Technically this is not true, because if the upstreams are the same then
+     * they are equal. However, this cannot be allowed to happen because it
+     * would break our administration
+     */
+    return this == &other;
+}
```
krasznaa, line 172: You could leave this one out. `vecmem::details::memory_resource_base` implement this function in the same way.

core/include/vecmem/memory/instrumenting_memory_resource.hpp
```diff
@@ -0,0 +1,155 @@
...
+        std::function<void(void*, std::size_t, std::size_t)> f);
+
+private:
+    virtual void* do_allocate(std::size_t, std::size_t) override;
+
+    virtual void do_deallocate(void* p, std::size_t, std::size_t) override;
+
+    virtual bool do_is_equal(const memory_resource&) const noexcept override;
+
+    memory_resource& m_upstream;
```
krasznaa, line 124: You documented all the other member variables. Why not this one? :stuck_out_tongue:

core/include/vecmem/memory/instrumenting_memory_resource.hpp
```diff
@@ -0,0 +1,155 @@
...
+ * Mozilla Public License Version 2.0
+ */
+
+#pragma once
+
+#include <cstddef>
+#include <functional>
+
+#include "vecmem/memory/details/memory_resource_base.hpp"
+#include "vecmem/memory/memory_resource.hpp"
```
krasznaa, line 15: I think it would be healthy to add an explicit include for the auto-generated `vecmem/vecmem_core_export.hpp`. `memory_resource_base.hpp` pulls it in, but we shouldn't rely on that.

core/src/memory/instrumenting_memory_resource.cpp
stephenswat: Fixed. :smile:

core/include/vecmem/memory/instrumenting_memory_resource.hpp
stephenswat: Added!

krasznaa:
Missing `<vector>` include. :stuck_out_tongue: The CI is working overtime...

krasznaa:
> I can't emulate Windows builds in my local CI environment. upside_down_face

The last one was coming from `libc++`. That, you technically could. :smiling_imp:

PR #131, opened 2021-10-11
Memory Monitor, main branch (2021.10.11.)

krasznaa:
With @konradkusiak97 making performance measurements using the [Acts](https://github.com/acts-project/acts) code at the moment, I was wondering how to make the best use of the recently introduced `vecmem::instrumenting_memory_resource` class.

I considered multiple things.
  - Extending `vecmem::instrumenting_memory_resource` to collect some other information beside "current outstanding memory";
  - Adding a couple of helper functions that would take an `std::vector<vecmem::instrumenting_memory_resource::memory_event>` argument, and calculate some useful metrics out of those event objects.

But after some pondering, I rather came up with this... That I would write a class that could listen to events from `vecmem::instrumenting_memory_resource`, and collect some useful metrics with that information. The reason I liked @stephenswat's `vecmem::instrumenting_memory_resource` design was exactly that it made it possible for users to write such classes in their own code. But I thought it could still be useful to provide a basic implementation as part of this project as well.

At the same time, the `vecmem::instrumenting_memory_resource::m_outstanding` variable has bugged me ever since #125. Since it just stands on its own in that class, sort of out of place. So with this update I removed that variable, as `vecmem::memory_monitor` is meant to take over its functionality. This way the `vecmem::instrumenting_memory_resource` and `vecmem::memory_monitor` classes each have their well defined purpose in my mind...

We may still want to add some helper functions that would "analyse" sets of `vecmem::instrumenting_memory_resource::memory_event` objects for some common metrics, but I didn't want to bloat this PR with something like that as well.

What do you think @stephenswat?

### DIFF
--- a/core/CMakeLists.txt
+++ b/core/CMakeLists.txt
@@ -75,6 +75,8 @@ vecmem_add_library( vecmem_core core
    "include/vecmem/utils/impl/copy.ipp"
    "src/utils/copy.cpp"
    "include/vecmem/utils/debug.hpp"
+   "src/utils/memory_monitor.cpp"
+   "include/vecmem/utils/memory_monitor.hpp"
    "include/vecmem/utils/type_traits.hpp"
    "include/vecmem/utils/types.hpp" )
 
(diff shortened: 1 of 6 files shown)


paulgessinger:
This looks pretty clever to me.

PR #136, opened 2021-10-30
Memory Resource Common Test Re-organisation, main branch (2021.10.30.)

krasznaa:
This is a follow-up to #135. I was considering asking @stephenswat for these modifications as part of that PR, but then decided that I would just do it myself. :stuck_out_tongue:

I made the common tests part of the `vecmem_testing_common` static library, moving most of their code into `.cpp` files.

At the same time I also removed references to CUDA from the documentation of the common tests. (There were still hints in there of where that code came from. :smile:)

I also changed the names of the files to reflect the names of the test classes that they hold. (As per the ATLAS coding standard. :wink:)


tests/common/memory_resource_test_basic.cpp
```diff
@@ -28,5 +18,3 @@ TEST_P(memory_resource_test_basic, allocations) {
         resource->deallocate(ptr, size);
     }
 }
-
-GTEST_ALLOW_UNINSTANTIATED_PARAMETERIZED_TEST(memory_resource_test_basic);
```
krasznaa, line 32: What was the situation with these `GTEST_ALLOW_UNINSTANTIATED_PARAMETERIZED_TEST` calls by the way? I don't think we need them for this code... :confused:

stephenswat (approved):
Sure, works for me. :stuck_out_tongue:

tests/common/memory_resource_test_basic.cpp
stephenswat: If you don't include this and you manage to somehow incorporate a test suite in your test binary, Google Test will throw an error. Whether that is desirable is debatable, but I decided to ignore that error for these re-usable tests, because the risk of accidental inclusion or linkage is greater.

krasznaa: I see. At first I thought I'd need this just to be able to define these test classes as part of the common static library. But as long as all of the client executables make use of this test, apparently that is not needed.

In this case I would leave these statements removed. The common test code is only meant to be used by this repository's tests. So we shouldn't worry about clients possibly not using this code correctly. And out of the gate any common test should be there so that it would be used by all "device library" tests.

If we need to, we can add such things back later on. But we should have a good reason for doing that first, before preemptively adding this.

PR #137, opened 2021-10-31
Windows Warning Fixes, main branch (2021.10.31.)

krasznaa:
Fixed warnings in the CUDA and SYCL tests, showing up on Windows. In all cases the warnings being about a reduction of "precision" on a given variable.

Unfortunately the CI will remain blind to these issues, as we won't be realistically testing the CUDA and SYCL code on Windows in it. 😦

At the same time, the CUDA warning was fairly weird. It only showed up when using [MSBuild](https://docs.microsoft.com/en-us/visualstudio/msbuild/msbuild) for the build. With [NMake](https://docs.microsoft.com/en-us/cpp/build/reference/nmake-reference) it did **not** show up. :confused:


stephenswat (approved):
Nothing controversial here.

PR #143, opened 2021-11-04
Source File Grouping, main branch (2021.11.04.)

krasznaa:
Introduced grouping for the source files, so that they would show up a little bit more nicely in [Xcode](https://developer.apple.com/xcode/) and [Visual Studio](https://visualstudio.microsoft.com/).

Since I myself practically only use [Visual Studio Code](https://code.visualstudio.com/) for my code editing these days, on which this setting has absolutely no effect, this was not an urgent update in this project. But since I've been talking about this as a possibility for such a long time already, most recently in https://github.com/acts-project/detray/pull/145, I really needed to demonstrate what I'm talking about exactly.

The change with this is that we go from the project showing up like this in Xcode:

Screen Shot 2021-11-04 at 2 40 16 PM

To it shoring up like this instead:

Screen Shot 2021-11-04 at 2 40 50 PM

It's not a huge improvement, but an improvement non the less... 😛

### DIFF
--- a/cmake/vecmem-functions.cmake
+++ b/cmake/vecmem-functions.cmake
@@ -26,6 +26,9 @@ function( vecmem_add_library fullname basename )
    # Parse the function's options.
    cmake_parse_arguments( ARG "" "TYPE" "" ${ARGN} )
 
+   # Group the source files.
+   vecmem_group_source_files( ${ARG_UNPARSED_ARGUMENTS} )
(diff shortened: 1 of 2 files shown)


stephenswat (approved):
I have absolutely no idea what is happening here or why, but I'll take your word for it. :stuck_out_tongue:

PR #145, opened 2021-11-14
SYCL Test Fixes, main branch (2021.11.14.)

krasznaa:
These are multiple fixes for the SYCL jagged vector tests.

For quite a while now these tests were only working with the [Level-0](https://dgpu-docs.intel.com/technologies/level-zero.html) driver for my Intel GPUs. The [OpenCL](https://www.intel.com/content/www/us/en/developer/articles/tool/opencl-drivers.html) drivers were not willing to run the tests successfully.

Until now I blamed this all on bugs in the OpenCL driver. But turns out, I was wrong... :frowning: All the failures were coming from 2 types of errors:
  - The host code not consistently waiting for every kernel to finish;
  - "High level" objects not being consistently constructed on top of the same queues.

The first types of errors, I believe, didn't show up when using the Level-0 driver, simply because the kernels run faster with that driver. But it's not impossible that the Level-0 driver has some inherent wait behaviour, while the OpenCL one (correctly) doesn't.

The second types of errors are a bit more understandable. With OpenCL two SYCL queues created on top of the same device are not equivalent to each other. Memory allocated through one may not be handled by the other. Without explicitly connecting the memory resource and `vecmem::sycl::copy` objects to the correct queue objects, these would internally create their own queue objects. Those queues would operate on the same device, but would not be able to collaborate with memory/kernels created through other queue instances.

With these in place, and all Intel drivers updated to their very latest versions, things sort of work... I'm still observing some weirdness with the `sycl_containers_test.atomic_memory` test, on the CPU OpenCL driver. That particular issue, according to https://github.com/intel/llvm/issues/3549, should've been fixed by now. Yet, the only setup in which it works successfully for me is on Windows (11), in a Debug build. :confused: On Linux, even with the latest CPU driver that I downloaded this last week, it still has issues. And even on Windows, when in Release mode, that particular test still crashes. (Not just fails, it crashes.)

So things are better now, but there are still some rough edges here and there...

### DIFF
--- a/tests/sycl/test_sycl_jagged_containers.sycl
+++ b/tests/sycl/test_sycl_jagged_containers.sycl
@@ -186,22 +186,25 @@ TEST_F(sycl_jagged_containers_test, mutate_in_kernel) {
     auto vec_data = vecmem::get_data(m_vec);
 
     // Run the linear transformation.
-    m_queue.submit([&const_data, &vec_data](cl::sycl::handler& h) {
-        // Create the kernel functor.
-        LinearTransformKernel kernel(const_data, vec_data, vec_data);
-        // Execute this kernel.
-        h.parallel_for<LinearTransformKernel>(
-            cl::sycl::range<1>(vec_data.m_size), kernel);
-    });
+    m_queue
+        .submit([&const_data, &vec_data](cl::sycl::handler& h) {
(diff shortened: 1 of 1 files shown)


krasznaa:
I'll assume silence means consent...

PR #146, opened 2021-11-19
Windows Support Update, main branch (2021.11.19.)

krasznaa:
As I discovered in https://github.com/acts-project/algebra-plugins/pull/42, using VecMem on Windows from another project is not as trivial as I thought it was. :frowning:

Now I made MSVC always set `__cplusplus` when using `vecmem::core`. Removing the setting meant just for the build of this project. Setting the appropriate flags through generator expressions, to make the code as robust as possible.

I also enabled the appropriate flag for setting `__cplusplus` with NVCC on windows. Since in [algebra-plugins](https://github.com/acts-project/algebra-plugins) we also need that.

Finally, suppressed the warning about exporting a class that inherits from a non-exported class, when using the NVCC compiler on Windows. Unfortunately I couldn't find any good documentation on when CUDA stopped using `#pragma diag_suppress` and switched to `#pragma nv_diag_suppress`. So the CI should tell us if this was before CUDA 10.2 or not... :thinking:

To make sure that all of this was working correctly, I made a copy of the CUDA memory resource test, and made NVCC compile it as well.

### DIFF
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -68,6 +68,8 @@ jobs:
         PLATFORM:
           - NAME: "CUDA"
             CONTAINER: "ghcr.io/acts-project/ubuntu1804_cuda:v11"
+          - NAME: "CUDA"
+            CONTAINER: "ghcr.io/acts-project/ubuntu2004_cuda:v13"
           - NAME: "HIP"
             CONTAINER: "ghcr.io/acts-project/ubuntu1804_rocm:v11"
           - NAME: "SYCL"
--- a/core/CMakeLists.txt
+++ b/core/CMakeLists.txt
@@ -103,6 +103,13 @@ target_compile_definitions( vecmem_core PUBLIC
    $<BUILD_INTERFACE:VECMEM_DEBUG_MSG_LVL=${VECMEM_DEBUG_MSG_LVL}>
    $<BUILD_INTERFACE:VECMEM_SOURCE_DIR_LENGTH=${VECMEM_SOURCE_DIR_LENGTH}> )
 
+# The library headers make checks on the value of the __cplusplus macro. So we
+# need to make sure that MSVC would always set that macro up when using this
+# library.
+target_compile_options( vecmem_core PUBLIC
+   $<$<AND:$<CXX_COMPILER_ID:MSVC>,$<COMPILE_LANGUAGE:CXX>>:/Zc:__cplusplus>
+   $<$<AND:$<CXX_COMPILER_ID:MSVC>,$<COMPILE_LANGUAGE:CUDA>>:-Xcompiler /Zc:__cplusplus> )
+
 # Figure out where to get <memory_resource> from.
 include( CheckCXXSourceCompiles )
 check_cxx_source_compiles( "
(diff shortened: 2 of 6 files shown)


krasznaa:
So... CUDA 10.2 is fine with the `nv_diag_suppress` pragmas, but I had to only conditionally set up the build of the C\+\+17 CUDA test. At the same time I added a build with CUDA 11.3, to actively test this C\+\+17 feature of the code.

.github/workflows/builds.yml
```diff
@@ -68,6 +68,8 @@ jobs:
         PLATFORM:
           - NAME: "CUDA"
             CONTAINER: "ghcr.io/acts-project/ubuntu1804_cuda:v11"
+          - NAME: "CUDA"
+            CONTAINER: "ghcr.io/acts-project/ubuntu2004_cuda:v13"
```
stephenswat, line 72: Would it maybe be worth it just replacing the previous CUDA platform? It might be confusing to have two CUDA platforms, and also it takes up more CI time.

core/CMakeLists.txt
```diff
@@ -103,6 +103,13 @@ target_compile_definitions( vecmem_core PUBLIC
    $<BUILD_INTERFACE:VECMEM_DEBUG_MSG_LVL=${VECMEM_DEBUG_MSG_LVL}>
    $<BUILD_INTERFACE:VECMEM_SOURCE_DIR_LENGTH=${VECMEM_SOURCE_DIR_LENGTH}> )
 
+# The library headers make checks on the value of the __cplusplus macro. So we
+# need to make sure that MSVC would always set that macro up when using this
+# library.
+target_compile_options( vecmem_core PUBLIC
+   $<$<AND:$<CXX_COMPILER_ID:MSVC>,$<COMPILE_LANGUAGE:CXX>>:/Zc:__cplusplus>
+   $<$<AND:$<CXX_COMPILER_ID:MSVC>,$<COMPILE_LANGUAGE:CUDA>>:-Xcompiler /Zc:__cplusplus> )
+
```
stephenswat, line 112: I'm a little confused here, I was under the impression that `__cplusplus` was a C++-standard definition that should _always_ be there if compiling C++. Is this not the case? :fearful: Does MVCC not adhere to this standard?

stephenswat (approved):

.github/workflows/builds.yml
krasznaa: :thinking: I'd prefer to keep CUDA 10.2 compatibility alive for a little longer. With the amount of CI jobs that we run these days, I think we should be able to afford this...

stephenswat: Sure, no problem. What do you think of naming one "CUDA 10.2" and the other "CUDA 11.3"? Might make it easier to see which of the CI jobs breaks.

core/CMakeLists.txt
krasznaa: This is a messy thing that you're shining light on with this question... :stuck_out_tongue:

https://docs.microsoft.com/en-us/cpp/build/reference/zc-cplusplus

Some of the Windows headers don't like it when `__cplusplus` is set to a "too high value". :confused: So by default MSVC sets it to `199711L`. Whatever the actual C\+\+ standard is that it really uses.

So yeah, this is something quite unfortunate. But since we depend on the "correct" value of `__cplusplus` in our public header, we have to make sure in this project's code that MSVC would always set that macro value correctly...

stephenswat: My god that is horrendously stupid (on MSVC's end). But yeah, I guess we need to fix it in this way then.

How on earth does anyone use MSVC to compile anything in production??

.github/workflows/builds.yml
stephenswat: Nevermind, looking at it that would break some of the CI logic. Let's just go with this.

PR #147, opened 2021-11-29
Asynchronous CUDA Copies, main branch (2021.11.29.)

krasznaa:
After the comments from @fwyzard in the morning, I decided to look at asynchronous CUDA copies once again. :smile: This is what I came up with...

In the same way in which the code hides `sycl::queue` from the public interface of `vecmem::sycl`, I hid `cudaStream_t` from the public interface of `vecmem::cuda`. Through the introduction of a `vecmem::cuda::stream_wrapper`, and an underlying `vecmem::cuda::details::opaque_stream` type.

With those in place I added a `vecmem::cuda::async_copy` type. I behaves exactly like `vecmem::cuda::copy`, besides doing everything asynchronously.

Finally I added a trivial test, just to see that things don't break in a very obvious way.

### DIFF
--- a/cuda/src/utils/cuda_wrappers.cpp
+++ b/cuda/src/utils/cuda_wrappers.cpp
@@ -14,6 +14,9 @@
 // CUDA include(s).
 #include <cuda_runtime_api.h>
 
+// System include(s).
+#include <cassert>
+
 namespace vecmem::cuda::details {
 
 int get_device() {
@@ -21,6 +24,12 @@ int get_device() {
     int d = 0;
     VECMEM_CUDA_ERROR_IGNORE(cudaGetDevice(&d));
     return d;
+}
+
+cudaStream_t get_stream(const stream_wrapper& stream) {
+
+    assert(stream.stream() != nullptr);
+    return reinterpret_cast<cudaStream_t>(stream.stream());
 }
 
 }  // namespace vecmem::cuda::details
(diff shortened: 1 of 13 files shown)


krasznaa:
:smile: I'm still pondering what to do for the memory resources... Since asynchronicity doesn't fit well with how [std::pmr::memory_resource](https://en.cppreference.com/w/cpp/memory/memory_resource) is designed.

It's good to hear though that, as I hoped, caching would be more performant than the asynchronous allocations/deallocations. Since this project is **really** focused on that. And not on clever ways in which asynchronicity could be expressed in an object-oriented way...

cuda/src/utils/cuda_wrappers.cpp
```diff
@@ -23,4 +26,10 @@ int get_device() {
     return d;
 }
 
+cudaStream_t get_stream(const stream_wrapper& stream) {
+
+    assert(stream.stream() != nullptr);
+    return reinterpret_cast<cudaStream_t>(stream.stream());
```
stephenswat, line 32: Is this safe? Surely these types are not layout-compatible? :fearful:

stephenswat (approved):
Right now, I don't fully see how exactly we use this, but I think this is at least a good start to start moving towards async memory transfers. Let's go ahead with this, and then we can iterate on it later.

cuda/src/utils/cuda_wrappers.cpp
krasznaa: I think it's "reasonably safe". Remember, `cudaStream_t` is a pointer type.

```
[bash][Legolas]:~ > grep -r "typedef.*cudaStream_t" software/cuda/11.4.3/x86_64-ubuntu2004/include/driver_types.h 
typedef __device_builtin__ struct CUstream_st *cudaStream_t;
[bash][Legolas]:~ >
```

Casting that back and forth to `void*` is not the most elegant code, but it's pretty safe.

PR #148, opened 2021-12-01
Add a simple suite of memory benchmarks

stephenswat:
You may have noticed the following graph in our vecmem talk at ACAT'21:

Screenshot from 2021-12-01 14-13-27

This graph was generated using vecmem (the data that is, not the image itself), but I had to implement some benchmarking code to do it. This pull request contains that benchmarking code. Now, please note that this code is somewhere between "throw-away one-time code" and "actual high-quality code", but I think it would be a shame to throw it away. For now, I am opening a draft merge request for it, and we can iterate on this to see what we want to do to get this into the actual vecmem code base.

Pinging @krasznaa, as part of this involves CMake which I have tried as best I can to integrate with the rest of the build system, but I am certain I missed a few things... :smile:


stephenswat:
Agreed, let's go for it. Thanks for the edits! :smile:

PR #149, opened 2021-12-03
Implement a custom main function for CUDA tests

stephenswat:
One of the goals of vecmem's tests is to make sure that there are no memory leaks. For CPU code, we can test this by using valgrind, and for CUDA we can test the same thing using `cuda-memcheck`. However, the CUDA memchecker has a peculiar feature where the leak checker (`--leak-check full`) only works if you make a call to `cudaDeviceReset` at the end of your program. We currently rely on our tests being linked against the GTest standard main function, which for obvious reasons does not include this test. In this commit, I add a new CUDA-specific Google Test main function which we can use instead, which makes the correct call to the CUDA device reset function.

### DIFF
--- a/tests/cuda/CMakeLists.txt
+++ b/tests/cuda/CMakeLists.txt
@@ -14,6 +14,22 @@ include( vecmem-compiler-options-cuda )
 # External dependency/dependencies.
 find_package( CUDAToolkit REQUIRED )
 
+add_library(
+   vecmem_testing_cuda_main
+   STATIC
+   test_cuda_main.cpp
+   "${CMAKE_CURRENT_SOURCE_DIR}/../../cuda/src/utils/cuda_error_handling.hpp"
+   "${CMAKE_CURRENT_SOURCE_DIR}/../../cuda/src/utils/cuda_error_handling.cpp"
+)
+
+target_link_libraries(
+   vecmem_testing_cuda_main
+   PUBLIC
+   CUDA::cudart
+   GTest::gtest
+   vecmem::core
+)
+
 # Test all of the CUDA library's features.
 vecmem_add_test( cuda
    "test_cuda_memory_resources.cpp"
@@ -26,7 +42,7 @@ vecmem_add_test( cuda
    "${CMAKE_CURRENT_SOURCE_DIR}/../../cuda/src/utils/cuda_error_handling.cpp"
    "${CMAKE_CURRENT_SOURCE_DIR}/../../cuda/src/utils/cuda_wrappers.hpp"
    "${CMAKE_CURRENT_SOURCE_DIR}/../../cuda/src/utils/cuda_wrappers.cpp"
-   LINK_LIBRARIES CUDA::cudart vecmem::core vecmem::cuda GTest::gtest_main
+   LINK_LIBRARIES CUDA::cudart vecmem::core vecmem::cuda vecmem_testing_cuda_main
                   vecmem_testing_common )
 
 # Set up a separate test that would ensure a C++17 standard. But only with
@@ -34,7 +50,7 @@ vecmem_add_test( cuda
 if( "${CUDAToolkit_VERSION}" VERSION_GREATER_EQUAL "11.0" )
    vecmem_add_test( cuda_cxx17
       "test_cuda_memory_resources.cu"
-      LINK_LIBRARIES vecmem::cuda GTest::gtest_main vecmem_testing_common )
+      LINK_LIBRARIES vecmem::cuda vecmem_testing_cuda_main vecmem_testing_common )
    set_target_properties( vecmem_test_cuda_cxx17 PROPERTIES
       CUDA_STANDARD 17 )
 endif()
--- a/tests/cuda/test_cuda_main.cpp
+++ b/tests/cuda/test_cuda_main.cpp
@@ -1,0 +1,27 @@
+/*
+ * VecMem project, part of the ACTS project (R&D line)
+ *
+ * (c) 2021 CERN for the benefit of the ACTS project
+ *
+ * Mozilla Public License Version 2.0
+ */
+
+#include <gtest/gtest.h>
+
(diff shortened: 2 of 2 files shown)


tests/cuda/CMakeLists.txt
```diff
@@ -14,6 +14,27 @@ include( vecmem-compiler-options-cuda )
 # External dependency/dependencies.
 find_package( CUDAToolkit REQUIRED )
 
+add_library(
+   vecmem_testing_cuda_main
+   SHARED
```
krasznaa, line 19: This should not have worked... :confused: We'll need to add `-Wl,--no-undefined` for GCC and Clang... (I don't think there's a more elegant way of doing it, than to just hardcode it.)

You see, `cuda_error_handling.cpp` should've been made part of this library if you wanted to use a shared library.

But let's take a further step back. I think this should be made into a static library. Having it as a shared library is just a bit over the top in my mind.

Once you made it into a static library, I would make `cuda_error_handling.cpp` part of that static library. And at the same time remove that source file from the executables.

tests/cuda/CMakeLists.txt
```diff
@@ -14,6 +14,27 @@ include( vecmem-compiler-options-cuda )
...
+   test_cuda_main.cu
+)
+
+target_link_libraries(
+   vecmem_testing_cuda_main
+   PUBLIC
+   CUDA::cudart
+   gtest
+   vecmem::core
+   vecmem::cuda
```
krasznaa, line 29: It should not need to link against either of those. The one header file that it uses from `vecmem::cuda` is a private header. So linking against `vecmem::cuda` doesn't even help with that...

tests/cuda/CMakeLists.txt
```diff
@@ -24,15 +45,17 @@ vecmem_add_test( cuda
...
 
 # Set up a separate test that would ensure a C++17 standard. But only with
 # CUDA 11+.
 if( "${CUDAToolkit_VERSION}" VERSION_GREATER_EQUAL "11.0" )
    vecmem_add_test( cuda_cxx17
       "test_cuda_memory_resources.cu"
-      LINK_LIBRARIES vecmem::cuda GTest::gtest_main vecmem_testing_common )
+      "${CMAKE_CURRENT_SOURCE_DIR}/../../cuda/src/utils/cuda_error_handling.hpp"
+      "${CMAKE_CURRENT_SOURCE_DIR}/../../cuda/src/utils/cuda_error_handling.cpp"
+      LINK_LIBRARIES CUDA::cudart vecmem::core vecmem::cuda vecmem::testing_cuda_main vecmem_testing_common )
```
krasznaa, line 58: Why `vecmem::core`? We must only explicitly link against libraries whose public headers we use. It is up to `vecmem::cuda` to ensure that we can use all of its public headers by just linking against that one library. (This is why it publicly depends on `vecmem::core`.)

tests/cuda/test_cuda_main.cu
```diff
@@ -0,0 +1,21 @@
...
+#include <gtest/gtest.h>
+
+#include "../../cuda/src/utils/cuda_error_handling.hpp"
+
+int main(int argc, char** argv) {
+    testing::InitGoogleTest(&argc, argv);
+
+    int r = RUN_ALL_TESTS();
+
+    VECMEM_CUDA_ERROR_CHECK(cudaDeviceReset());
```
krasznaa, line 18: As you saw from the CI, this will have to be made a bit more elaborate... In the past I did things like:

```c++
int nDevices = 0;
cudaGetDeviceCount(&nDevices);
if(nDevices > 0) {
   VECMEM_CUDA_ERROR_CHECK(cudaDeviceReset());
}
```

Notice that the return value of `cudaGetDeviceCount(...)` is not checked. Since running that function on a system without the NVidia driver loaded will result in an error code return. It will however silently leave `nDevices` at `0` in that case.

Also: this file should be called `test_cuda_main.cpp`. Let's only use `.cu` for the files that can only be processed by `nvcc`. (I.e. mostly just files with kernel declarations and launches.)

tests/cuda/CMakeLists.txt
```diff
@@ -14,6 +14,27 @@ include( vecmem-compiler-options-cuda )
...
+   vecmem_testing_cuda_main
+   SHARED
+   test_cuda_main.cu
+)
+
+target_link_libraries(
+   vecmem_testing_cuda_main
+   PUBLIC
+   CUDA::cudart
+   gtest
```
krasznaa, line 27: This should be `GTest::gtest`. (`gtest` only works if GoogleTest is built as part of VecMem. When taking GoogleTest from "the system", it doesn't.)

tests/cuda/CMakeLists.txt
```diff
@@ -14,6 +14,27 @@ include( vecmem-compiler-options-cuda )
...
+   vecmem::core
+   vecmem::cuda
+)
+
+add_library(
+   vecmem::testing_cuda_main
+   ALIAS
+   vecmem_testing_cuda_main
+)
```
krasznaa, line 36: I know that these aliases are created in the other Acts projects. But I don't see a reason for them existing.

Aliases should be there to provide user friendly names for the "public" libraries of a project. To make it possible to use for instance `vecmem::core` with that name regardless of whether it is picked up pre-compiled, or is being built as part of the current project.

This library should only be used by these tests, and nobody else. So setting up the alias seems to be just a waste of a few lines of code. (This is why `vecmem_testing_common` has the name that it does...)

krasznaa (changes requested):
Just a few comments here and there... :stuck_out_tongue:

stephenswat:
Well, I think I broke my own "comments per line of code" record here, but it should be better now. The only thing I can't fix is the library linking against `vecmem::core`, because the CUDA utils import `types.hpp` from the core library.

krasznaa (approved):
Let's do it...

PR #150, opened 2021-12-08
Shared Library Linking Update, main branch (2021.12.08.)

krasznaa:
Triggered by the updates in #149, I added `-Wl,--no-undefined` for shared library linking with GCC/Clang. To make a bit more sure that the linking setup for our libraries is correct.

At the same time simplified the setup of the basic C\+\+ warning flags for GCC/Clang a little. These need to be set up a bit more awkwardly for HIP and SYCL, because of imperfections in our implementation of those languages in CMake. But for C\+\+ we don't have to be that complicated...


stephenswat (approved):
Looks good.

PR #152, opened 2021-12-09
Static Array Update, main branch (2021.12.09.)

krasznaa:
Triggered by #151, I came up with this.

As I wrote there, `vecmem::static_array` has done things in a slightly funky way so far. But I had to realise that the libstdc\+\+ developers came up with a better design for their array class after all. :stuck_out_tongue: So I now embraced that setup in `vecmem::static_array` as well.

This means that the class would not have any explicit constructors. It is now declared as a "simple" struct with just a single member variable. Which the language can figure out itself how to generate functional constructors for...

One downside of this is that while with the previous implementation it was possible to write things like

```c++
vecmem::static_array<int, 4> a(1, 2, 3, 4);
```

, now "only" the curly-braced formalism works. I.e.:

```c++
vecmem::static_array<int, 4> a{1, 2, 3, 4};
```

But
  - `std::array` behaves the same;
  - this should be a small enough price to pay for making `vecmem::static_array` functional with non-copyable/non-movable elements.

I also had to tweak `vecmem::static_vector` slightly, since it was sharing some code with `vecmem::static_array` until now. But with the new implementation it doesn't do that anymore.

I added a simple test for all device languages, just to make sure that this would work. And it seems to. :smile:

### DIFF
--- a/core/include/vecmem/containers/details/static_array_traits.hpp
+++ b/core/include/vecmem/containers/details/static_array_traits.hpp
@@ -1,0 +1,31 @@
+/** VecMem project, part of the ACTS project (R&D line)
+ *
+ * (c) 2021 CERN for the benefit of the ACTS project
+ *
+ * Mozilla Public License Version 2.0
+ */
+#pragma once
+
+// System include(s).
+#include <cstddef>
+
+namespace vecmem {
+namespace details {
+
+/// Helper type for an array with a given type and size
+///
+/// This is needed to handle zero-sized arrays correctly. As those are not
+/// part of the C++ standard.
+///
+template <typename T, std::size_t size>
+struct static_array_type {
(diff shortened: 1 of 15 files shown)


stephenswat (approved):
Looks good to me.

beomki-yeo:
This is definitely good. detray also uses curly-braced constructor in many places so I was going to file an issue for this.

beomki-yeo (approved):

beomki-yeo:
@krasznaa Could you launch v0.8.0 so I can use the new feature in detray?

krasznaa:
> @krasznaa Could you launch v0.8.0 so I can use the new feature in detray?

Done. https://github.com/acts-project/vecmem/releases/tag/v0.8.0 :wink:

PR #153, opened 2021-12-09
CUDA Windows Fix, main branch (2021.12.09.)

krasznaa:
Disabled the NVCC specific warning about using `std::unique_ptr` in an exported (DLL) class. (`vecmem::cuda::stream_wrapper`)

It was already disabled for MSVC, but I forgot that the class would be seen by NVCC as well...


stephenswat (approved):
Looks perfectly harmless.

PR #155, opened 2021-12-10
Implement resource-aware unique pointer types

stephenswat:
For quite a while now, we have had some very limited, mostly internal support for unique pointers in vecmem. This is a shame, because I think it would be really good to have outward-facing support for these types too. This pull request adds two new types and two new functions to the vecmem ecosystem which are designed to let us use unique pointers in host code for easy memory management.

The two new classes are `vecmem::unique_obj_ptr` and `vecmem::unique_alloc_ptr`. This division into two types of unique pointer is important, because we need to be able to make this work with both host-accessible memory and host-inaccessible memory. The semantics, roughly, are as follows. Unique object pointers are designed to work with any type, but they require the memory resource they use to be host-accessible. On the other side, unique allocation pointers are designed to work only with trivially constructible and destructible types, but they can work with any memory resource, even host-inaccessible ones. These restrictions are enforced via SFINAE.

Both of these types seamlessly support non-array types `T` and unbounded array types `T[]`. To stay in line with what the C++ standard library does, bounded array types `T[N]` are not supported.

Principally, these objects are designed to be built using four functions:
1. `make_unique_obj<T>(memory_resource&, Args&&...)` creates a new object, constructing it by forwarding the given arguments. This is analogous to `std::make_unique<T>(Args&&...)`. Works only for non-array types.
2. `make_unique_obj<T[]>(memory_resource&, std::size_t)` creates a new array, constructing each element using the default constructor. This is, again, analogous to `std::make_unique<T[]>(std::size_t)`. Work only for array types, for obvious reasons.
3. `make_unique_alloc<T>(memory_resource&)` creates a new trivial object, but does not touch the memory in any way to allow for usage with host-inaccessible memory resources.
4. `make_unique_alloc<T[]>(memory_resource&, std::size_t)` creates a new array, leaving every value unititialized (but in a valid state, because `T` must be trivially constructible).

Let me know what you think of this idea. :smile:


stephenswat:
Okay, sorry, but that MSVC error is utter nonsense...

tests/core/test_core_unique_ptr.cpp
```diff
@@ -0,0 +1,174 @@
...
+ */
+
+#include <gtest/gtest.h>
+
+#include "vecmem/memory/host_memory_resource.hpp"
+#include "vecmem/memory/memory_resource.hpp"
+#include "vecmem/memory/unique_ptr.hpp"
+
+namespace {
+static vecmem::host_memory_resource mr;
```
krasznaa, line 16: Why is the MSVC error nonsense? In that context the `mr` name hides the `mr` variable declared here.

You can argue that we shouldn't care about such symbol name hiding in the first place. But as long as that is enabled as a warning, MSVC is correctly identifying that that is what's happening here... 🤔

stephenswat: It's identifying the shadowing correctly, but think about what the consequence of this is: you would be completely unable to ever use any global variables because you would run the risk of importing some header that uses the same name in a function somewhere... That is totally unworkable. Sure, you could argue that you should not have global variables in the first place, but when writing tests they are pretty useful...

krasznaa: :confused:  We are building VecMem with a higher-than-default warning level.

https://github.com/acts-project/vecmem/blob/main/cmake/vecmem-compiler-options-cpp.cmake#L32

We've already come across this exact behaviour in #93. As we discussed back then, template instantiations work in curious ways in MSVC. Also as we discussed there, we should just respect this behaviour of MSVC at `/W4`. Client code of VecMem doesn't have to do so though.

So just rename the memory resource in the test (`mr` is a bad name for a global variable anyway), and let's be done with this...

stephenswat: Done. :slightly_smiling_face:

paulgessinger:
Neat!

core/include/vecmem/memory/unique_ptr.hpp
```diff
@@ -0,0 +1,462 @@
+#include <type_traits>
```
krasznaa, line 1: Please add the copyright statement to the new files.

core/include/vecmem/memory/unique_ptr.hpp
```diff
@@ -0,0 +1,462 @@
...
+
+#include "vecmem/memory/memory_resource.hpp"
+
+namespace vecmem {
+namespace {
```
krasznaa, line 6: I don't think that an unnamed namespace would do much good in a header file. :thinking: I don't think I've seen such a setup in a public header before... I would prefer to put these types into the `vecmem::details::` namespace. As we've done for a number of other types already.

core/include/vecmem/memory/unique_ptr.hpp
```diff
@@ -0,0 +1,462 @@
...
+ *
+ * @warning This type should never be used with host-inaccessible memory! We
+ * cannot check this at compile-time or at run-time, so it is the
+ * responsibility of the user to ensure that this requirement is respected.
+ */
+template <typename T,
+          std::enable_if_t<!std::is_array_v<T> ||
+                               (std::is_array_v<T> && std::extent_v<T> == 0),
+                           bool> = true>
+struct unique_obj_deleter {
```
krasznaa, line 35: Does this need to be a template? :thinking: The exact type is not needed by the class. It just needs to know whether the type is an array or not.

I would rather create two different deleter classes. One meant for arrays, and one for single objects. We could make `vecmem::unique_obj_ptr` choose between the two using a traits class. Like:

```c++
template <typename T>
using unique_obj_ptr = std::unique_ptr<T, details::unique_obj_deleter<T>::type>;
```

I'm just a pretty keen on putting as much of the code into compiled sources as possible, and only keeping stuff templated that really needs to be.

core/include/vecmem/memory/unique_ptr.hpp
```diff
@@ -0,0 +1,462 @@
...
+     *
+     * @param mr The memory resource to use for deallocation.
+     * @param p The pointer which we will want to deallocate.
+     * @param s The size of the allocation.
+     * @param a The alignment of the allocation, with 0 representing no
+     * alignment.
+     * @param n The number of elements in the allocation.
+     */
+    unique_obj_deleter(memory_resource& mr, pointer_t p, std::size_t s,
+                       std::size_t a = 0, std::size_t n = 1)
```
krasznaa, line 66: Shouldn't we just take the default alignment from [std::pmr::memory_resource](https://en.cppreference.com/w/cpp/memory/memory_resource)? I.e.:
  - https://en.cppreference.com/w/cpp/memory/memory_resource/allocate
  - https://en.cppreference.com/w/cpp/memory/memory_resource/deallocate

```c++
alignof(std::max_align_t)
```

Then there would be no reason for the `if (m_align > 0)` check later on.

core/include/vecmem/memory/unique_ptr.hpp
```diff
@@ -0,0 +1,462 @@
...
+     * @brief Activate the deletion mechanism of the deleter.
+     *
+     * @param p The pointer to deallocate.
+     *
+     * @note The pointer argument to this function is in some ways obsolete,
+     * because we store the pointer as a member variable. However, having both
+     * makes it easier to cross-check the logic.
+     */
+    void operator()(pointer_t p) const {
+        assert(p == m_ptr);
```
krasznaa, line 115: The `m_ptr` variable is really only used for this one line. Is it worth to declare it just for this? :thinking:

core/include/vecmem/memory/unique_ptr.hpp
```diff
@@ -0,0 +1,462 @@
...
+ * @note This deleter class only supports types that are both trivially
+ * constructible and trivially destructible.
+ */
+template <
+    typename T,
+    std::enable_if_t<std::is_trivially_constructible_v<std::remove_extent_t<T>>,
+                     bool> = true,
+    std::enable_if_t<std::is_trivially_destructible_v<std::remove_extent_t<T>>,
+                     bool> = true>
+struct unique_alloc_deleter {
```
krasznaa, line 190: Just like for `unique_obj_deleter`, I would prefer to make this class non-templated. After all, it only has a single implementation anyway.

As before, we can put the SFINAE logic (without actual SFINAE...) into a traits class instead.

tests/core/test_core_unique_ptr.cpp
```diff
@@ -0,0 +1,174 @@
...
+public:
+    B(void) = delete;
+
+    B(int, float) { ++g_construct; }
+
+    ~B() { ++g_destruct; }
+};
+}  // namespace
+
+TEST(CoreUniqueObjPtrTest, EmptyPointerPrimitive) {
```
krasznaa, line 38: I like `TEST_F` a little better, as it ensures that you've attached the tests to the same test case. While with `TEST` you can make typos, and the compiler doesn't complain about it.

Though this is a highly subjective comment... Even if it would also allow you to put the memory resource into the test case, along with the test classes and counters. Which would even allow (with some additional tweaks) to get rid of the `static`-ness of the counters all together.

krasznaa (changes requested):
We seem to have come full circle with this PR... :thinking: Coming back to the constructs that triggered the VecMem developments in the first place...

https://github.com/acts-project/acts/blob/main/Plugins/Cuda/include/Acts/Plugins/Cuda/Utilities/Arrays.hpp

core/include/vecmem/memory/unique_ptr.hpp
stephenswat: It does pretty much what you would expect: it creates a translation-unit-local anonymous namespace, except it's still inside of the `vecmem`. But sure, I can move them to `vecmem::details` no problem. :smile:

core/include/vecmem/memory/unique_ptr.hpp
stephenswat: Well, we need to make sure that `std::unique_ptr` only accepts certain kinds of types for its different partial applications, and the only way I see to do that elegantly would be to add SFINAE mechanisms to the deleter objects; that way, the specialization failures bubble up to the level of `std::unique_ptr` nicely, and that allows us to avoid having to reimplement `std::unique_ptr` completely. Do you have any other ideas of how you could achieve this without templating?

core/include/vecmem/memory/unique_ptr.hpp
```diff
@@ -0,0 +1,462 @@
...
+
+    /**
+     * @brief Default-construct a new unique object deleter.
+     *
+     * Having a default constructor for this class is not ideal, as having an
+     * empty deleter here doesn't really make any sense. However, if the
+     * deleter class is not default-constructible, then neither is the
+     * std::unique_ptr class using it. For that reason, we want it here.
+     */
+    unique_obj_deleter(void) = default;
```
paulgessinger, line 49: The void is C, right?. Is this necessary in C++?

core/include/vecmem/memory/unique_ptr.hpp
stephenswat: Okay, I see what you mean - I think. I'll see if this can be done using a type trait.

core/include/vecmem/memory/unique_ptr.hpp
stephenswat: No, it's not necessary. Just personal preference from when I used to write C. Can get rid of it.

paulgessinger: Ah not necessary, just curious.

core/include/vecmem/memory/unique_ptr.hpp
krasznaa: I played a bit just now. This may not be the greatest way of doing it, but it seems to (sort of...) work:

```c++
#include <iostream>
#include <memory>
#include <type_traits>

class A {
public:
   A() = default;

   int value() const { return m_value; }
   void setValue(int value) { m_value = value; }

private:
   int m_value;
};

class B {
public:
   B() = default;

   int value() const { return m_value; }
   void setValue(int value) { m_value = value; }

private:
   int m_value = 42;
};

struct unique_alloc_deleter {
   void operator()(void* ptr) {
      std::cout << "Deleting: " << ptr << std::endl;
   }
};

template <typename T>
struct unique_alloc_deleter_t {
  struct chosen_type_is_not_trivially_constructible {};
  using type = std::conditional_t<std::is_trivially_constructible<T>::value, unique_alloc_deleter,
				  chosen_type_is_not_trivially_constructible>;
};

template <typename T>
using unique_alloc_ptr = std::unique_ptr<T, typename unique_alloc_deleter_t<T>::type>;

int main() {

   std::cout << "A is trivially constructible: " << std::is_trivially_constructible<A>::value << std::endl;
   std::cout << "B is trivially constructible: " << std::is_trivially_constructible<B>::value << std::endl;

   unique_alloc_ptr<A> a(reinterpret_cast<A*>(0x1234));
   unique_alloc_ptr<B> b(reinterpret_cast<B*>(0x2345));

   return 0;
}
```

This produces the following compilation error:

```
[bash][Legolas]:cpp > g++ -std=c++17 trivial_construct.cpp 
In file included from /usr/include/c++/9/memory:80,
                 from trivial_construct.cpp:3:
/usr/include/c++/9/bits/unique_ptr.h: In instantiation of ‘std::unique_ptr<_Tp, _Dp>::~unique_ptr() [with _Tp = B; _Dp = unique_alloc_deleter_t<B>::chosen_type_is_not_trivially_constructible]’:
trivial_construct.cpp:63:54:   required from here
/usr/include/c++/9/bits/unique_ptr.h:288:56: error: static assertion failed: unique_ptr's deleter must be invocable with a pointer
  288 |  static_assert(__is_invocable<deleter_type&, pointer>::value,
      |                                                        ^~~~~
/usr/include/c++/9/bits/unique_ptr.h:292:17: error: no match for call to ‘(std::unique_ptr<B, unique_alloc_deleter_t<B>::chosen_type_is_not_trivially_constructible>::deleter_type {aka unique_alloc_deleter_t<B>::chosen_type_is_not_trivially_constructible}) (std::remove_reference<B*&>::type)’
  292 |    get_deleter()(std::move(__ptr));
      |    ~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~
[bash][Legolas]:cpp >
```

What's the exact error message with SFINAE? I just wonder which one would be more understandable...

krasznaa: :confused: Honestly, what is the SFINAE output supposed to look like here? Since even with a pretty pure setup I don't get super readable error messages.

```c++
#include <iostream>
#include <memory>
#include <type_traits>

class A {
public:
   A() = default;

   int value() const { return m_value; }
   void setValue(int value) { m_value = value; }

private:
   int m_value;
};

class B {
public:
   B() = default;

   int value() const { return m_value; }
   void setValue(int value) { m_value = value; }

private:
   int m_value = 42;
};

struct unique_alloc_deleter {
   void operator()(void* ptr) {
      std::cout << "Deleting: " << ptr << std::endl;
   }
};

template <typename T, std::enable_if_t<std::is_trivially_constructible<T>::value, bool> = true>
using unique_alloc_ptr = std::unique_ptr<T, unique_alloc_deleter>;

int main() {

   std::cout << "A is trivially constructible: " << std::is_trivially_constructible<A>::value << std::endl;
   std::cout << "B is trivially constructible: " << std::is_trivially_constructible<B>::value << std::endl;

   unique_alloc_ptr<A> a(reinterpret_cast<A*>(0x1234));
   unique_alloc_ptr<B> b;

   return 0;
}
```

With this I get:

```
[bash][Legolas]:cpp > g++ -std=c++17 trivial_construct.cpp 
trivial_construct.cpp: In function ‘int main()’:
trivial_construct.cpp:56:22: error: no type named ‘type’ in ‘struct std::enable_if<false, bool>’
   56 |    unique_alloc_ptr<B> b;
      |                      ^
trivial_construct.cpp:56:22: note: invalid template non-type parameter
[bash][Legolas]:cpp >
```

Since we only provide a single implementation for this type, SFINAE does not seem to kick in anyway. So my previous proposal with a "descriptively named struct" may be the most readable output that we can generate...

stephenswat: SFINAE never really gives you readable messages, does it? That only happens when you start using `static_assert`, and of course `static_assert` does not really participate in SFINAE. Anyway, these classes have some of their inner logic depend on template types as well. We would need to find a way to encapsulate that in a type trait - or reduce their parameter space and explicitly specialize them - in order to make them work.

krasznaa: I would've expected a message like "`unique_alloc_ptr<B>` type is unknown", or something similar. This is what I would've understood by your "the specialization failures bubble up to the level of `std::unique_ptr` nicely" comment. :confused:

I don't know what you mean about the inner logic depending on the template type... For this class (`unique_alloc_deleter`) there is no type dependence whatsoever. For `unique_obj_deleter` you would have to split it into two classes. (As I wrote in that comment.)

So I'm still very much on the side of implementing these deleters as non-template classes. I just really don't like over-using templates... :frowning: (Which is unfortunately very hard not to do in this project.)

core/include/vecmem/memory/unique_ptr.hpp
stephenswat: In C-land, `f()` can mean a function `f` with an unknown number of arguments; the only way to really define a function with no arguments in C is `f(void)`, and I think it's just stuck with me that this is the nicer way to express that. In C++, `f()` and `f(void)` are equivalent, as far as I know.

`void` is the C/C++ equivalent of what we call the unit type, which is inhabited by exactly one value `()` (written that way because it is the unit of the product types (also known as tuples) - it's basically a tuple of zero types). _Very_ confusingly, it's not the uninhabited _void_ type, which doesn't exist in C or C++... Anyway, adding the void argument is a vestigial way to say that the function only depends on the implicitly constructible unit type `void`, rather than being variadic.

I'll put the type theory rambles back into the cupboard now. :smile:

core/include/vecmem/memory/unique_ptr.hpp
stephenswat: I think I would prefer to keep it this way: once we do the allocation, there is no way to re-obtain the alignment later, and it is a required argument to the deallocate function. Right now, there are no cases where we don't use the default alignment, but I think it would be a solid idea to keep the possibility open to support this in the future. In general, I think we've been a little (too) lax in supporting alignment throughout vecmem as a whole, and I am quite worried that this will come bite us in the behind at some point.

On the other hand the YAGNI principle would dictate not having this, so I am happy to go either way.

core/include/vecmem/memory/unique_ptr.hpp
stephenswat: Sure, I'll get rid of it.

core/include/vecmem/memory/unique_ptr.hpp
krasznaa: I think you misunderstood me... I don't want to get rid of the `m_align` variable. I was just proposing to replace `0` with `alignof(std::max_align_t)` as its default value in the constructor.

Then the deallocation could simply write:

```c++
m_mr->deallocate(p, m_size, m_align);
```

Instead of the current:

```c++
        if (m_align > 0) {
            m_mr->deallocate(p, m_size, m_align);
        } else {
            m_mr->deallocate(p, m_size);
        }
```

tests/core/test_core_unique_ptr.cpp
stephenswat: I've moved all the tests into fixtures and gotten rid of the global variables. I also split the tests into two files, to make it even clearer that there are two sets of tests here. Let me know if this is to your liking.

core/include/vecmem/memory/unique_ptr.hpp
```diff
@@ -0,0 +1,462 @@
+/*
+ * VecMem project, part of the ACTS project (R&D line)
+ *
+ * (c) 2021 CERN for the benefit of the ACTS project
+ *
+ * Mozilla Public License Version 2.0
+ */
+
```
krasznaa, line 8: I only noticed this now as well... There's no `#pragma once` at the top...

core/include/vecmem/memory/unique_ptr.hpp
stephenswat: It would be possible, but I must admit I don't really see the point. The only thing I see this changing is that you start making some assumptions on the default value of `std::pmr::memory_resource::allocate`'s default alignment argument - if those change for whatever reason (unlikely, sure, but not impossible) then suddenly the deleter is no longer correct. The current implementation treats the memory resource as the single source of truth, rather than making these assumptions. What would you say is the advantage of having `std::max_align_t` be the default?

core/include/vecmem/memory/unique_ptr.hpp
stephenswat: Okay, I've looked at this a little more and I can see this being feasible for trivial types, but I am still a little bit miffed about how we can call the destructors of types without templating. _Maybe_ this would be possible if the type has virtual destructors, but that seems like a dangerous assumption to make. Do you have any ideas on how to make this work?

krasznaa: You got me with the destructor... Indeed, `unique_obj_deleter` does need the concrete type to call the destructor correctly. So that one does need to remain a template... :thinking:

But `unique_alloc_deleter` does not need to do that. For this class we can really just call `deallocate` on the memory resource, and that's it. And that we can very much do through a `void*` pointer. And `std::unique_ptr` absolutely allows the deleter to receive a `void*` pointer.

stephenswat: Alright, I've removed the templates as far as I deem possible, and I've moved the implementation to a separate `.cpp` file. I hope this is what you intended. :smile:

core/include/vecmem/memory/unique_ptr.hpp
krasznaa: :thinking: I assumed that if cppreference spelled it out, the default argument value might be part of the standard. But the standard indeed does not seem to define what the default value should be...

http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2017/n4659.pdf

(Section 23.12.2.)

So fine, let's use 0 as the default/undefined value then...

core/src/memory/details/unique_alloc_deleter_impl.cpp
```diff
@@ -0,0 +1,61 @@
...
+ *
+ * Mozilla Public License Version 2.0
+ */
+
+#include <cassert>
+
+#include "vecmem/memory/unique_ptr.hpp"
+
+namespace vecmem::details {
+unique_alloc_deleter_impl::unique_alloc_deleter_impl(void) = default;
```
krasznaa, line 14: Huhh... I didn't know that you could do this... But yeah, let's put the default implementations into the `.cpp` file!

I would just make the other constructors/assignment functions "default" as well.

krasznaa (commented):
I myself would've preferred a "neater file naming", but that is not a deal breaker at this point...

It's neat though that once `mr` is not declared as a global variable in the tests, MSVC is not complaining about it anymore. :smile: So all in all, I would argue that the warning from it is not so bananas as it would seem at first. :smile:

stephenswat:
Okay, I've re-done the file layout in (what you will hopefully agree) a neater way. :smile: You're absolutely right about the constructors, they can be defaulted and most are not necessary, so I defaulted them and removed some to simplify the code.

stephenswat:
Oh, I guess MSVC can't default `operator=`? :confused:

stephenswat:
Maybe it will accept this.

krasznaa (approved):
All right, let's use this one. Even if the

```c++
    using pointer_t =
        std::conditional_t<std::is_array_v<T>, std::decay_t<T>, T*>;
```

declarations in the `make_unique_alloc` functions are not really necessary. Since those are already specializations for array and non-array types. So the `std::conditional_t` in the function body is really not needed. :stuck_out_tongue:

PR #156, opened 2021-12-13
Use new unique pointers internally

stephenswat:
Previously, we were using some unique pointers internally for, for example, our array types and our buffer types. This pull request changes this to use the new `vecmem::unique_obj_ptr` and `vecmem::unique_alloc_ptr` types, which are drop-in replacements. This homogenizes the interal and external representation of these smart pointers.


krasznaa (approved):
Looks good! No notes. :smile: (I did run all the CUDA/HIP/SYCL tests locally, just to make sure that all of the tests would run successfully. Since I've made plenty of mistakes with those buffer classes in the past...)

stephenswat:
Sweet! I was pleasantly surprised how easy this switch was to make, to be honest. :smiley:

PR #154, opened 2021-12-10
Reducing alloc calls in arena memory resource

Yhatoh:
Hello,

Sorry for the time i take it for this. 

I tried several forms to reduce the allocations calls, the solution a made was creating a value of super_blocks, the idea is when you need more memory made an alloc of a big size of memory (more than the asked, in the case you need more memory in the future). This made a reduction in the calls of allocs (or that is what i saw).

Also, i fixed something in the  destructor, because doing the other solution i figured out a weird case, but i change the destructor and it works.

I am attentive to any comments :D. 

Cheers,
Gabriel

PR #157, opened 2021-12-16
Reduce Allocations in Arena Memory Resource, main branch (2021.12.16.)

krasznaa:
Manually copied @Yhatoh's update from #154 into a single commit.

Instead of trying to squash the commits from that PR into a single one, and then rebase it on top of the current main branch (which I tried, but after the second set of conflicts I just gave up), it seemed easier to just create a new commit by hand.

These updates are meant to reduce the number of allocations from the arena allocator. Also fixing an issue in the `vecmem::details::arena` destructor, and hiding `vecmem::details::minimum_superblock_size` from the outside world.


krasznaa:
@stephenswat, I'd leave the review to you. I'm not convinced I fully understand what the code is doing. :frowning:

stephenswat:
I will take a look at it after lunch. Either way, we should set the author on this commit to `Yhatoh <gacarmonat@gmail.com>`. :smile:

Yhatoh:
Hi!

First at all, happy new year! I wanted to ask in what state is this PR. Do you need more information about the change or maybe do you found something about the changes that are weird?

Cheers,
Gabriel

krasznaa:
Stephen had enough time to raise concerns at this point. :stuck_out_tongue: Since all the existing tests find the changes to be okay, let's just go forward with them. :wink:

Issue #159, opened 2022-01-25
vecmem::data::vector_buffer doesn't work for `unsigned long`


beomki-yeo:
CUDA version : 11.5.0

`vector_buffer` doesnt work with `unsigned long` with following error message:

`what():  /home/beomki/projects/vecmem/vecmem/cuda/src/memory/cuda/managed_memory_resource.cpp:33 Failed to execute: cudaFree(p) (misaligned address)`

`int` or `unsigned int` was OK

you can reproduce the error with the following codes

### test_cuda_containers.cpp

```c++
TEST_F(cuda_containers_test, buffer_type_test) {

    // The memory resource(s).
    vecmem::cuda::managed_memory_resource managed_resource;

    vecmem::data::vector_buffer<int> int_vec(3, 0, managed_resource);
    m_copy.setup(int_vec);

    int_buffer_test(int_vec); // OK

    vecmem::data::vector_buffer<unsigned int> uint_vec(3, 0, managed_resource);
    m_copy.setup(uint_vec);

    uint_buffer_test(uint_vec); // OK

    vecmem::data::vector_buffer<unsigned long> ulong_vec(3, 0,
                                                         managed_resource);
    m_copy.setup(ulong_vec);

    ulong_buffer_test(ulong_vec);  // doesnt work
}
```

### test_cuda_containers_kernels.cuh

```c++
void int_buffer_test(vecmem::data::vector_view<int> vec);

void uint_buffer_test(vecmem::data::vector_view<unsigned int> vec);

void ulong_buffer_test(vecmem::data::vector_view<unsigned long> vec);
```

### test_cuda_containers_kernels.cu

```c++
__global__ void int_buffer_test_kernel(
    vecmem::data::vector_view<int> vec_data) {
    vecmem::device_vector<int> vec(vec_data);
    vec.push_back(0);
}
void int_buffer_test(vecmem::data::vector_view<int> vec) {
    int_buffer_test_kernel<<<1, 1>>>(vec);
    // Check whether it succeeded to run.
    VECMEM_CUDA_ERROR_CHECK(cudaGetLastError());
    VECMEM_CUDA_ERROR_CHECK(cudaDeviceSynchronize());
}
__global__ void uint_buffer_test_kernel(
    vecmem::data::vector_view<unsigned int> vec_data) {
    vecmem::device_vector<unsigned int> vec(vec_data);
    vec.push_back(0);
}
void uint_buffer_test(vecmem::data::vector_view<unsigned int> vec) {
[...]

PR #160, opened 2022-01-25
Buffer Alignment Fixes, main branch (2022.01.25.)

krasznaa:
As @beomki-yeo discovered in #159, the vector buffers were not laying out their memory quite correctly until now.

When using an element type with `sizeof(TYPE) > sizeof(unsigned int)`, it could happen that the pointers provided by the buffers for the vector elements would not have an alignment that CUDA device code would be happy with. Resulting in a cryptic error from the CUDA runtime.

This PR introduces code that would pad the buffers in a way that would make the vector elements aligned correctly for both 1D and 2D (jagged) re-sizable vectors. (Since the issue was only there for re-sizable vectors in the first place.)

I also added tests inspired by Beomki's code to explicitly test both `vecmem::data::vector_buffer` and `vecmem::data::jagged_vector_buffer` for this from now on.

Others thinking about this a bit will definitely help. As I'm not absolutely convinced that I did this correctly. Actually, I'm a bit more convinced about the code I wrote for jagged vectors than the one I did for "normal" ones. (I wrote the jagged vector code after the normal one...)

### DIFF
--- a/core/include/vecmem/containers/impl/jagged_vector_buffer.ipp
+++ b/core/include/vecmem/containers/impl/jagged_vector_buffer.ipp
@@ -10,6 +10,7 @@
 // System include(s).
 #include <cassert>
 #include <cstddef>
+#include <memory>
 #include <numeric>
 #include <vector>
 
@@ -25,18 +26,6 @@ std::vector<std::size_t> get_sizes(
         result[i] = jvv.m_ptr[i].size();
     }
     return result;
-}
-
-/// Get the padding needed after the size array for correct alignment
-template <typename TYPE>
-std::size_t get_buffer_alignment_padding(std::size_t size) {
-
-    // Total size of the size array.
-    const std::size_t size_size =
-        size * sizeof(typename vecmem::data::jagged_vector_buffer<
-                      TYPE>::value_type::size_type);
-    // Return the padding needed after the size array.
-    return (alignof(TYPE) - (size_size % alignof(TYPE)));
 }
 
 /// Function allocating memory for @c vecmem::data::jagged_vector_buffer
@@ -60,19 +49,27 @@ allocate_jagged_buffer_outer_memory(
 template <typename TYPE>
 vecmem::unique_alloc_ptr<char[]> allocate_jagged_buffer_inner_memory(
     const std::vector<std::size_t>& sizes, vecmem::memory_resource& resource,
-    bool isResizable) {
-    typename vecmem::data::jagged_vector_buffer<TYPE>::size_type byteSize =
-        std::accumulate(sizes.begin(), sizes.end(),
-                        static_cast<std::size_t>(0)) *
-        sizeof(TYPE);
+    bool isResizable, std::size_t& bufferSize) {
+
+    // Alignment for the vector elements.
+    static constexpr std::size_t alignment = alignof(TYPE);
+
+    // Determine the allocation size.
+    bufferSize = std::accumulate(sizes.begin(), sizes.end(),
+                                 static_cast<std::size_t>(0)) *
+                 sizeof(TYPE);
+    // Increase this size if the buffer describes a resizable vector.
     if (isResizable) {
-        byteSize +=
+        bufferSize +=
             sizes.size() * sizeof(typename vecmem::data::jagged_vector_buffer<
-                                  TYPE>::value_type::size_type) +
-            get_buffer_alignment_padding<TYPE>(sizes.size());
+                                  TYPE>::value_type::size_type);
+        // Further increase this size so that we could for sure align the
+        // payload data correctly.
+        bufferSize = ((bufferSize + alignment - 1) / alignment) * alignment;
     }
 
-    return vecmem::make_unique_alloc<char[]>(resource, byteSize);
+    // Return a smart pointer with the allocation.
+    return vecmem::make_unique_alloc<char[]>(resource, bufferSize);
 }
 
 }  // namespace
@@ -101,8 +98,9 @@ jagged_vector_buffer<TYPE>::jagged_vector_buffer(
           sizes.size(),
           (host_access_resource == nullptr ? resource
                                            : *host_access_resource))),
-      m_inner_memory(
-          ::allocate_jagged_buffer_inner_memory<TYPE>(sizes, resource, false)) {
+      m_inner_memory_size(0u),
+      m_inner_memory(::allocate_jagged_buffer_inner_memory<TYPE>(
+          sizes, resource, false, m_inner_memory_size)) {
 
     // Point the base class at the newly allocated memory.
     base_type::m_ptr =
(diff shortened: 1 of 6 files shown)


stephenswat:
Looks good, but I think this could be made more robust (and simpler!) by relying a little more on the standard library:

1. Take your header data type _s_ of which you need _n_, and your value data type _t_ of which you need _m_.
2. Calculate the maximum alignment of your types: `constexpr std::size_t a = std::max(alignof(s), alignof(t));`
3. Calculate the total allocation size: `constexpr std::size_t b = n * sizeof(s) + m * sizeof(t);`
4. Round the size to a multiple of `a`: `constexpr std::size_t b' = ((b + a - 1) / a) * a`;
5. Allocate some region of memory starting at `p` of size `b'` (this doesn't need to be aligned to `a`, but it can be).
6. Use `std::align` to handle the alignments:
    ```cpp
   // Total size of the allocation, rounded up.
   std::size_t bp = ...;

   // Some (possibly unaligned) allocation.
   void * p memory = allocate(bp);

   // Mutable copies of the allocation size and pointer.
   std::size_t r = bp;
   void * q = p;

   // Calculate the size of the header.
   std::size_t header_size = n * sizeof(s);

   // Request the first address in the memory that is aligned to the alignment
   // requirements of the header, and reserve the needed memory.
   // This returns a pointer, and updates q and r (with the alignment only).
   void * m_header_ptr = std::align(alignof(s), header_size, q, r);

   // Manually add the allocated size to our working pointer, and decrease the
   // remaining size of our allocation.
   q = static_cast<void *>(static_cast<char *>(p) + header_size);
   r -= header_size;

   // Calculate the size of the data.
   std::size_t data_size = m * sizeof(t);

   // Again, request the first address _past the header_ that fits the data
   // alignment requirements. If all is well, this should always succeed.
   void * m_data_ptr = std::align(alignof(t), data_size, q, r);

   // Fail-safe, if this happens something is very wrong
   assert(m_data_ptr);
    ```

stephenswat:
Also, we might want to extend the tests with some ludicrously aligned types, e.g.:

```cpp
struct alignas(512) T {
   int _;
};
```

krasznaa:
The way that allocations with CUDA, HIP and SYCL work, the address given by those language specific allocators is **always** aligned. Quite a bit so. Since in my tests I've been seeing printouts like:

```
[ RUN      ] cuda_containers_test.large_buffer
[vecmem] cuda/src/memory/cuda/managed_memory_resource.cpp:25 Allocated 32 bytes at 0x7f985c000000
[vecmem] cuda/src/utils/cuda/copy.cpp:76 Set 4 bytes to 0 at 0x7f985c000000 with CUDA
[vecmem] core/include/vecmem/utils/impl/copy.ipp:31 Prepared a device vector buffer of capacity 3 for use on a device (ptr: 0x7f985c000000)
[vecmem] cuda/src/utils/cuda/copy.cpp:55 Performed unknown memory copy of 4 bytes from 0x7f985c000000 to 0x7ffc59dd91c4
[vecmem] cuda/src/memory/cuda/managed_memory_resource.cpp:25 Allocated 72 bytes at 0x7f985c000200
[vecmem] cuda/src/memory/cuda/managed_memory_resource.cpp:25 Allocated 88 bytes at 0x7f985c000400
[vecmem] cuda/src/utils/cuda/copy.cpp:76 Set 12 bytes to 0 at 0x7f985c000400 with CUDA
[vecmem] cuda/src/utils/cuda/copy.cpp:55 Performed unknown memory copy of 12 bytes from 0x7f985c000400 to 0x560fae0f32a0
[vecmem] cuda/src/memory/cuda/managed_memory_resource.cpp:32 De-allocating memory at 0x7f985c000400
[vecmem] cuda/src/memory/cuda/managed_memory_resource.cpp:32 De-allocating memory at 0x7f985c000200
[vecmem] cuda/src/memory/cuda/managed_memory_resource.cpp:32 De-allocating memory at 0x7f985c000000
[       OK ] cuda_containers_test.large_buffer (16 ms)
```

So I don't think we need to worry about the alignment of the "size variables".

However I'm very receptive to suggestions on making the alignment of the "data area" inside of the allocation more robust. But I don't think [std::align](https://en.cppreference.com/w/cpp/memory/align) is appropriate for this. :thinking: Since we do need to know the total size of the allocation before we would have a chance of calling `std::align` on a pointer.

The recommendation about using a type with a custom allocation is a good one.

stephenswat:
> The way that allocations with CUDA, HIP and SYCL work, the address given by those language specific allocators is always aligned. Quite a bit so.

This is absolutely true, but in this case I would personally prefer to be too general than too specific. It's possible (although unlikely) that someone might make a buffer on a host memory resource, which does not have the alignment you mentioned. Just because memory bugs are often _so_ nasty to debug, I think we should make the effort here to make sure this _always_ works.

> But I don't think std::align is appropriate for this. thinking Since we do need to know the total size of the allocation before we would have a chance of calling std::align on a pointer.

This is exactly what my pseudocode does, though, right? I don't personally see the problem. You just allocate a little bit too much memory, to make sure you've always got enough. Size the excess memory is always less than the alignment size this overhead goes to zero for any reasonably sized allocation anyway.

krasznaa:
:thinking: Should we still just delegate the alignment of the "first" address (the one used for the size array) to the memory resource? After all, memory resources do have an alignment parameter in their `allocate(...)` functions.

Then we could just deal with hand-aligning the pointer for the "data array" in the buffer classes.

stephenswat:
The benefit of that is that if you align your initial request to the maximum alignment of your header and data types, you are guaranteed to be able to start your header right at the start of your allocation, but you still need to pad the start of your data region (which we can do with `std::align`).

The downside is that we then need to start implementing alignment in all of our up- and downstream memory resources. :worried:

krasznaa:
Call me lazy, but I would skip the type with the custom alignment in this PR after all... :frowning: We should add it afterwards once we find the time...

core/include/vecmem/containers/impl/jagged_vector_buffer.ipp
```diff
@@ -129,16 +141,29 @@ jagged_vector_buffer<TYPE>::jagged_vector_buffer(
...
-    std::ptrdiff_t ptrdiff =
+    void* unaligned_start_ptr =
+        m_inner_memory.get() +
         (capacities.size() * sizeof(typename value_type::size_type));
+    std::size_t buffer_size =
+        m_inner_memory_size -
+        (capacities.size() * sizeof(typename value_type::size_type));
+    char* start_ptr = static_cast<char*>(
+        std::align(alignof(TYPE), dataSize, unaligned_start_ptr, buffer_size));
+    std::ptrdiff_t ptrdiff = 0;
```
paulgessinger, line 159: Call me stupid, but this code could use some more comments explaining what's happening.

krasznaa: Really, it's not obvious right away? :smile: After I composed the lines with having 4-5 different tabs open that I would switch between, to try to come up with the correct incantation... :stuck_out_tongue:

paulgessinger:
Maybe the top level documentation of `jagged_vector_buffer` and `vector_buffer` could get some brief explanation of the memory layout and its alignment, what do you think?

stephenswat:
Okay, let's go ahead with this. We can revisit later.

stephenswat (approved):

krasznaa:
> Just to be clear, looking at the code: we're going for the approach of assuming that the memory resource returns sufficiently aligned memory?

Yes, after some pondering I decided that for now I would just go with that. Since I really believe that to force the "beginning" of an allocation returned by a memory resource to be aligned somehow, we really must use the machinery provided by `vecmem::memory_resource`. And yes, this does mean that we'll need to be careful in all our downstream memory resources... (The upstream ones for now seem to be pretty conservative with alignment anyway.)

PR #161, opened 2022-01-25
Improve alignment guarantees of (jagged) vector buffers

stephenswat:
> We may want to think about opening a task for doing some review of the alignment handling in the code in a better way later on, but we can sleep on that once at least...

I skipped the sleeping part, here is a continuation of #160 which provides some stronger guarantees on memory alignment.

Work in progress for now.


stephenswat:
Okay, I had a lot of fun with this. Too much, perhaps.

core/include/vecmem/containers/details/aligned_multiple_placement.hpp
```diff
@@ -0,0 +1,22 @@
+/*
+ * VecMem project, part of the ACTS project (R&D line)
+ *
+ * (c) 2021 CERN for the benefit of the ACTS project
```
krasznaa, line 4: You should use `2021-2022` in all the new/modified files. :wink:

core/include/vecmem/containers/impl/aligned_multiple_placement.ipp
```diff
@@ -0,0 +1,286 @@
...
+ * @tparam Ts The types to allocate memory for.
+ * @tparam Ps Positional parameter types, which should all be std::size or a
+ * compatible type, and the length must be equal to that of `Ts`.
+ * @param r The memory resource to use for allocation.
+ * @param ps The set of sizes of the differently typed regions.
+ * @returns A tuple of |Ts| + 1 elements, where the first element is a unique
+ * allocation pointer to the complete block of memory (such that it may be
+ * freed), and the other values represent the addresses in memory where the
+ * regions begin.
+ */
```
krasznaa, line 193: This is an excellent piece of documentation, which should not be hidden in the `.ipp` file. I think this should be moved into the `.hpp` file instead.

core/include/vecmem/containers/impl/aligned_multiple_placement.ipp
```diff
@@ -0,0 +1,286 @@
...
+template <typename... Ts, typename... Ps>
+std::tuple<vecmem::unique_alloc_ptr<char[]>, std::add_pointer_t<Ts>...>
+aligned_multiple_placement(vecmem::memory_resource &r, Ps... ps) {
```
krasznaa, line 196: Is there any upside to templating the size arguments? :thinking: Why not just specify explicitly in the interface that this needs to be a variadic list of `std::size_t`-s? That could save us one `static_assert(...)` in the code.

core/include/vecmem/containers/impl/jagged_vector_buffer.ipp
```diff
@@ -128,10 +103,18 @@ jagged_vector_buffer<TYPE>::jagged_vector_buffer(
...
-          capacities, resource, true, m_inner_memory_size)) {
+                                           : *host_access_resource))) {
+    using header_t = typename vecmem::data::jagged_vector_buffer<
+        TYPE>::value_type::size_type;
+    // Determine the allocation size.
+    std::size_t total_elements = std::accumulate(
+        capacities.begin(), capacities.end(), static_cast<std::size_t>(0));
+
+    header_t* header_ptr;
+    TYPE* data_ptr;
```
krasznaa, line 114: This is the static code analyzer talking... These pointers should be explicitly initialised to `nullptr`. Otherwise static analyzers will complain.

core/include/vecmem/containers/impl/jagged_vector_buffer.ipp
```diff
@@ -141,37 +124,13 @@ jagged_vector_buffer<TYPE>::jagged_vector_buffer(
...
     std::ptrdiff_t ptrdiff = 0;
     for (std::size_t i = 0; i < capacities.size(); ++i) {
         new (host_ptr() + i) value_type(
             static_cast<typename value_type::size_type>(capacities[i]),
-            reinterpret_cast<typename value_type::size_pointer>(
-                m_inner_memory.get() +
-                i * sizeof(typename value_type::size_type)),
-            reinterpret_cast<TYPE*>(start_ptr + ptrdiff));
-        ptrdiff += capacities[i] * sizeof(TYPE);
+            &header_ptr[i], &data_ptr[ptrdiff]);
```
krasznaa, line 132: The two are pretty much equivalent, but I think using `data_ptr + ptrdiff` instead of `&data_ptr[ptrdiff]` would be better here. That's why the variable is called `ptrdiff`, and is of type `std::ptrdiff_t` after all. :stuck_out_tongue:

core/include/vecmem/containers/impl/jagged_vector_buffer.ipp
```diff
@@ -14,6 +14,8 @@
 #include <numeric>
 #include <vector>
 
+#include "../details/aligned_multiple_placement.hpp"
```
krasznaa, line 17: Just because we didn't use relative path names in other places either, I would go with `vecmem/containers/details/aligned_multiple_placement.hpp` here as well. But of course this is a largely subjective point.

core/include/vecmem/containers/impl/aligned_multiple_placement.ipp
```diff
@@ -0,0 +1,286 @@
...
+ *
+ * Mozilla Public License Version 2.0
+ */
+
+#pragma once
+
+#include <tuple>
+#include <type_traits>
+
+#include "vecmem/utils/type_traits.hpp"
```
krasznaa, line 14: I notice that you don't like to put comments into the include statements. But that's the only way in which we can force clang-format to allow us putting our own includes before the standard library includes.

This comes from one of the ATLAS coding rules. That with the includes we should always go "from the most specific to the least specific headers". Of course this leaves a lot of room for interpretation still. But at the very least it means that our own headers need to be included before the standard library ones. (This is mainly to better test that those headers work well by themselves as well.)

core/include/vecmem/containers/details/aligned_multiple_placement.hpp
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * Mozilla Public License Version 2.0
+ */
+
+#pragma once
+
+#include <tuple>
+#include <type_traits>
```
krasznaa, line 12: You forgot to include `vecmem/memory/unique_ptr.hpp`.

krasznaa (changes requested):
You have indeed done some fancy coding here... :wink:

I have to admit I didn't chew through the code of `vecmem::details::aligned_multiple_placement(...)`. :frowning: I just assumed that the logic there would be sound.

Instead I just stuck to the annoying thing, and commented on a lot of more "formal" issues.

core/include/vecmem/containers/impl/aligned_multiple_placement.ipp
stephenswat: Oh, so _that_ is why we always comment the includes! I had no idea. :smile:

core/include/vecmem/containers/impl/aligned_multiple_placement.ipp
stephenswat: Is that... Possible? I know that `f(std::size_t ... args)` is invalid C++ syntax, how else would you do this? :worried:

stephenswat: You could do it with initializer lists if that would be your preferred option.

krasznaa: I think the incantation is:

```c++
template<std::size_t... SIZES>
void foo(SIZES... sizes) {
...
}
```

I've used such things for instance in:

https://gitlab.cern.ch/atlas/athena/-/blob/master/Control/AthCUDA/AthCUDAKernel/AthCUDAKernel/ArrayKernelTaskImpl.cuh#L722-736

stephenswat: Although you can't nicely fold initializer lists like you can template packs :cry:

stephenswat: > I think the incantation is

This makes the sizes compile-time variables though, which is not what we want here.

stephenswat: This somewhat recent C++ proposal leads me to believe that this is, currently, impossible: http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p1219r0.html

krasznaa: :thinking: That's true...

Then... and hear me out... Why not just use an array?

```c++
template <typename... Ts>
std::tuple<vecmem::unique_alloc_ptr<char[]>, std::add_pointer_t<Ts>...>
aligned_multiple_placement(vecmem::memory_resource &r, const std::array<std::size_t, sizeof...(Ts)>& ps) {
```

I **think** this could be made to work...

stephenswat: Hmm, that would work, but I don't see the advantage of that over an initializer list, and I don't really see the advantage of the initializer over the template pack, to be completely honest.

The advantage of the template pack being that we get to use this cool fold expression :smile::

```cpp
    const std::size_t bytes =
        ((ps * sizeof(Ts)) + ...) + sizeof...(Ts) * alignment;
```

krasznaa: Hmm... fine... You would indeed have to introduce yet another helper function to do this with an array...

Fine, let's go with this interface then...

core/include/vecmem/containers/details/aligned_multiple_placement.hpp
```diff
@@ -0,0 +1,95 @@
...
+ * freed), and the other values represent the addresses in memory where the
+ * regions begin.
+ */
+template <typename... Ts, typename... Ps>
+std::tuple<vecmem::unique_alloc_ptr<char[]>, std::add_pointer_t<Ts>...>
+aligned_multiple_placement(vecmem::memory_resource &r, Ps... ps);
+}  // namespace details
+}  // namespace vecmem
+
+#include "../impl/aligned_multiple_placement.ipp"
```
krasznaa, line 95: Let's use an absolute path here as well, since you buckled in the other places. :smile:

krasznaa (changes requested):
Just one more quibble, and then I'll stop. :stuck_out_tongue:

core/include/vecmem/containers/details/aligned_multiple_placement.hpp
```diff
@@ -0,0 +1,95 @@
...
+ *     aligned_multiple_placement<int, long, short>(r, 3, 2, 7)
+ *
+ * Under these rules, the function would allocate a chunk of memory that looks
+ * like this:
+ *
+ *      ┌────────────────────────────────────────────────────────────────┐
+ * Byte │╭ 00      ╭ 10      ╭ 20      ╭ 30      ╭ 40      ╭ 50      ╭ 60│
+ * Data │-----###IIIIIIIIIIII####LLLLLLLLLLLLLLLLSSSSSSSSSSSSSS###-------│
+ * Info │╰ A  ╰ B╰ C             ╰ D             ╰ E             ╰ F     │
+ *      └────────────────────────────────────────────────────────────────┘
```
paulgessinger, line 42: 😍

stephenswat: I knew you would like this :partying_face:

core/include/vecmem/containers/details/aligned_multiple_placement.hpp
stephenswat: Fixed!

krasznaa (approved):
Let's do this!

PR #162, opened 2022-01-28
SYCL Improvements, main branch (2022.01.28.)

krasznaa:
These have both bothered me a number of times already, so I finally wanted to give a proper solution to them.

First off, `vecmem::sycl_polyfill` needed to be switched to position independent code. Otherwise `vecmem::sycl` can not be linked successfully to shared libraries, in the cases where the polyfill is needed.

The more complicated issue was that with recent [intel/llvm](https://github.com/intel/llvm) nightlies, when building code for the CUDA or HIP backend, such warnings would be printed by the compiler due to issues in the compiler itself:

```
warning: linking module '/home/krasznaa/software/intel/clang/nightly-20220115/x86_64-ubuntu2004-gcc9-opt/lib/clang/14.0.0/../../clc/remangled-l64-signed_char.libspirv-nvptx64--nvidiacl.bc': Linking two modules of different target triples: '/home/krasznaa/software/intel/clang/nightly-20220115/x86_64-ubuntu2004-gcc9-opt/lib/clang/14.0.0/../../clc/remangled-l64-signed_char.libspirv-nvptx64--nvidiacl.bc' is 'nvptx64-unknown-nvidiacl' whereas '/data/ssd-1tb/projects/vecmem/vecmem/sycl/src/utils/sycl/device_selector.sycl' is 'nvptx64-nvidia-cuda'
 [-Wlinker-warnings]
```

So now I turned off the use of `-Werror` when even an empty `.sycl` file would not manage to compile successfully with it being used. I was considering adding `-Wno-linker-warnings` in some cases, but this seemed more robust. (Hopefully the Intel code will stop issuing these annoying warnings in not too long...)

### DIFF
--- a/cmake/vecmem-check-sycl-code-compiles.cmake
+++ b/cmake/vecmem-check-sycl-code-compiles.cmake
@@ -1,12 +1,21 @@
 # VecMem project, part of the ACTS project (R&D line)
 #
-# (c) 2021 CERN for the benefit of the ACTS project
+# (c) 2021-2022 CERN for the benefit of the ACTS project
 #
 # Mozilla Public License Version 2.0
+
+# CMake version requirement.
+cmake_minimum_required( VERSION 3.14 )
+
+# CMake include(s).
+include( CMakeParseArguments )
 
 # Helper function for checking if some SYCL files can be built into an
 # executable.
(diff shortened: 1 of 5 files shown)


krasznaa:
Let me just go ahead with this one...

PR #163, opened 2022-02-02
SYCL Improvements, main branch (2022.02.02.)

krasznaa:
Added a setting for the SYCL compiler's implicit directories/libraries. Thereby avoiding having cmake specify `-isystem /usr/include`, which breaks Clang. (And not just the Intel compiler.)

This fixes the issues described in https://github.com/intel/llvm/issues/4940 and in https://github.com/acts-project/traccc/pull/112. I finally had to figure out a proper fix in order to be able to set up a CI job for [traccc](https://github.com/acts-project/traccc), since it would be very inconvenient to pick up Boost from anything other than `/usr` in the CI. (The fixes are simple in the end, but finding the right incantation took some digging...)

While at it, I switched to using the `.obj` extension for object files on non-UNIX systems. (Cough... Windows... cough...)

Pinging @SylvainJoube and @konradkusiak97.

### DIFF
--- a/cmake/sycl/CMakeDetermineSYCLCompiler.cmake
+++ b/cmake/sycl/CMakeDetermineSYCLCompiler.cmake
@@ -1,6 +1,6 @@
 # VecMem project, part of the ACTS project (R&D line)
 #
-# (c) 2021 CERN for the benefit of the ACTS project
+# (c) 2021-2022 CERN for the benefit of the ACTS project
 #
 # Mozilla Public License Version 2.0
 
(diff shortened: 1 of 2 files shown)


SylvainJoube:
Thank you very much ! I will try it as soon as possible.

SylvainJoube:
It works perfectly with the `DPC++ Compiler 2021.1 (2020.10.0.1113)` on my local computer, thank you so much !

It does not seem to work as well with hipSYCL on a remote server, but it probably has to do with the old version of hipSYCL and CUDA (10.1) installed. I personally don't need it to work on this server now anyways.

PR #164, opened 2022-02-04
LLVM CUDA Compatibility, main branch (2022.02.04.)

krasznaa:
Triggered by @beomki-yeo's test, I updated the code to only set the nvcc specific flag if nvcc is being used.

I was quite surprised myself how easy it was to get the code to work with `clang++` like this. :open_mouth:


beomki-yeo (approved):

beomki-yeo:
once it is merged, new tag please so I can test it on traccc

PR #168, opened 2022-02-24
sycl::atomic_ref, main branch (2022.02.24.)

krasznaa:
When `sycl::atomic_ref` is available, `vecmem::atomic` is now set up simply as a thin layer above that class.

Since oneAPI only added support for `sycl::atomic_ref` relatively recently, the CMake configuration needs to check whether that type is available with the particular SYCL compiler that we use, or not. If not, we stick to the code that we had so far. If it **is** available, then `vecmem::atomic` just inherits from `sycl::atomic_ref`. ("Making use" of `sycl::atomic_ref` inside of `vecmem::atomic` turned out way more expensive LoC-wise.)

Note that while writing this PR, I had to realise that a larger change should actually be done. I'm thinking of renaming `vecmem::atomic` to `vecmem::atomic_ref`, and making it use `sycl::atomic_ref` and `cuda::atomic_ref` (https://nvidia.github.io/libcudacxx/) underneath as appropriate. And even [std::atomic_ref](https://en.cppreference.com/w/cpp/atomic/atomic_ref) once we switch to C\+\+20 with these projects. :wink: (This will require updating all users of this type a little bit.)

### DIFF
--- a/core/CMakeLists.txt
+++ b/core/CMakeLists.txt
@@ -168,7 +168,7 @@ if(NOT VECMEM_HAVE_DEFAULT_RESOURCE)
       FILES "src/memory/default_resource_polyfill.cpp" )
 endif()
 
-# Figure out how to produce SYCL debug printouts.
+# Figure out how to use various SYCL features.
 if( VECMEM_BUILD_SYCL_LIBRARY )
 
    # Test which printf function(s) is/are available.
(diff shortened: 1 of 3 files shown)


paulgessinger (approved):
Looks good as far as I can tell.

krasznaa:
We may also want to extend the list of CI jobs at one point. :thinking: Since as I was absolutely convinced it would, the CI job is not able to use `sycl::atomic_ref`.

```
...
-- Performing Test VECMEM_HAVE_SYCL_ATOMIC_REF
-- Performing Test VECMEM_HAVE_SYCL_ATOMIC_REF - Failed
...
```

https://github.com/acts-project/vecmem/runs/5319257866?check_suite_focus=true

PR #170, opened 2022-03-09
CI Updates, main branch (2022.03.09.)

krasznaa:
This is a collection of a number of updates. We recently added some more Docker images to [acts-project/machines](https://github.com/acts-project/machines) with @paulgessinger, which I now wanted to finally make use of. So I basically only wanted to update `build.yml`. Every other change in the PR is only there to support those updates. :stuck_out_tongue:

I added additional build jobs with the CUDA and HIP backends of Intel's compiler, to explicitly test those in the CI finally as well. The HIP backend unfortunately does not work in Debug mode at the moment, so the configuration needs to be a bit elaborate for that.

I had to update to the very latest version of [Google Benchmark](https://github.com/google/benchmark) to avoid warnings coming from the oneAPI C\+\+ compiler to make the build fail.

This also highlighted that the setup of the compiler flags for the benchmarks included in the repository was not quite perfect. Though that was not why the oneAPI compiler would receive the `-Werror` flag... (Still, I decided to clean that up as part of this PR.)

Finally, I removed the explicit setting for the build system generator on `windows-latest`. This should make the project more robust against future updates of GitHub.

### DIFF
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -69,16 +69,27 @@ jobs:
         PLATFORM:
           - NAME: "CUDA"
             CONTAINER: "ghcr.io/acts-project/ubuntu1804_cuda:v11"
+            OPTIONS:
           - NAME: "CUDA"
             CONTAINER: "ghcr.io/acts-project/ubuntu2004_cuda:v13"
+            OPTIONS:
           - NAME: "HIP"
             CONTAINER: "ghcr.io/acts-project/ubuntu1804_rocm:v11"
+            OPTIONS:
           - NAME: "SYCL"
             CONTAINER: "ghcr.io/acts-project/ubuntu2004_oneapi:v20"
+            OPTIONS:
           - NAME: "SYCL"
             CONTAINER: "ghcr.io/acts-project/ubuntu1804_cuda_oneapi:v20"
-          - NAME: "SYCL"
-            CONTAINER: "ghcr.io/acts-project/ubuntu1804_rocm_oneapi:v20"
+            OPTIONS: -DVECMEM_BUILD_CUDA_LIBRARY=FALSE
+        include:
+          - BUILD:
+              TYPE: "Release"
+              MSG_LVL: 0
+            PLATFORM:
+              NAME: "SYCL"
+              CONTAINER: "ghcr.io/acts-project/ubuntu1804_rocm_oneapi:v20"
+              OPTIONS: -DVECMEM_BUILD_HIP_LIBRARY=FALSE
 
     # The system to run on.
     runs-on: ubuntu-latest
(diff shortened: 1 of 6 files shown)


.github/workflows/builds.yml
```diff
@@ -26,11 +26,11 @@ jobs:
             MSG_LVL: 5
         PLATFORM:
           - OS: "ubuntu-latest"
-            GENERATOR: "Unix Makefiles"
+            GENERATOR: -G "Unix Makefiles"
           - OS: "macos-latest"
-            GENERATOR: "Xcode"
+            GENERATOR: -G "Xcode"
```
paulgessinger, line 31: I think this is going away. 

Any reason for not using ninja on all platforms?

benchmarks/core/CMakeLists.txt
```diff
@@ -1,9 +1,12 @@
 # VecMem project, part of the ACTS project (R&D line)
 #
-# (c) 2021 CERN for the benefit of the ACTS project
+# (c) 2021-2022 CERN for the benefit of the ACTS project
 #
 # Mozilla Public License Version 2.0
 
+# Project include(s).
+include( vecmem-compiler-options-cpp )
```
paulgessinger, line 8: Not sure I fully get why this has to move but ok.

.github/workflows/builds.yml
```diff
@@ -69,12 +69,27 @@ jobs:
...
+            CONTAINER: "ghcr.io/acts-project/ubuntu1804_cuda_oneapi:v20"
+            OPTIONS: -DVECMEM_BUILD_CUDA_LIBRARY=FALSE
+        include:
+          - BUILD:
+              TYPE: "Release"
+              MSG_LVL: 0
+            PLATFORM:
+              NAME: "SYCL"
+              CONTAINER: "ghcr.io/acts-project/ubuntu1804_rocm_oneapi:v20"
+              OPTIONS: -DVECMEM_BUILD_HIP_LIBRARY=FALSE
```
paulgessinger, line 92: Maybe I don't get the structure but the upper job does not build the CUDA lib but uses the CUDA oneAPI image, the lower one uses the rocm oneAPI image but goes not build the HIP library. 

I'm sure this makes sense, but why?

.github/workflows/builds.yml
krasznaa: I find that it tests the build configuration better if you use it with as many different generators as possible. A configuration that works with Ninja may very well not work with MSBuild, and vice versa.

On the `macos-latest` platform the "unique generator" available is Xcode. So I thought this would be a good way to make sure that this would continue working.

We could switch some of the builds to Ninja if you'd like. For instance all of the optimised container builds could be done with Ninja, while the debug builds would be done with GNU Make. 🤔

benchmarks/core/CMakeLists.txt
krasznaa: Previously, when it was in `benchmarks/CMakeLists.txt`, the settings coming from this vecmem specific file were becoming visible to the build configuration of Google Benchmark as well. That is what I wanted to avoid. So that the external's build would be as un-modified as possible.

.github/workflows/builds.yml
krasznaa: For the "container builds" we only run unit tests for the SYCL builds. Because the CUDA and HIP tests both need NVIDIA and AMD GPUs to be available to run correctly.

By default these rocm_oneapi and cuda_oneapi packages make the project build both HIP/SYCL and CUDA/SYCL libraries/tests. At which point the CI would try to run the HIP and CUDA unit tests in them.

So it's not that the build for these would be harmful to enable. It's that I didn't want to try to specify that everything should be built in these images, but only the non-CUDA / non-HIP tests should be run.

.github/workflows/builds.yml
paulgessinger: Actually, scratch that. CMake apparently implemented support for Xcode 12+'s "new" build system. So let's stick to Xcode.

.github/workflows/builds.yml
paulgessinger: Ok I see.

paulgessinger (approved):

PR #171, opened 2022-03-29
Correctly import cassert.h in unique obj deleter

stephenswat:
A simple fix for a simple problem: I was running into some error where `assert` was an undefined name, and this turned out to be because one of the vecmem headers did not import it correctly. This commit fixes that.


krasznaa (approved):
Of course this is good...

PR #179, opened 2022-06-02
Add virtual destructors to copy classes

konradkusiak97:
I noticed during my SYCL developments in traccc that the virtual destructors were missing from copy classes. So here they are :smiley:

### DIFF
--- a/core/include/vecmem/utils/copy.hpp
+++ b/core/include/vecmem/utils/copy.hpp
@@ -62,6 +62,9 @@ public:
     /// @name 1-dimensional vector data handling functions
     /// @{
 
+    /// Virtual destructor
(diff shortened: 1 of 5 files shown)


krasznaa (changes requested):
It's enough to declare a virtual destructor for `vecmem::copy`. We don't need to explicitly declare the destructors for all of the derived classes.

It's sort of a question of coding convention how people handle this. But in ATLAS code in general we don't declare empty virtual destructors at every level of the inheritance tree, only in the base class.

But to make it clear, the change in `vecmem::copy` is exactly what I had in mind. We should just remove all of the other changes.

krasznaa (approved):
:+1:

Issue #182, opened 2022-06-27
Using binary page memory resource in Traccc for IO takes more time than expected. 


Chamodya-ka:
Using binary page memory for file IO and algorithm IO in Traccc consumes more time than without using any downstream memory resource, this is contradicting to what is expected. The times get worse after each event. Provided below are file IO times for `traccc_seq_example_cuda   ` computing 10 events using managed memory resource and binary page memory resource using managed memory as upstream.

`Managed memory resource`
```

$ build/bin/traccc_seq_example_cuda --detector_file=tml_detector/trackml-detector.csv --digitization_config_file=tml_detector/default-geometric-config-generic.json --cell_directory=tml_full/ttbar_mu200/ --events=10 --input-binary --run_cpu=0
Running build/bin/traccc_seq_example_cuda tml_detector/trackml-detector.csv tml_full/ttbar_mu200/ 10
0 File_IO: 0.127675
1 File_IO: 0.0567025
2 File_IO: 0.0524904
3 File_IO: 0.0517798
4 File_IO: 0.0586979
5 File_IO: 0.0564024
6 File_IO: 0.0545458
7 File_IO: 0.0526243
8 File_IO: 0.0541083
9 File_IO: 0.0492084
```
`Binary page memory resource`
```
$ build/bin/traccc_seq_example_cuda --detector_file=tml_detector/trackml-detector.csv --digitization_config_file=tml_detector/default-geometric-config-generic.json --cell_directory=tml_full/ttbar_mu200/ --events=10 --input-binary --run_cpu=0
Running build/bin/traccc_seq_example_cuda tml_detector/trackml-detector.csv tml_full/ttbar_mu200/ 10
0 File_IO: 0.849157
1 File_IO: 1.45001
2 File_IO: 1.58589
3 File_IO: 1.69902
4 File_IO: 1.74604
5 File_IO: 1.86683
6 File_IO: 1.9433
7 File_IO: 2.03939
8 File_IO: 2.17419
9 File_IO: 2.15853
```
In addition, when using binary page memory resource with host memory resource as upstream for traccc_seq_example (cpu) algorithm IO takes a really long time. Below are 2 events
`host memory resource `
```
$ build/bin/traccc_seq_example --detector_file=tml_detector/trackml-detector.csv --digitization_config_file=tml_detector/default-geometric-config-generic.json --cell_directory=tml_full/ttbar_mu200/ --events=2 --input-binary 
[...]

PR #185, opened 2022-07-04
Add additional benchmarks for buddy allocator

stephenswat:
This commit adds more benchmarks to our benchmarking suite, in particular those designed to benchmark the buddy allocator. These are the benchmarks I have used to debug #182.

### DIFF
--- a/benchmarks/cuda/benchmark_cuda.cpp
+++ b/benchmarks/cuda/benchmark_cuda.cpp
@@ -6,12 +6,15 @@
  */
 
 // VecMem include(s).
+#include <vecmem/memory/binary_page_memory_resource.hpp>
 #include <vecmem/memory/cuda/device_memory_resource.hpp>
 #include <vecmem/memory/cuda/host_memory_resource.hpp>
 #include <vecmem/memory/cuda/managed_memory_resource.hpp>
 
 // Google benchmark include(s).
 #include <benchmark/benchmark.h>
+
+#include <vector>
 
 static vecmem::cuda::device_memory_resource device_mr;
 void BenchmarkCudaDevice(benchmark::State& state) {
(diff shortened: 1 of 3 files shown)


benchmarks/cuda/benchmark_cuda.cpp
```diff
@@ -6,13 +6,17 @@
  */
 
 // VecMem include(s).
+#include <vecmem/memory/binary_page_memory_resource.hpp>
```
krasznaa, line 9: It's a very petty thing, but now that the benchmark directly depends on `vecmem::core`, please list that library explicitly in:

https://github.com/acts-project/vecmem/blob/main/benchmarks/cuda/CMakeLists.txt#L19

Of course in the current setup, since `vecmem::cuda` publicly depends on `vecmem::core`, the build succeeds just fine. This is just one of the things that I'm sensitive to with my build system hat on. :stuck_out_tongue:

benchmarks/cuda/benchmark_cuda.cpp
```diff
@@ -6,13 +6,17 @@
...
 #include <vecmem/memory/cuda/device_memory_resource.hpp>
 #include <vecmem/memory/cuda/host_memory_resource.hpp>
 #include <vecmem/memory/cuda/managed_memory_resource.hpp>
 
 // Google benchmark include(s).
 #include <benchmark/benchmark.h>
 
+#include <iostream>
```
krasznaa, line 17: I guess this was left here by mistake. :stuck_out_tongue:

krasznaa (commented):
Of course I'm all on board with the new benchmarks, I just had to leave some properly anal comments...

benchmarks/cuda/benchmark_cuda.cpp
stephenswat: Fixed, thanks. :slightly_smiling_face:

benchmarks/cuda/benchmark_cuda.cpp
stephenswat: Indeed it was, fixed.

krasznaa (approved):
Looks good.

PR #186, opened 2022-07-04
Fix missing include in `unique_alloc_deleter.hpp`

stephenswat:
This commit fixes a missing inclusion in the `unique_alloc_deleter.hpp` file, which had not previously manifested itself because the include (`cassert`) was being caught by another header file.


krasznaa (approved):
Yepp.

PR #187, opened 2022-07-04
Accelerate buddy allocator internals

stephenswat:
As we have seen in #182, the buddy allocator we have can sometimes be pretty slow. This was not due to an issue with the fundamental design of the allocator, but rather with its internal structure and implementation. This commit updates the allocator to use a contiguous array of pages, which simplifies a lot of operations and reduces the need for pointer chasing, and significantly improves the performance of the allocator.

Relies on #185 and #186.

### DIFF
--- a/core/src/memory/binary_page_memory_resource_impl.cpp
+++ b/core/src/memory/binary_page_memory_resource_impl.cpp
@@ -17,7 +17,7 @@
 #include <optional>
 #include <stdexcept>
 
-#ifdef _MSC_VER
+#ifdef VECMEM_HAVE_LZCNT_U64
 #include <intrin.h>
 #endif
 
@@ -36,11 +36,19 @@ std::size_t round_up(std::size_t size) {
     return 0;
 }
 
-std::size_t __inline clzl(std::size_t i) {
-#ifdef _MSC_VER
+inline std::size_t clzl(std::size_t i) {
+#if defined(VECMEM_HAVE_LZCNT_U64)
     return _lzcnt_u64(i);
-#else
+#elif defined(VECMEM_HAVE_BUILTIN_CLZL)
     return __builtin_clzl(i);
+#else
+    std::size_t b;
+    for (b = 0;
+         !((i << b) & (static_cast<std::size_t>(1UL)
+                       << (std::numeric_limits<std::size_t>::digits - 1UL)));
+         ++b)
+        ;
+    return b;
 #endif
 }
 }  // namespace
--- a/core/src/memory/binary_page_memory_resource_impl.cpp
+++ b/core/src/memory/binary_page_memory_resource_impl.cpp
@@ -14,23 +14,34 @@
 // System include(s).
 #include <algorithm>
 #include <memory>
-#include <stack>
+#include <optional>
 #include <stdexcept>
+
+#ifdef _MSC_VER
+#include <intrin.h>
+#endif
 
 namespace {
 /**
- * @brief Rounds a size up to the nearest power of two.
+ * @brief Rounds a size up to the nearest power of two, and returns the power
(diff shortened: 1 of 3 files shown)


core/src/memory/binary_page_memory_resource_impl.cpp
```diff
@@ -14,24 +14,35 @@
...
     return 0;
 }
+
+std::size_t __inline clzl(std::size_t i) {
+#ifdef _MSC_VER
+    return _lzcnt_u64(i);
+#else
+    return __builtin_clzl(i);
+#endif
+}
```
krasznaa, line 45: As discussed in person, let me give an alternative implementation here.

I would much prefer for CMake to make the decision on which function the code should use internally. With something like:

```cmake
include( CheckCXXSymbolExists )
check_cxx_symbol_exists( "__lzcnt_u64" "intrin.h" VECMEM_HAVE_LZCNT_U64 )
if( VECMEM_HAVE_LZCNT_U64 )
   target_compile_definitions( vecmem_core
      PRIVATE VECMEM_HAVE_LZCNT_U64 )
endif()
check_cxx_symbol_exists( "__builtin_clzl" "" VECMEM_HAVE_BUILTIN_CLZL )
if( VECMEM_HAVE_BUILTIN_CLZL )
   target_compile_definitions( vecmem_core
      PRIVATE VECMEM_HAVE_BUILTIN_CLZL )
endif()
```

And then the code could look like:

```c++
std::size_t clzl(std::size_t i) {
#if defined(VECMEM_HAVE_LZCNT_U64)
    return _lzcnt_u64(i);
#elif defined(VECMEM_HAVE_BUILTIN_CLZL)
    return __builtin_clzl(i);
#else
    // Hand-written implementation for the function...
#endif
}
```

Note that I have some reservations against `__inline` as well. :thinking: Can't we just use `inline` instead?

https://en.cppreference.com/w/cpp/language/inline

If not, then we'll need to figure out with CMake also whether `__inline` is available from the compiler being used...

stephenswat:
CMake is such a fucking parody of a build system. :unamused:

stephenswat:
So apparently `check_cxx_source_compiles`'s default behaviour is to print a log message when it succeeds, but to be completely quiet if it fails.

Because yes, that makes a lot of sense. :+1:

stephenswat:
Should be fixed on Mac now.

krasznaa:
Yeah, I should've thought about this sooner... 😦 The same thing was hitting me back in the day with https://github.com/root-project/root/pull/4561/files back in the day...

stephenswat:
Well, it's not your responsibility to remember all these things. Just would be nice if they were a little easier to debug. :smile:

core/src/memory/binary_page_memory_resource_impl.cpp
```diff
@@ -14,24 +14,43 @@
 // System include(s).
 #include <algorithm>
 #include <memory>
-#include <stack>
+#include <optional>
 #include <stdexcept>
 
+#ifdef _MSC_VER
```
krasznaa, line 20: I would modify this to using `VECMEM_HAVE_LZCNT_U64` as well. Since that test explicitly checks the existence of this header file as well...

krasznaa (changes requested):
Just one more complaint...

core/src/memory/binary_page_memory_resource_impl.cpp
stephenswat: Fixed!

krasznaa (approved):
Okay, let's go ahead with it finally.

If I didn't mention it before, I like the use of [std::optional](https://en.cppreference.com/w/cpp/utility/optional) for the code's internals. :wink:

Issue #183, opened 2022-06-28
Alignment support should be added to all memory resources


stephenswat:
Following #181, this is a long-term goal we should complete; we have a lot of memory resources which are not current able to support alignment.

krasznaa:
I believe this was done a while ago... 🤔

PR #191, opened 2022-07-14
Aligned SYCL Memory Allocations, main branch (2022.07.14.)

krasznaa:
Switched to using aligned memory allocations in the SYCL memory resources.

While in practice it should not make a difference for our current code, since SYCL provides a version of its memory allocation functions with custom alignments, we should use those instead of the generic allocation functions.

Note that these functions are part of the [SYCL2020](https://www.khronos.org/registry/SYCL/specs/sycl-2020/pdf/sycl-2020.pdf) standard. But that is already required in our code for un-aligned allocations anyway. (At one point we'll need to teach the build system to require the support for SYCL2020 during the configuration...)

This is probably the simplest part of implementing #183. Note that CUDA and HIP do not provide support for custom alignment values in their memory allocation functions. But since in practice all memory allocations are `0x400` aligned in CUDA, that should not cause any actual problems any time soon...

### DIFF
--- a/sycl/src/memory/sycl/device_memory_resource.sycl
+++ b/sycl/src/memory/sycl/device_memory_resource.sycl
@@ -1,6 +1,6 @@
 /** VecMem project, part of the ACTS project (R&D line)
  *
- * (c) 2021 CERN for the benefit of the ACTS project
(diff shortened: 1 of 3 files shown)


stephenswat (approved):
LGTM.

PR #192, opened 2022-07-14
Aligned Host Memory Allocations, main branch (2022.07.14.)

krasznaa:
Switched `vecmem::host_memory_resource` to using [std::aligned_alloc](https://en.cppreference.com/w/cpp/memory/c/aligned_alloc). This is a follow-up from #191, providing yet another component for #183.

At the same time updated the documentation and debug messages of the code a little.

### DIFF
--- a/core/src/memory/host_memory_resource.cpp
+++ b/core/src/memory/host_memory_resource.cpp
@@ -14,6 +14,7 @@
 // System include(s).
 #include <algorithm>
 #include <cassert>
+#include <cstddef>
 #include <cstdlib>
 #include <memory>
 
@@ -30,13 +31,21 @@ void *host_memory_resource::do_allocate(std::size_t bytes,
     const std::size_t align = std::max(alignment, alignof(std::max_align_t));
     const std::size_t size = bytes + (align - (bytes % align));
     void *ptr = std::aligned_alloc(align, size);
+    // If the alignment failed for some reason, throw an exception.
+    if (ptr == nullptr) {
+        throw std::bad_alloc();
+    }
 #else
-    // Ask for enough memory that will allow us to be able to align the
+    // Ask for enough memory that will allow us to be able to align
     // the returned pointer for sure, **and** be able to store the
     // pointer to the beginning of the unaligned memory blob before
     // the aligned pointer.
     std::size_t alloc_bytes = bytes + alignment + sizeof(void *) - 1;
     void *unaligned_ptr = std::malloc(alloc_bytes);
+    // If the alignment failed for some reason, throw an exception.
+    if (unaligned_ptr == nullptr) {
+        throw std::bad_alloc();
+    }
     // Make a pointer that has space "before it" for the
     // pointer to the beginning of the unaligned space.
     void *ptr = static_cast<char *>(unaligned_ptr) + sizeof(void *);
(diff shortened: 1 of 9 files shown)


core/src/memory/host_memory_resource.cpp
```diff
@@ -12,21 +12,60 @@
...
+    // pointer to the beginning of the unaligned memory blob before
+    // the aligned pointer.
+    std::size_t alloc_bytes = bytes + alignment + sizeof(void *) - 1;
+    void *unaligned_ptr = std::malloc(alloc_bytes);
+    // Get an aligned pointer that has space "before it" for the
+    // pointer to the beginning of the unaligned space.
+    void *ptr = static_cast<char *>(unaligned_ptr) + sizeof(void *);
+    std::align(alignment, bytes, ptr, alloc_bytes);
+    // Store the unaligned pointer's value just before the aligned
+    // pointer for later use.
```
stephenswat, line 45: I would strongly suggest we add an assertion here to make sure this call succeeds. If I remember correctly, this function returns a `nullptr` if there was insufficient space, which we should check for.

core/src/memory/host_memory_resource.cpp
```diff
@@ -12,21 +12,60 @@
...
+    // Ask for enough memory that will allow us to be able to align the
+    // the returned pointer for sure, **and** be able to store the
+    // pointer to the beginning of the unaligned memory blob before
+    // the aligned pointer.
+    std::size_t alloc_bytes = bytes + alignment + sizeof(void *) - 1;
+    void *unaligned_ptr = std::malloc(alloc_bytes);
+    // Get an aligned pointer that has space "before it" for the
+    // pointer to the beginning of the unaligned space.
+    void *ptr = static_cast<char *>(unaligned_ptr) + sizeof(void *);
+    std::align(alignment, bytes, ptr, alloc_bytes);
```
stephenswat, line 43: Note that the size `alloc_bytes` has shrunk by the size of a void pointer here, which we should account for so that `std::align` can properly sanity check the call.

core/src/memory/host_memory_resource.cpp
krasznaa: You're okay with my updates, right...? (I completely forgot about this PR...)

core/src/memory/host_memory_resource.cpp
```diff
@@ -12,21 +12,68 @@
...
+    // pointer to the beginning of the unaligned space.
+    void *ptr = static_cast<char *>(unaligned_ptr) + sizeof(void *);
+    // The bytes available for alignment are a little less than what we
+    // allocated.
+    std::size_t align_bytes = alloc_bytes - sizeof(void *);
+    // Perform the alignment.
+    if (std::align(alignment, bytes, ptr, align_bytes) == nullptr) {
+        // The alignment was meant to succeed. If it didn't, we messed
+        // something up in the code. :-/
+        return nullptr;
```
stephenswat, line 50: By the C++ standard, we should throw an exception here, not return a nullptr. We should also attempt to deallocate the memory we got from the system (`unaligned_ptr`) in this case.

stephenswat (changes requested):

core/src/memory/host_memory_resource.cpp
krasznaa: I went for an assertion...

stephenswat (approved):

PR #199, opened 2022-10-17
Resizable Jagged Vector Copies, main branch (2022.10.17.)

### LINKED ISSUE #95: Resizable (jagged) vectors can't have a non-zero starting size
As I mentioned in #94, this has been bothering me for a while... With a constructor like

```c++
    /// Resizable data constructor
    vector_buffer(size_type capacity, size_type size,
                  memory_resource& resource);
```

, one should assume that the created vector buffer would have a starting size of `size`. But in the current implementation it doesn't. It always starts from 0. :frowning:

When you set up a resizable (jagged) vector currently, you do it like:

```c++
    vecmem::data::vector_buffer<int> output_buffer(input.size(), 0,
                                                   device_resource);
    m_copy.setup(output_buffer);
```

Where `m_copy` is one of the "platform specific" copy objects. The problem here is that the buffer constructor currently has no way of storing the starting size that it received, in the memory that it would allocate with the memory resource that it receives. In case the memory resource allocates non-host-accessible memory, one needs to write to that memory using the appropriate "copy class". But by the time that we use the `vecmem::copy::setup(...)` function, that original size variable is lost. (The buffer does not store it anywhere.)

There are a couple of ways out of it, but they either require some bigger re-organisation in the code, or a setup that I wouldn't really like. I think the following are all our options here:
  - We add one more variable to the buffer classes, which would hold the "starting size" of the buffer. This variable would then be used by `vecmem::copy::setup(...)` when initialising the (device) memory. Technically this could work (I think...), but I just really don't like this design. (But maybe you guys could convince me otherwise...)
  - In order to create a resizable (jagged) vector buffer, one would need to provide both a memory resource, and a copy object to its constructor. So that the constructor itself would take care of all of the setup that it needs. I'm leaning towards this option at the moment... (However we might run into problems with circular dependencies on this one, since `vecmem::copy` already knows about the buffer types...)
  - We create "factory functions" on `vecmem::copy` for constructing resizable (jagged) vector buffers. This is **very** similar to option number 2, and would introduce a design that we don't use anywhere else in the project so far.

So... Any objections to having to create resizable buffers with code like:

```c++
vecmem::cuda::device_memory_resource resource;
vecmem::cuda::copy copy;

vecmem::data::vector_buffer<int> buffer1(1000, 100, resource, copy);
vecmem::data::jagged_vector_buffer<int> buffer2({1000, 1000, 1000}, {10, 100, 200}, resource, copy);
```

?

@konradkusiak97, I'm interested in your opinion as well. :wink:

krasznaa:
Continuing with hunting down bugs found by @guilhermeAlmeida1, I now made copies from resizable jagged vectors possible.

The previously implemented code broke down if the source of the copy was a resizable `jagged_vector_buffer`, which was not filled with data completely. Which is exactly what happens during "doublet counting" in [traccc](https://github.com/acts-project/traccc). The good thing is that the assertions in the code were correct. They did catch this logic error. But only when building the code in "Debug" mode. :stuck_out_tongue:

In the updated setup the copies generally use the vector capacities instead of the sizes, in order to have a chance for optimised copies out of resizable `jagged_vector_buffer`-s.

At the same time I introduced the `vecmem::copy::set_sizes(...)` function for conveniently setting the sizes of a jagged vector (buffer). Which can come in handy when it's the target container that is resizable, and not the source one. (Though this latter operation I didn't really test yet.) And I also plan to use this new function to finally resolve #95 in the non-too-distant future.

@guilhermeAlmeida1, please check if this branch works correctly with your code. I will still need to add some meaningful unit tests to this PR before I would push it in. Since unfortunately the copies are woefully under-tested in this repository right now. :frowning:

### DIFF
--- a/core/include/vecmem/utils/copy.hpp
+++ b/core/include/vecmem/utils/copy.hpp
@@ -133,6 +133,12 @@ public:
     std::vector<typename data::vector_view<TYPE>::size_type> get_sizes(
         const data::jagged_vector_view<TYPE>& data);
 
+    /// Helper function for setting the sizes of a resizable jagged vector
+    template <typename TYPE>
+    void set_sizes(
+        const std::vector<typename data::vector_view<TYPE>::size_type>& sizes,
+        data::jagged_vector_view<TYPE> data);
+
     /// @}
 
 protected:
@@ -145,8 +151,10 @@ protected:
 private:
     /// Helper function performing the copy of a jagged array/vector
     template <typename TYPE1, typename TYPE2>
-    void copy_views_impl(std::size_t size, const data::vector_view<TYPE1>* from,
-                         data::vector_view<TYPE2>* to, type::copy_type cptype);
+    void copy_views_impl(
+        const std::vector<typename data::vector_view<TYPE1>::size_type>& sizes,
+        const data::vector_view<TYPE1>* from, data::vector_view<TYPE2>* to,
+        type::copy_type cptype);
     /// Helper function for getting the sizes of a jagged vector/buffer
     template <typename TYPE>
     std::vector<typename data::vector_view<TYPE>::size_type> get_sizes_impl(
--- a/core/include/vecmem/utils/impl/copy.ipp
+++ b/core/include/vecmem/utils/impl/copy.ipp
@@ -16,6 +16,7 @@
 // System include(s).
 #include <algorithm>
 #include <cassert>
+#include <stdexcept>
 
 namespace vecmem {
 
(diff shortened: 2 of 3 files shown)


guilhermeAlmeida1:
> @guilhermeAlmeida1, please check if this branch works correctly with your code. 

On my side, the copying seems to be working correctly now.

krasznaa:
Okay, the PR is now ready for serious review.

At the same time: I found multiple things during this development process that I'll open additional PRs for. One of those will need minor adjustments in the downstream projects as well...

krasznaa:
Just 2 things:
  - The Windows CI build/test is still invaluable. It pointed out a real issue in the testing code with uninitialised memory which none of the other tests were sensitive to (should we enable an address sanitizer build?);
  - Debugging assertions in Visual Studio is much more of a pain than I could've ever imagined. I thought it would be a convenient GUI experience. But no... :frowning:

PR #205, opened 2022-11-04
CI Update, main branch (2022.11.04.)

krasznaa:
Upgraded to [CUDA 11.6.2](https://developer.nvidia.com/cuda-11-6-2-download-archive) (https://github.com/acts-project/machines/pull/72) and [intel/llvm 2022-06](https://github.com/intel/llvm/releases/tag/2022-06) (https://github.com/acts-project/machines/pull/73).

The first one will help with upgrading to even newer versions of the Intel compiler on top of Ubuntu 20.04, and the latter will hopefully put a stop to all of the CI failures that have been plaguing us in the last few weeks.

### DIFF
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -76,7 +76,7 @@ jobs:
             CONTAINER: "ghcr.io/acts-project/ubuntu1804_cuda:v11"
             OPTIONS:
           - NAME: "CUDA"
-            CONTAINER: "ghcr.io/acts-project/ubuntu2004_cuda:v13"
+            CONTAINER: "ghcr.io/acts-project/ubuntu2004_cuda:v32"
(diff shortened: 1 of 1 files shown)


krasznaa:
Well, this is disappointing. :frowning: Apparently we'll need to give up on the `ubuntu1804_cuda_oneapi` tests all together. It seems that contrary to Intel's claims their compiler is not all that compatible with CUDA 10.2 by now after all. :frowning: (I was very much hoping that the `2022-06` tag of te Intel compiler would prove more robust, but apparently the Intel compiler version was not the issue.)

We should switch to using `ubuntu2004_cuda_oneapi` for this project's tests once that image is made. Until then let's keep this PR hanging...

krasznaa:
Closing in favour of #206.

PR #208, opened 2022-11-14
Copy and Assignment Fixes, main branch (2022.11.14.)

krasznaa:
As @guilhermeAlmeida1 found in https://github.com/acts-project/traccc/pull/271, the "data owning" types couldn't be default-constructed so far. Which there can be a legitimate use case for. (As shown in that PR.)

While adding those default constructors I noticed that copy constructors and assignment operators on the "view types" were also a bit all over the place. So I now tried to clean those up a bit as well.

Also added compatison operators for the view types.

Finally, added some extra tests at the end to make sure that all of this would work as intended.

### DIFF
--- a/core/include/vecmem/containers/data/jagged_vector_buffer.hpp
+++ b/core/include/vecmem/containers/data/jagged_vector_buffer.hpp
@@ -53,6 +53,9 @@ public:
 
     /// @}
 
+    /// Default constructor
+    jagged_vector_buffer();
+
     /// Constructor from an existing @c vecmem::data::jagged_vector_view
     ///
     /// @param other The existing @c vecmem::data::jagged_vector_view object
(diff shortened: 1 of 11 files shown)


core/include/vecmem/containers/data/jagged_vector_buffer.hpp
```diff
@@ -95,6 +98,16 @@ class jagged_vector_buffer : public jagged_vector_view<TYPE> {
                          memory_resource& resource,
                          memory_resource* host_access_resource = nullptr);
 
+    /// Copy constructor
+    jagged_vector_buffer(const jagged_vector_buffer&) = default;
```
krasznaa, line 102: Crud... I was meaning to remove these copy constructors... :frowning: (These are not really well defined for the buffer types.) Shoot...

PR #209, opened 2022-12-13
Asynchronous Copies, main branch (2022.12.13.)

krasznaa:
As I mentioned at the last parallelization meeting, the current implementation we have for jagged vector copies does not allow for asynchronous copies. I had to re-think the code a bit to allow proper asynchronous operations in [traccc](https://github.com/acts-project/traccc).

First off I removed the "hidden" optimisation from the jagged vector copy operator. It will no longer create a "host buffer" itself to minimise the H<->D memory copies. Instead the users must do this explicitly themselves. Like:

```c++
vecmem::copy host_copy;
vecmem::cuda::host_memory_resource cuda_host_mr;
vecmem::cuda::copy cuda_copy;
vecmem::jagged_vector<Foo> host_vec = ...;
auto host_buffer = host_copy.to(vecmem::get_data(host_vec, cuda_host_mr);
vecmem::data::jagged_vector_buffer<Foo> device_buffer{...};
cuda_copy(host_buffer, device_buffer);
```

I agree, it seems like a fair amount of additional code. :frowning: We'll have to see how we could optimise things a bit more later on. But in such a setup it is possible to perform the H<->D copy asynchronously. (It is now up to the user to make sure that the copy source and target would be alive for long enough.)

And to test that, I updated `vecmem::copy` to now provide an "asynchronous interface". One that would be backwards compatible with the interface that we had already. So that clients that are not interested in asynchronous operations would be able to forego asynchronous operations.

Updated `vecmem::cuda::async_copy` and introduced `vecmem::sycl::async_copy` to implement this new asynchronous interface.

To test the asynchronous implementation a little, I also updated some of the SYCL unit tests to use `vecmem::sycl::async_copy` and include explicit  synchronisation points in the right places.

There are some things about this implementation that I'm still not crazy about, but I wanted to get some feedback on this style of code organisation finally.

### DIFF
--- a/core/include/vecmem/utils/abstract_event.hpp
+++ b/core/include/vecmem/utils/abstract_event.hpp
@@ -1,0 +1,32 @@
+/*
+ * VecMem project, part of the ACTS project (R&D line)
+ *
+ * (c) 2022 CERN for the benefit of the ACTS project
+ *
+ * Mozilla Public License Version 2.0
+ */
+#pragma once
+
+namespace vecmem {
+
+/// Interface that language specific "events" need to implement
+///
+/// "Events" provide a way to have synchronization points in a program's
+/// execution flow. Asynchronous memory operations in different languages
+/// provide ways of explicitly waiting for the completion of certain operations.
+/// This abstract interface is used in this project to provide a uniform
+/// interface to language specific "event" objects that would provide such
+/// synchronization points.
+///
+struct abstract_event {
+
+    /// Virtual destructor to make vtable happy
+    virtual ~abstract_event() {}
+
+    /// Function that would block the current thread until the event is
+    /// complete
+    virtual void wait() = 0;
+
+};  // struct abstract_event
+
+}  // namespace vecmem
--- a/core/include/vecmem/utils/copy.hpp
+++ b/core/include/vecmem/utils/copy.hpp
@@ -13,10 +13,12 @@
 #include "vecmem/containers/data/vector_buffer.hpp"
 #include "vecmem/containers/data/vector_view.hpp"
 #include "vecmem/memory/memory_resource.hpp"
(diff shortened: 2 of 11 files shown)


krasznaa:
It may be worth noting that I spent some time thinking about the return types of the `vecmem::copy::to` functions. In this PR those functions remain explicitly synchronous.

I was wondering about using [std::future](https://en.cppreference.com/w/cpp/thread/future) like:

```c++
    template <typename TYPE>
    std::future<data::vector_buffer<std::remove_cv_t<TYPE>>> to(
        const data::vector_view<TYPE>& data, memory_resource& resource,
        type::copy_type cptype = type::unknown);
```

But (while not actually having tried to implement it yet) I don't think that will actually work. :frowning: If we want to make those functions asynchronous, I think we'll need to come up with a `vecmem::future<T>` type instead.

However that will be for a future PR to work out the details in.

HadrienG2:
Note that std::future is remarkably useless as a future API in that it does not allow you to do much with futures besides awaiting them one by one, which all but misses the point of asynchronous programming.

As an alternative, consider using future implementations with continuation support and collective wait operations (any, all...). I heard good things about the Lazy and Folly libraries in this respect.

stephenswat:
There is an experimental monad interface for `std::future` that resolves this: https://en.cppreference.com/w/cpp/experimental/future/then

But yes, it's experimental.

HadrienG2:
Given that Bartosz Milewski was already advocating for this addition to the C++ standard in 2014, it seems it's not on a very fast track to landing https://bartoszmilewski.com/2014/02/26/c17-i-see-a-monad-in-your-future/

Most likely C++ standard people would rather have it superseded by an overengineered monster like std::execution.

stephenswat (approved):

PR #214, opened 2023-01-17
Doxygen Namespace Fixes, main branch (2023.01.17.)

krasznaa:
Fixed the Doxygen documentation for all used namespaces.

Had to remove the `/**` formalism from the file headers, as these were confusing Doxygen. :frowning:  Making it think that those blocks were documenting the first thing (usually a namespace) defined in the file.

At the same time added explicit documentation for all of the namespaces used in the project.


stephenswat:
I would like to file a complaint about changing the headers in 122 files and then sneaking in some additional changes as well. :laughing:

stephenswat (approved):

PR #220, opened 2023-02-23
Change buffer constructor size type to template parameter.

guilhermeAlmeida1:
Would be useful to construct buffer objects with input size vectors with types other than `std::size_t`.


core/include/vecmem/containers/data/jagged_vector_buffer.hpp
```diff
@@ -79,7 +79,8 @@ class jagged_vector_buffer : public jagged_vector_view<TYPE> {
     ///        be host accessible.
     /// @param host_access_resource An optional host accessible memory
     ///        resource. Needed if @c resource is not host accessible.
-    jagged_vector_buffer(const std::vector<std::size_t>& sizes,
+    template <typename size_t = std::size_t>
```
stephenswat, line 82: Can we name this `size_type`? `size_t` is too easily confused with `std::size_t`.

core/include/vecmem/containers/impl/jagged_vector_buffer.ipp
```diff
@@ -70,8 +70,9 @@ jagged_vector_buffer<TYPE>::jagged_vector_buffer(
 }
 
 template <typename TYPE>
+template <typename size_t>
```
stephenswat, line 73: Is there a reason this cannot be a single-layer template?

core/include/vecmem/containers/impl/jagged_vector_buffer.ipp
```diff
@@ -102,10 +103,10 @@ jagged_vector_buffer<TYPE>::jagged_vector_buffer(
...
```
stephenswat, line 106: Same as above.

stephenswat (changes requested):

core/include/vecmem/containers/impl/jagged_vector_buffer.ipp
krasznaa: I myself like it better this way. The buffer can be the same whether you created it with the help of `std::vector<std::size_t>` or `std::vector<unsigned int>`. Since these types are not used **anywhere** outside of the constructor(s) anyway.

guilhermeAlmeida1: That was my first attempt but this seems not to be accepted:
```
home/gumetelo/vecmem/vecmem/core/include/vecmem/containers/impl/jagged_vector_buffer.ipp:105:1: error: too many template parameters in template redeclaration
template <typename TYPE, typename size_t>
^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/gumetelo/vecmem/vecmem/core/include/vecmem/containers/data/jagged_vector_buffer.hpp:28:1: note: previous template declaration is here
template <typename TYPE>
```

Apparently the template in the definition of the class needs to be separated from the one only for the constructor?

core/include/vecmem/containers/data/jagged_vector_buffer.hpp
```diff
@@ -93,8 +94,9 @@ class jagged_vector_buffer : public jagged_vector_view<TYPE> {
     ///        be host accessible.
     /// @param host_access_resource An optional host accessible memory
     ///        resource. Needed if @c resource is not host accessible.
-    jagged_vector_buffer(const std::vector<std::size_t>& sizes,
-                         const std::vector<std::size_t>& capacities,
+    template <typename size_t = std::size_t>
```
krasznaa, line 97: I would make this slightly more elaborate...

First off, I don't think we need a default type. The compiler should be able to deduce the type from the arguments that it receives for the constructor.

But second, it could be nice to use SFINAE to only allow integer types to be used. :wink:

core/include/vecmem/containers/impl/jagged_vector_buffer.ipp
stephenswat: Ah yes good point, that error does make sense.

core/include/vecmem/containers/data/jagged_vector_buffer.hpp
stephenswat: And only unsigned integers, please!

guilhermeAlmeida1: I'll add the checks for uint types only.
I put the default type in to make all the CI tests work as-is. e.g. 
https://github.com/acts-project/vecmem/blob/main/tests/cuda/test_cuda_jagged_vector_view.cpp#L170-L171
And I don't really see how it can harm us

krasznaa: Okay, that's a good point. Let's keep the `std::size_t` type by default then. (I'm a bit surprised that C\+\+ didn't default to an `std::vector<int>`  type in that case though. Those number literals should've defaulted to `int`... :thinking:)

stephenswat (approved):
Looks good. :+1:

krasznaa:
By SFINAE, I meant SFINAE...

core/include/vecmem/containers/impl/jagged_vector_buffer.ipp
```diff
@@ -102,9 +106,13 @@ jagged_vector_buffer<TYPE>::jagged_vector_buffer(
 }
 
 template <typename TYPE>
+template <typename SIZE_TYPE,
+          std::enable_if_t<std::is_integral<SIZE_TYPE>::value &&
+                               std::is_unsigned<SIZE_TYPE>::value,
```
guilhermeAlmeida1, line 111: The second one implies the first so there was no need for both checks though

stephenswat:
There is no possibility of overload resolution ambiguities in this code, so making this SFINAE instead of using static asserts is really just making the code more complex and the errors less readable for no good reason... :confused:

krasznaa:
We already have a number of constructors. :confused: Why couldn't we add more later on? For which SFINAE **is** the way to go.

core/include/vecmem/containers/impl/jagged_vector_buffer.ipp
krasznaa: `std::is_unsigned<bool>::value == true`

stephenswat: Just for reference, `std::is_integral<bool>::value` is also true. :wink:

krasznaa: Yeah, I realised after writing this down... :frowning: Okay, so `std::is_integral` is indeed quite useless at the moment...

stephenswat: Just to be clear, I do actually like the `std::is_integral` being there. `std::is_unsigned` is not defined as the type being a member of a predefined set of unsigned types, but rather it is defined as `std::is_unsigned<T>` iff. T(-1) > T(1).

It's rather easy to come up with a pathological unsigned floating point type that the standard library would classify as unsigned! So the integral check does have a use. :slightly_smiling_face:

PR #228, opened 2023-03-30
Add target option for atomic ref

guilhermeAlmeida1:
Current SYCL implementation of `device_atomic_ref` only allows using device-wide atomics in global space. 
This PR introduces the possibility of targetting local space instead.

### DIFF
--- a/core/include/vecmem/memory/device_atomic_ref.hpp
+++ b/core/include/vecmem/memory/device_atomic_ref.hpp
@@ -1,7 +1,7 @@
 /*
  * VecMem project, part of the ACTS project (R&D line)
  *
- * (c) 2022 CERN for the benefit of the ACTS project
+ * (c) 2022-2023 CERN for the benefit of the ACTS project
  *
  * Mozilla Public License Version 2.0
  */
@@ -9,6 +9,13 @@
 
 // System include(s).
 #include <atomic>
+
+namespace vecmem {
+
+/// Custom definition for the adress space
+enum class device_address_space { global = 0, local = 1 };
+
+}  // namespace vecmem
 
 #if (defined(CL_SYCL_LANGUAGE_VERSION) || defined(SYCL_LANGUAGE_VERSION)) && \
     defined(VECMEM_HAVE_SYCL_ATOMIC_REF)
@@ -21,12 +28,34 @@ namespace vecmem {
 /// Define @c vecmem::memory_order as @c sycl::memory_order
 using memory_order = ::sycl::memory_order;
 
+namespace details {
+template <device_address_space address>
+struct sycl_address_space {};
+
+template <>
+struct sycl_address_space<device_address_space::global> {
+    static constexpr ::sycl::memory_order ord = ::sycl::memory_order::relaxed;
+    static constexpr ::sycl::memory_scope scp = ::sycl::memory_scope::device;
+    static constexpr ::sycl::access::address_space add =
+        ::sycl::access::address_space::global_space;
+};
+template <>
+struct sycl_address_space<device_address_space::local> {
+    static constexpr ::sycl::memory_order ord = ::sycl::memory_order::relaxed;
+    static constexpr ::sycl::memory_scope scp =
+        ::sycl::memory_scope::work_group;
+    static constexpr ::sycl::access::address_space add =
+        ::sycl::access::address_space::local_space;
+};
+}  // namespace details
+
 /// @c vecmem::atomic_ref equals @c sycl::atomic_ref with "modern SYCL"
-template <typename T>
+template <typename T,
+          device_address_space address = device_address_space::global>
 using device_atomic_ref =
-    ::sycl::atomic_ref<T, ::sycl::memory_order::relaxed,
-                       ::sycl::memory_scope::device,
-                       ::sycl::access::address_space::global_space>;
+    ::sycl::atomic_ref<T, details::sycl_address_space<address>::ord,
+                       details::sycl_address_space<address>::scp,
+                       details::sycl_address_space<address>::add>;
 
 }  // namespace vecmem
 
@@ -40,7 +69,7 @@ namespace vecmem {
 using memory_order = std::memory_order;
 
 /// @c vecmem::atomic_ref equals @c std::atomic_ref in host code with C++20
-template <typename T>
+template <typename T, device_address_space = device_address_space::global>
 using device_atomic_ref = std::atomic_ref<T>;
 
 }  // namespace vecmem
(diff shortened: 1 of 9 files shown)


stephenswat (commented):
Just be aware that this creates an awkward API-incompatibility between the SYCL version of this type and the non-SYCL types, where the SYCL versions take a second `address` template parameters, while the other do not. This could be awkward in "shared code" developments.

core/include/vecmem/memory/device_atomic_ref.hpp
```diff
@@ -22,11 +29,15 @@ namespace vecmem {
...
 using device_atomic_ref =
     ::sycl::atomic_ref<T, ::sycl::memory_order::relaxed,
-                       ::sycl::memory_scope::device,
-                       ::sycl::access::address_space::global_space>;
+                       address == address_space::global
+                           ? ::sycl::memory_scope::device
+                           : ::sycl::memory_scope::work_group,
+                       address == address_space::global
+                           ? ::sycl::access::address_space::global_space
+                           : ::sycl::access::address_space::local_space>;
```
krasznaa, line 40: :thinking: Let's do it more properly. It's very likely that 2 values is what `vecmem::device_address_space` is ever going to have. But even so, let's not hard-code this assumption everywhere in our code.

Instead, let's use some "trait classes" here. Like:

```c++
template <device_address_space address>
struct sycl_address {};
template <>
struct sycl_address<device_address_space::global> {
   using memory_scope = ::sycl::memory_scope::device;
   using address_space = ::sycl::access::address_space::global_space;
};
template <>
struct sycl_address<device_address_space::local> {
   using memory_scope = ::sycl::memory_scope::work_group;
   using address_space = ::sycl::access::address_space::local_space;
};
template <typename T, device_address_space address = device_address_space::global>
using device_atomic_ref = ::sycl::atomic_ref<T, sycl_address<address>::memory_scope, sycl_address<address>::address_space>;
```

(Code of course not actually tested...)

core/include/vecmem/memory/device_atomic_ref.hpp
```diff
@@ -40,7 +51,7 @@ namespace vecmem {
 using memory_order = std::memory_order;
 
 /// @c vecmem::atomic_ref equals @c std::atomic_ref in host code with C++20
-template <typename T>
+template <typename T, address_space address = address_space::global>
```
krasznaa, line 54: You should be able to avoid warnings about an unused template parameter by not giving the value a name. Like:

```c++
template <typename T, device_address_space = device_address_space::global>
```

(Just like you would for regular function arguments.)

Though I'm not 1000% sure about this... :confused:

core/include/vecmem/memory/impl/device_atomic_ref.ipp
```diff
@@ -14,41 +14,51 @@
...
-    cl::sycl::atomic_##FNAME<value_type>(      \
-        cl::sycl::atomic<value_type>(cl::sycl::global_ptr<value_type>(PTR)))
-#define __VECMEM_SYCL_ATOMIC_CALL1(FNAME, PTR, ARG1)                         \
-    cl::sycl::atomic_##FNAME<value_type>(                                    \
-        cl::sycl::atomic<value_type>(cl::sycl::global_ptr<value_type>(PTR)), \
+#define __VECMEM_SYCL_ATOMIC_CALL0(FNAME, PTR)           \
+    cl::sycl::atomic_##FNAME<value_type, address>( \
+        cl::sycl::atomic<value_type, address == address_space::global
+                           ? ::sycl::memory_scope::device
+                           : ::sycl::memory_scope::work_group>(     \
```
krasznaa, line 21: Use the previously shown trait classes here as well.

core/include/vecmem/memory/impl/device_atomic_ref.ipp
```diff
@@ -14,41 +14,51 @@
...
+            cl::sycl::global_ptr<value_type>(PTR)))
```
krasznaa, line 22: I believe you'll also need to abstract this call. It shouldn't always be `sycl::global_ptr`.

krasznaa (changes requested):
I know the compilation itself is still broken, but to remind you: You'll need to add equivalents of the SYCL unit test for CUDA and HIP as well. :wink:

core/include/vecmem/memory/device_atomic_ref.hpp
krasznaa: And to make it clean, the actual struct should have some nicer name. Something like `vecmem::details::sycl_address_space`. (Matching with the `vecmem::device_address_space` name...)

stephenswat:
What is the envisioned functionality on CUDA devices here, which do not take an address space specifier?

krasznaa:
> What is the envisioned functionality on CUDA devices here, which do not take an address space specifier?

With this PR we will just ignore the second template parameter for the CUDA and HIP implementations. But at one point we should look into [cuda::atomic_ref](https://nvidia.github.io/libcudacxx/extended_api/synchronization_primitives/atomic_ref.html). The templating on that one is a bit different than what SYCL does, but I believe the proposed schema here (with a simple global/local setup) will be possible to fit into the libcu\+\+ setup as well. :thinking:

tests/sycl/test_sycl_containers.sycl
```diff
@@ -294,6 +294,73 @@ TEST_F(sycl_containers_test, atomic_device_memory) {
     }
 }
 
+/// Test atomic access to local memory
+TEST_F(sycl_containers_test, atomic_local_ref) {
+
+    // Create the SYCL queue that we'll be using in the test.
+    cl::sycl::queue queue;
```
krasznaa, line 301: Something like

```c++
if (queue.get_device().get_info<::sycl::info::device_type>() != ::sycl::info::device_type::gpu) {
    GTEST_SKIP();
}
```

is still missing. (For now let's only enable these things for GPUs. As I doubt that "local memory" is even defined for CPU OpenCL devices. May be worth checking though...)

guilhermeAlmeida1: Yup, I also couldn't find any clear documentation on whether or not that exists.
I'll push the latest version here, which I think should pass all tests, but I'm still not understanding some issues I'm running into locally,

guilhermeAlmeida1:
This should now be ready for review, but I've yet not been able to run the tests with the old oneAPI version locally, and this isn't running in the CI as well

Some of the naming of little messy with the local/shared difference between SYCL / CUDA & HIP

guilhermeAlmeida1:
I'm changing this back into a draft for now to ensure it doesn't get merged in before I test it with the old version of oneAPI locally.

guilhermeAlmeida1:
This no longer a draft as it has now been tested with oneAPI v2021.4.0 and passed.

krasznaa (approved):
Let's finally get this in! The code looks good all around. Even if I'm not the biggest fan of using `add` for `address` in `details::sycl_address_space`. :stuck_out_tongue: (But that's hidden enough that it shouldn't confuse the users.)

PR #230, opened 2023-05-31
Introduce vecmem::sycl::local_accessor, main branch (2023.05.31.)

krasznaa:
While in [traccc](https://github.com/acts-project/traccc) I told @guilhermeAlmeida1 that we shouldn't over-think how we use local/shared memory with SYCL, in this project I finally changed my mind...

With modern versions of Intel's compiler there is a warning about using `::sycl::access::target::local` on a "normal" `::sycl::accessor`. One is rather supposed to use `::sycl::local_accessor` nowadays. So to help with using the correct type in our code, I now introduced `vecmem::sycl::local_accessor`, which would resolve to the correct type with whatever SYCL compiler we use. By checking during CMake configuration whether `::sycl::local_accessor` is usable, and if it is, using that. If it's not, it falls back on using `::sycl::accessor`.

I thought this would be a more robust setup than to rely on some compiler macros.

As long as people are on board with this, we should eventually update the traccc code to use this type as well.

### DIFF
--- a/sycl/CMakeLists.txt
+++ b/sycl/CMakeLists.txt
@@ -84,6 +84,14 @@ if( VECMEM_HAVE_SYCL_MEMSET )
    target_compile_definitions( vecmem_sycl PRIVATE VECMEM_HAVE_SYCL_MEMSET )
 endif()
 
+# Check if sycl::local_accessor is available.
+vecmem_check_sycl_code_compiles( VECMEM_HAVE_SYCL_LOCAL_ACCESSOR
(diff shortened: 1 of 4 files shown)


guilhermeAlmeida1:
This works fine by me, but I have a question. If Intel's design choice is to send these kinds of warnings or in some cases disfavour backwards compatiblity, should it be up to vecmem to provide a SYCL to "old-SYCL" common interface, when it's no longer about common code between different platforms but a single one?

PR #231, opened 2023-06-01
(Open)SYCL Fixes, main branch (2023.06.01.)

krasznaa:
This is to make our project work correctly with [OpenSYCL 0.9.4](https://github.com/OpenSYCL/OpenSYCL/releases/tag/v0.9.4). (Formerly known as the artist [hipSYCL](https://github.com/OpenSYCL/OpenSYCL).)

As it turns out, we've been using the "SYCL namespace" a bit too loosely lately. :frowning: The SYCL2020 standard is actually fairly clear on what one should do.

https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:headers-and-namespaces

  - When using `<CL/sycl.hpp>`, you need to use the `::cl::sycl` namespace;
  - and with `<sycl/sycl.hpp>` you need to use the `::sycl` namespace.

The Intel implementation is quite forgiving with this, it provides all types in both the `::cl::sycl` and `::sycl` namespaces in all cases. But OpenSYCL rather goes for the following strict setup:

| Header | Namespace |
|---|---|
| `<CL/sycl.hpp>` | `::cl::sycl` |
| `<sycl/sycl.hpp>` | `::sycl` |
| `<hipSYCL/sycl.hpp>` | `::hipsycl` |

With all this in mind, since `<CL/sycl.hpp>` + `::cl::sycl` are the things supported since the earliest times, I went with those in our code. :thinking: (We could go for the `<sycl/sycl.hpp>` + `::sycl` combination in the future at one point, once SYCL2020 becomes the minimum that we would support.)

I also made sure that when no [oneAPI printf](https://intel.github.io/llvm-docs/doxygen/namespacesycl_1_1__V1_1_1ext_1_1oneapi_1_1experimental.html#a087ad354789a1bc1659c7bbea1f5a3f3) function is available, the standard [::printf](https://en.cppreference.com/w/cpp/io/c/fprintf) function could be used correctly at least in host code during SYCL code compilation. To make this work I had to fix some "actual bugs" hiding in our code...

### DIFF
--- a/core/CMakeLists.txt
+++ b/core/CMakeLists.txt
@@ -196,9 +196,9 @@ if( VECMEM_BUILD_SYCL_LIBRARY )
          VECMEM_SYCL_PRINTF_FUNCTION=cl::sycl::ONEAPI::experimental::printf )
    else()
       message( WARNING "No valid printf function found for SYCL."
-         " Enabling debug messages will likely not work." )
+         " Enabling debug messages will likely not work in device code." )
       target_compile_definitions( vecmem_core PUBLIC
-         VECMEM_ONEAPI_PRINTF_FUNCTION=printf )
+         VECMEM_SYCL_PRINTF_FUNCTION=printf VECMEM_MSG_ATTRIBUTES= )
    endif()
 
    # Test whether sycl::atomic_ref is available.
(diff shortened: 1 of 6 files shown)


guilhermeAlmeida1 (approved):
Looks good to me

beomki-yeo:
Can we get this in traccc soon?

PR #232, opened 2023-06-09
Ref: Allow 0 byte (de)allocation

guilhermeAlmeida1:


stephenswat (approved):
Looks good, thanks! :)

krasznaa (changes requested):
The initiative is very good, but I would do it a little differently. 🤔

As long as we want to return `nullptr` for 0 sized requests without any error notifications (which I can be on board with), then let's just add

```c++
if (nbytes == 0) {
    return nullptr;
}
```

at the very beginning of every single `do_allocate(...)` function, and add

```c++
if (p == nullptr) {
    return;
}
```

at the beginning of every `do_deallocate(...)` function.

Not just the SYCL ones, but for all! Sure, right now you were only seeing issues with SYCL. But we should make sure that all our memory resources behave the same way. Independent of how CUDA and HIP (will) react to 0 sized allocations for instance.

krasznaa:
Also, could you update (possibly in a separate PR) https://github.com/acts-project/vecmem/blob/main/core/include/vecmem/containers/impl/jagged_vector_buffer.ipp not to ask for 0-sized allocations? The `vecmem::data::vector_buffer` and `vecmem::data::jagged_vector_data` classes already explicitly don't ask for any allocation when the size of the allocation would be zero.
  - https://github.com/acts-project/vecmem/blob/main/core/include/vecmem/containers/impl/jagged_vector_data.ipp#L24-L26
  - https://github.com/acts-project/vecmem/blob/main/core/include/vecmem/containers/impl/vector_buffer.ipp#L32-L35

guilhermeAlmeida1:
> Also, could you update (possibly in a separate PR) https://github.com/acts-project/vecmem/blob/main/core/include/vecmem/containers/impl/jagged_vector_buffer.ipp not to ask for 0-sized allocations? The `vecmem::data::vector_buffer` and `vecmem::data::jagged_vector_data` classes already explicitly don't ask for any allocation when the size of the allocation would be zero.
> 
> * https://github.com/acts-project/vecmem/blob/main/core/include/vecmem/containers/impl/jagged_vector_data.ipp#L24-L26
> * https://github.com/acts-project/vecmem/blob/main/core/include/vecmem/containers/impl/vector_buffer.ipp#L32-L35

Ah, that explains why I was only running into this issue with jagged vectors. I assume it's still probably better to have the possibility to use `allocate()` and `deallocate()` with 0 bytes anyways even if I fix that.

krasznaa:
> > Also, could you update (possibly in a separate PR) https://github.com/acts-project/vecmem/blob/main/core/include/vecmem/containers/impl/jagged_vector_buffer.ipp not to ask for 0-sized allocations? The `vecmem::data::vector_buffer` and `vecmem::data::jagged_vector_data` classes already explicitly don't ask for any allocation when the size of the allocation would be zero.
> > 
> > * https://github.com/acts-project/vecmem/blob/main/core/include/vecmem/containers/impl/jagged_vector_data.ipp#L24-L26
> > * https://github.com/acts-project/vecmem/blob/main/core/include/vecmem/containers/impl/vector_buffer.ipp#L32-L35
> 
> Ah, that explains why I was only running into this issue with jagged vectors. I assume it's still probably better to have the possibility to use `allocate()` and `deallocate()` with 0 bytes anyways even if I fix that.

Yes. This PR should continue. I only brought up the jagged buffer issue in addition. :wink:

krasznaa (approved):
I'm on board with this. Though... what happened with updating the [memory resource unit test](https://github.com/acts-project/vecmem/blob/main/tests/common/memory_resource_test_basic.ipp) to test 0 sized allocations? I was on board with including that update in this PR as well...

guilhermeAlmeida1:
> I'm on board with this. Though... what happened with updating the [memory resource unit test](https://github.com/acts-project/vecmem/blob/main/tests/common/memory_resource_test_basic.ipp) to test 0 sized allocations? I was on board with including that update in this PR as well...

Oh, I went back a commit and completely forgot about that. Will fix

krasznaa (approved):
:+1:

PR #234, opened 2023-06-12
Doxygen Improvement, main branch (2023.06.12.)

krasznaa:
Last week I was pointing people at the `vecmem::details::aligned_multiple_placement` function, and realised along the way that its documentation doesn't show up all too well on the generated Doxygen page. :frowning:

image

With this change the documentation looks like:

image

So this is just to help my OCD... :stuck_out_tongue:


stephenswat (approved):
I forgot all about this funky bit of documentation. :smile:

PR #235, opened 2023-07-03
Windows Warning Fixes, main branch (2023.07.03.)

krasznaa:
MSVC and CUDA on Windows don't like it when we use an STL type member variable in an exported class. Which I absolutely want to do in these places. So ended up silencing these warnings explicitly.

Unfortunately our CI is blind to these, as all of these warnings were in the SYCL code. (And there's no good way to do a SYCL build on Windows in the CI...)

I've been trying to find a way in which I could hide the smart pointers from the class's header, but was not able to find one. 😦 So massive amounts of pre-compiler statements it is...

### DIFF
--- a/sycl/include/vecmem/utils/sycl/async_copy.hpp
+++ b/sycl/include/vecmem/utils/sycl/async_copy.hpp
@@ -1,7 +1,7 @@
 /*
  * VecMem project, part of the ACTS project (R&D line)
  *
- * (c) 2021-2022 CERN for the benefit of the ACTS project
+ * (c) 2021-2023 CERN for the benefit of the ACTS project
  *
  * Mozilla Public License Version 2.0
  */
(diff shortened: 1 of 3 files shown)


paulgessinger:
I guess there's an underlying reason why they don't want you to export STL types. Maybe for our purposes it's fine to just disable the warning, but I think generally people use PIMPL to get around this.

krasznaa:
> I guess there's an underlying reason why they don't want you to export STL types. Maybe for our purposes it's fine to just disable the warning, but I think generally people use PIMPL to get around this.

What I'm doing here **is** PIMPL. I just don't want to use `new` and `delete` by hand. And to be fair, even https://en.cppreference.com/w/cpp/language/pimpl is describing PIMPL with [std::unique_ptr](https://en.cppreference.com/w/cpp/memory/unique_ptr). 🤔 

Yes, the warning is definitely there so that when you build a DLL, it would not depend on the exact version of the C\+\+ standard library that it was built with. But since this project fundamentally relies on STL (with how we implement different memory resources), this ship has sailed anyway... (Note that our project **can** be built as a set of static libraries. Which can make it possible to use it "without warnings" in a Windows project. But that's another story.)

paulgessinger:
I got the idea :)

I don't see a way around this if you don't want to do `new`/`delete` in these cases, so let's go!

PR #240, opened 2023-10-12
Memory Resource Re-Design, main branch (2023.10.12.)

krasznaa:
In a bit of a monstrous PR, I re-designed `vecmem::details::memory_resource_base` a little. It is now a project specific interface, which mimics [std::pmr::memory_resource](https://en.cppreference.com/w/cpp/memory/memory_resource), but doesn't actually inherit from it.

In addition I created `vecmem::details::memory_resource_adaptor`, which could be used to provide classes implementing the `std::pmr::memory_resource` interface using classes that implement the `vecmem::details::memory_resource_base` interface.

I also split most of the memory resources in the `vecmem::core` library in two. A "main class" implementing the `vecmem::details::memory_resource_base` interface, and an "implementation class" that doesn't inherit from anything, it just implements the behaviour of the class in the appropriate way, with a [PIMPL](https://en.cppreference.com/w/cpp/language/pimpl) design.

The goal of all of this was to get rid of a whole lot of MSVC warning suppressions from the code. For this purpose I couldn't even use [std::unique_ptr](https://en.cppreference.com/w/cpp/memory/unique_ptr) in my PIMPL implementation, as that would've completely undone the whole purpose of why I'm separating the implementation from the interface...

While at it, I also made some changes to unify the implementation of the different memory resources a bit. Since at this point, why not... 😛

The public API of the project does not change with this update. In the end I did need to update one of the unit tests, but just because that test was using `vecmem::details::memory_resource_base` directly. The ABI of the libraries does change significantly however. So if you guys are okay with this change, I think this will be the point at which we could tag version `1.0.0`, with a new library ABI version. 😉 (Switching it from `0` to `1`.)


core/include/vecmem/memory/details/arena_memory_resource.hpp
```diff
@@ -0,0 +1,66 @@
...
+    virtual void* mr_allocate(std::size_t bytes, std::size_t) override;
+    /// De-allocate a previously allocated memory block
+    virtual void mr_deallocate(void* p, std::size_t bytes,
+                               std::size_t) override;
+
+    /// @}
+
+private:
+    /// Object performing the heavy lifting for the memory resource
+    arena_memory_resource_impl* m_impl;
```
paulgessinger, line 62: You mentioned this in the description already, but it's really unfortunately not being able to use smart-pointers for the PIMPL here. If there's absolutely now way to get MSVC to swallow that, fine, but it would really be a shame.

paulgessinger: I *think* MSVC should actually allow you to export `std::unique_ptr` in principle.

krasznaa: It's a pretty complicated class. I looked at it again for a bit yesterday in `libstdc++`. And I can sort of see Microsoft's point after that.

Since what they want is that after you build your DLL with MSVC XX, that DLL would be usable as-is with MSVC YY. And that is not guaranteed with `std::unique_ptr` in the "memory footprint" of your classes. 🤔 Note that various of our memory resources still receive references to STL types. I guess one could argue there that as long as these are references, the actual implementation in the client that provides the reference can be different from the implementation that the VecMem code saw when it was compiled. As long as the API didn't change, this would still keep the code working.

However the logic of even that breaks down with the `vecmem::instrumenting_memory_resource::get_events()` function. 😦 Since that returns an `std::vector` of objects. And in case `std::vector` changes its internal representation between the time that `vecmem_core.dll` is built, and that the user makes use of it, that function would change its ABI. 🤔 

But, MSVC doesn't complain about it, so I'm not going to worry about it for now... 😛

paulgessinger: But does it actually **fail** due to `unique_ptr` then? I see the argument about broken ABI compat because it's not guaranteed, but that's not something that we want to try to even guarantee to that degree.

If we can make MSVC shut up about `unique_ptr`, I'd be in favor of technically not having real ABI stability, but using the smart pointer instead of manual deletes.

paulgessinger: It would reduce the number of warnings being suppressed but I see your point. 

Another option would be to not export the class altogether, but only the methods that we actually want to support. In that case I think MSVC would not complain about `unique_ptr`. That's not super great either, of course.

Ok, then I guess we should follow this up with a custom unique smart pointer to hide all the `delete`s and raw `new` calls, in my opinion.

krasznaa: As long as the client needs to be able to construct the memory resource object, they need to know how to construct an `std::unique_ptr`. (Or at least how much memory to reserve on the stack for making that happen.) We can't suppress exporting this aspect of the class. 😦

Note that since it's a private member, technically it's only the vecmem code that actually interacts with the `std::unique_ptr` ABI. I think it's more of a technical thing of how the Windows ABI communicates how much memory is needed for a given object, through whatever mechanism they have for exporting symbols. 🤔

**Very** long story short, let's add a simple smart pointer for this purpose in a follow-up PR.

krasznaa:
@stephenswat, if you have any objections, let me know shortly. As I plan to push this in, in not too long.

stephenswat:
Sure, go ahead.

PR #243, opened 2023-10-24
DevContainer Update, main branch (2023.10.24.)

krasznaa:
Updated the DevContainer setup of the repository to use a new AlmaLinux 9 based image, which would have recent CUDA, HIP and oneAPI versions.

Greatly simplifying its setup along the way, with the help of the [ms-vscode.cmake-tools](https://marketplace.visualstudio.com/items?itemName=ms-vscode.cmake-tools) plugin, which takes care of pretty much all of the integration between [CMake](https://cmake.org/) and [VSCode](https://code.visualstudio.com/) for the project.

Note that in this setup, all the "language specific" code is correctly parsed by IntelliSense as well. :smile:

image
image
image

(You decide whose documentation is best. :stuck_out_tongue:)

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,11 +1,11 @@
 # VecMem project, part of the ACTS project (R&D line)
 #
-# (c) 2021 CERN for the benefit of the ACTS project
+# (c) 2021-2023 CERN for the benefit of the ACTS project
 #
 # Mozilla Public License Version 2.0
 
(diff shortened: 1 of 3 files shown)


krasznaa:
I have to admit, one reason why I'm making this update is that I intend to try how far I can get with developing the project's code on my iPad through GitHub Codespaces. :laughing:

paulgessinger (approved):
Sure, why not.

PR #244, opened 2023-10-25
Warning Suppression Removal, main branch (2023.10.25.)

krasznaa:
Being drunk on the success of #242, narrowed the exports on more classes. This allowed the removal of all warning suppression rules from the public headers of the project.

Had to add an explicit destructor for `vecmem::cuda::async_copy`, so that client code would not need to know how to destruct the private members of the class.

Note that at this stage the only remaining warning suppression is in: https://github.com/acts-project/vecmem/blob/main/sycl/src/utils/sycl/cuda_assert_polyfill.sycl But with that being a hacky workaround to begin with, that will just stay as-is...

### DIFF
--- a/cuda/include/vecmem/utils/cuda/async_copy.hpp
+++ b/cuda/include/vecmem/utils/cuda/async_copy.hpp
@@ -1,7 +1,7 @@
 /*
  * VecMem project, part of the ACTS project (R&D line)
  *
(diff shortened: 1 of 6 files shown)


stephenswat (approved):
Looks good. :slightly_smiling_face:

PR #245, opened 2023-10-31
Multi-Alloc Fix, main branch (2023.10.31.)

krasznaa:
Made the `vecmem::details::aligned_multiple_placement` function work with other than 2 arguments. As it turns out, so far no client needed to use a different number of allocations through this function. :confused: But in my tests for an SoA based EDM for [traccc](https://github.com/acts-project/traccc) (more on that once I have something usable to show...), I managed to run into this. :thinking:

Note that how much of the code style in `vecmem::details::max` comes from CoPilot, we may need to include authorship credits for it at one point. :stuck_out_tongue:

### DIFF
--- a/core/include/vecmem/containers/impl/aligned_multiple_placement.ipp
+++ b/core/include/vecmem/containers/impl/aligned_multiple_placement.ipp
@@ -159,7 +159,7 @@ aligned_multiple_placement(vecmem::memory_resource &r, Ps... ps) {
      * parameter pack, which will determine the amount of additional space
      * we need to allocate.
      */
-    std::size_t alignment = std::max(alignof(Ts)...);
+    std::size_t alignment = vecmem::details::max(alignof(Ts)...);
 
     /*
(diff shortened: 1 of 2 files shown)


stephenswat (approved):
I'll approve this because I like the implementation of `vecmem::details::max`, but there is actually an easier way of doing this, namely in `aligned_multiple_placement.ipp`:

```cpp
std::size_t alignment = std::max({alignof(Ts)...});
```

Note the addition of the curly braces to create an initializer list to invoke signature 3 of [`std::max`](https://en.cppreference.com/w/cpp/algorithm/max).

But as I said I like the functional style of the `vecmem::details::max` implementation. :)

stephenswat:
Also, either way - well spotted!

krasznaa:
Huhh... Clearly I was google-ing badly. :confused: Since of course I was looking for a standard algorithm/utility for this at first.

Either way, let's just go with this then. I'll anyway have plenty of more variadic templates with the EDM developments...

PR #250, opened 2023-12-05
Extra Utility Types, main branch (2023.12.05.)

krasznaa:
This is step 1 in getting #246 into the repository.

Introduced `vecmem::tuple`, `vecmem::details::disjunction` and `vecmem::details::negation`. All in preparation for the SoA EDM base classes.
  - `vecmem::details::disjunction` is a stand-in for [std::disjunction](https://en.cppreference.com/w/cpp/types/disjunction) with [C++14](https://en.cppreference.com/w/cpp/14). (Which is still our minimum requirement for device code!)
  - In a similar manner, `vecmem::details::negation` is a stand-in for [std::negation](https://en.cppreference.com/w/cpp/types/negation) in device code.
  - `vecmem::tuple` is a shameless copy of [detray::tuple](https://github.com/acts-project/detray/blob/main/core/include/detray/utils/tuple.hpp). With very slight modifications for [C++14](https://en.cppreference.com/w/cpp/14) compatibility.

Not sure yet how I'll make the rest of #246 digestible, but for now let's just get this reviewed. Then we'll see about the rest. :wink:

### DIFF
--- a/core/CMakeLists.txt
+++ b/core/CMakeLists.txt
@@ -124,6 +124,8 @@ vecmem_add_library( vecmem_core core
    "include/vecmem/utils/debug.hpp"
    "src/utils/memory_monitor.cpp"
    "include/vecmem/utils/memory_monitor.hpp"
+   "include/vecmem/utils/tuple.hpp"
+   "include/vecmem/utils/impl/tuple.ipp"
    "include/vecmem/utils/type_traits.hpp"
    "include/vecmem/utils/types.hpp" )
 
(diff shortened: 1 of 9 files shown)


stephenswat (approved):

krasznaa:
Only noticed with a delay that the constructors of `vecmem::tuple` should be simplified a bit. Just as for `vecmem::data::vector_view` and friends, if we want to use such types directly as parameters of a SYCL kernel, they have to have "proper default constructors".

And this change seemingly also brought a build failure forward. :thinking: Interesting that Clang (which is what I tested this change with) didn't complain about that variable. :thinking:

PR #251, opened 2023-12-06
Add SoA View and Device Classes, main branch (2023.12.06.)

krasznaa:
Following #250, this is the 2nd PR for reviewing everything from #246. 😄

Introduced `vecmem::edm::view` and `vecmem::edm::device`. (And all the utility code that they require.) They can't be used for much at the moment, without adding some host types as well, but they allow for some trivial tests even like this at least.
  - `vecmem::edm::device` is not much more than a `vecmem::tuple` of appropriate types, with some helper functions. The implementation gets a bit complicated even like this, but logically that class is really just some syntactic sugar around a tuple.
    * The only slightly tricky thing it provides is the ability to resize / enlarge resizable containers in a thread-safe way.
  - `vecmem::edm::view` on the other hand may look a bit over-complicated in this PR. 🤔
    * `m_capacity` is simply the overall capacity of the SoA container being described;
    * `m_views` is a tuple of "view types" for the individual variables. Logically this is still meant to be simple to follow.
    * `m_payload` is a type-less view of the memory block holding the payload of the container. In case it uses a contiguous memory allocation. Which the `vecmem::edm::buffer` class, introduced in a follow-up PR, will provide. When creating a view on top of a host object directly, this view will be invalid.
    * `m_size` is a type-less view of the memory block holding the resizable size(s) of the container. The thing that `vecmem::copy` would be able to easily initialize to `0` for resizable containers/buffers. (For non-resizable containers this view is invalid.)
    * `m_layout` and `m_host_layout` are the type-less views of the data describing "the layout" of the view's payload. Which is needed for describing jagged vectors. `vecmem::copy` will set up containers for usage on a device using these views. (Copying from the host view to the device one when necessary.)

As I wrote, this PR does not actively test every aspect of these classes. For that we'll need the `vecmem::edm::host` and `vecmem::edm::buffer` classes as well. (Coming soon to a repository near you. :wink:)

### DIFF
--- a/core/CMakeLists.txt
+++ b/core/CMakeLists.txt
@@ -61,6 +61,15 @@ vecmem_add_library( vecmem_core core
    "include/vecmem/memory/unique_ptr.hpp"
    "include/vecmem/memory/details/is_aligned.hpp"
    "src/memory/details/is_aligned.cpp"
+   # EDM types.
+   "include/vecmem/edm/device.hpp"
+   "include/vecmem/edm/impl/device.ipp"
+   "include/vecmem/edm/view.hpp"
+   "include/vecmem/edm/impl/view.ipp"
+   "include/vecmem/edm/schema.hpp"
+   "include/vecmem/edm/details/device_traits.hpp"
+   "include/vecmem/edm/details/schema_traits.hpp"
+   "include/vecmem/edm/details/view_traits.hpp"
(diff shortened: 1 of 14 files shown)


krasznaa:
Taking an executive decision here for the sake of time. :wink:

PR #255, opened 2023-12-10
Add Copy for SoA Containers, main branch (2023.12.10.)

krasznaa:
In the 5th step of the #246 saga.

Taught `vecmem::copy` about the `vecmem::edm` types.

At the same time fixed a logic bug in `vecmem::copy::set_sizes` (noticed this one while making a mistake in the testing code...), and made `vecmem::edm::host::resize` a bit smarter. (Making sure that jagged vectors would be set up with the correct memory resource after the resize.)

Note that not everything is implemented for the SoA types yet. Generic `vecmem::copy::get_sizes` and `vecmem::copy::set_sizes` functions should still be added later on. I just couldn't come up with a good API for them yet. (Similar to how `vecmem::edm::buffer` cannot be set up yet for jagged vector variables of differing inner sizes.)

### DIFF
--- a/core/include/vecmem/utils/impl/copy.ipp
+++ b/core/include/vecmem/utils/impl/copy.ipp
@@ -10,6 +10,7 @@
 // VecMem include(s).
 #include "vecmem/containers/details/resize_jagged_vector.hpp"
 #include "vecmem/containers/jagged_vector.hpp"
+#include "vecmem/edm/details/schema_traits.hpp"
 #include "vecmem/memory/host_memory_resource.hpp"
 #include "vecmem/utils/debug.hpp"
 #include "vecmem/utils/type_traits.hpp"
@@ -20,6 +21,7 @@
 #include <numeric>
 #include <sstream>
 #include <stdexcept>
+#include <type_traits>
 
 namespace vecmem {
 
@@ -373,13 +375,14 @@ copy::event_type copy::set_sizes(
          i < data.size(); ++i) {
         if (data.host_ptr()[i].size_ptr() == nullptr) {
             perform_copy = false;
-        } else if (perform_copy == true) {
+            if (data.host_ptr()[i].capacity() != sizes[i]) {
+                throw std::runtime_error(
+                    "Non-resizable jaggged vector does not match the requested "
+                    "size");
+            }
+        } else if (perform_copy == false) {
             throw std::runtime_error(
                 "Inconsistent target jagged vector view received for resizing");
-        } else if (data.host_ptr()[i].capacity() != sizes[i]) {
-            throw std::runtime_error(
-                "Non-resizable jaggged vector does not match the requested "
-                "size");
         }
     }
     // If no copy is necessary, we're done.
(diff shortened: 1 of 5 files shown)


core/include/vecmem/utils/impl/copy.ipp
```diff
@@ -395,6 +398,138 @@ copy::event_type copy::set_sizes(
...
+    static_assert(sizeof...(VARTYPES1) == sizeof...(VARTYPES2),
+                  "Can only use compatible types in the copy");
+    static_assert(std::is_same<edm::schema<VARTYPES1...>,
+                               edm::schema<VARTYPES2...>>::value ||
+                      details::is_same_nc<edm::schema<VARTYPES1...>,
+                                          edm::schema<VARTYPES2...>>::value,
+                  "Can only use compatible types in the copy");
+
+    // Resize the output object to the correct size(s).
+    resize_impl(from_view, to_vec, cptype,
```
paulgessinger, line 526: Ah I see, in this case it does. So view -> view assumes it's already correctly sized, and view->vector will resize, then call view -> view ?

core/include/vecmem/utils/impl/copy.ipp
```diff
@@ -543,4 +678,216 @@ bool copy::is_contiguous(const data::vector_view<TYPE>* data,
     return true;
 }
 
+template <typename... VARTYPES, std::size_t INDEX, std::size_t... Is>
```
paulgessinger, line 681: NIT: I guess this recursion could be implemented without the full integer sequence by just decrementing.

core/include/vecmem/utils/impl/copy.ipp
```diff
@@ -395,6 +398,138 @@ copy::event_type copy::set_sizes(
     return create_event();
 }
 
+template <typename SCHEMA>
+copy::event_type copy::setup(edm::view<SCHEMA> data) const {
```
paulgessinger, line 402: NIT: `init()`?

tests/core/test_core_edm_copy.cpp
```diff
@@ -0,0 +1,279 @@
...
+        vecmem::data::buffer_type::fixed_size};
+    m_copy.setup(buffer1);
+
+    vecmem::edm::buffer<jagged_schema> buffer2{
+        std::vector<std::size_t>{2, 0, 3}, m_resource, nullptr,
+        vecmem::data::buffer_type::fixed_size};
+    m_copy.setup(buffer2);
+
+    m_copy(vecmem::get_data(host), buffer1, vecmem::copy::type::host_to_host);
+    m_copy(buffer1, buffer2, vecmem::copy::type::host_to_host);
```
paulgessinger, line 235: Should this test the error case where the sizes don't line up but `fixed_size` is given?

tests/core/test_core_edm_copy.cpp
```diff
@@ -0,0 +1,279 @@
...
+    fill(host);
+
+    vecmem::edm::buffer<jagged_schema> buffer1{
+        std::vector<std::size_t>{2, 0, 3}, m_resource, nullptr,
+        vecmem::data::buffer_type::fixed_size};
+    m_copy.setup(buffer1);
+
+    vecmem::edm::buffer<jagged_schema> buffer2{
+        std::vector<std::size_t>{5, 5, 5}, m_resource, nullptr,
+        vecmem::data::buffer_type::resizable};
```
paulgessinger, line 272: Do you test the resizing somewhere?

paulgessinger (commented):
Looks pretty sophisticated. I can't claim I've wrapped my head around every single, but I think the factorization of resizing vs copying makes sense.

If I remember correctly, copy itself is already set up to handle non-trivially copyable types?

core/include/vecmem/utils/impl/copy.ipp
krasznaa: Hmm... I do actually like `init(...)` better than `setup(...)`. But we already have 2 `vecmem::copy::setup(...)` functions for 1D and jagged vectors. :thinking:

In this PR I would stay with the `setup(...)` name. But I'm absolutely open to renaming it to `init(...)` at one point. At that point we'll have to update all the client code. (`setup(...)` is called in a lot of places in [traccc](https://github.com/acts-project/traccc).)

core/include/vecmem/utils/impl/copy.ipp
```diff
@@ -395,6 +398,138 @@ copy::event_type copy::set_sizes(
...
+                  "Can only use compatible types in the copy");
+    static_assert(
+        std::conjunction_v<std::is_same<VARTYPES1, VARTYPES2>...> ||
+            std::conjunction_v<
+                edm::type::details::is_same_nc<VARTYPES1, VARTYPES2>...>,
+        "Can only use compatible types in the copy");
+
+    // First, handle the simple case, when both views have a contiguous memory
+    // layout.
+    if ((from_view.payload().ptr() != nullptr) &&
```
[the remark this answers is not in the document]
krasznaa: This was a little bit hinted at in #254, but I didn't write it down really explicitly.

Let's start with how we currently determine contiguity for jagged vectors. That is done here:

https://github.com/acts-project/vecmem/blob/main/core/include/vecmem/utils/impl/copy.ipp#L528-L544

While it may look correct, it's actually kind of fragile. See:

https://forums.developer.nvidia.com/t/single-cudamemcpy-across-multiple-allocations

Under the right (or wrong...) circumstances that logic can mis-categorize a jagged vector view. :frowning: (It's just circumstantial enough that I still didn't fix it.)

At first I thought that I would write some similar, albeit more complicated logic for checking if an entire SoA container is contiguous in memory. Based on the pointers of the views of the individual variables in the container. But while for jagged vectors this logic is already fragile enough, for SoA containers it would be super brittle. Because: alignment!

In `vecmem::edm::buffer` I ended up using [vecmem::details::aligned_multiple_placement](https://github.com/acts-project/vecmem/blob/main/core/include/vecmem/containers/details/aligned_multiple_placement.hpp). This is because when allocating space for a bunch of arrays of different types, the very end of one array's block may not be the correct starting position for the next array's block. So even if a container is held by `vecmem::edm::buffer` in a single memory block, the individual views inside the `vecmem::edm::view` object may not be pointing at contiguous spaces in memory.

I considered a number of horrible solutions to this, but in the end decided to rather just add some variables to `vecmem::edm::view` that are **only** set by `vecmem::edm::buffer`. The `payload()` view is one of these. When constructing a `vecmem::edm::view` from a `vecmem::edm::host` container, `vecmem::edm::view::payload()` will never point to anything. Only when `vecmem::edm::view` describes a buffer, does `vecmem::edm::view::payload()` point to a non-null memory address. And since the memory allocated by `vecmem::edm::buffer` is always contiguous by construction, if `payload()` is "valid", the copy code can know that it can just copy things as they are. :wink:

I could continue the description a little longer as well, but hopefully this conveys the main idea. :wink:

core/include/vecmem/utils/impl/copy.ipp
```diff
@@ -395,6 +398,138 @@ copy::event_type copy::set_sizes(
...
+
+        // Copy the payload with a single copy operation.
+        operator()(from_view.payload(), to_view.payload(), cptype);
+
+        // If the target view is resizable, set its size.
+        if (to_view.size().ptr() != nullptr) {
+            // If the source is also resizable, the situation should be simple.
+            if (from_view.size().ptr() != nullptr) {
+                // Check that the sizes are the same.
+                if (from_view.size().size() != to_view.size().size()) {
```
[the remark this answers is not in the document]
krasznaa: The naming is probably not too intuitive here. :frowning:

`vecmem::edm::view::size()` returns a `vecmem::data::vector_view<char>` object. It's a type-less block of memory, which describes "the size" of the container in some way. Again, as described in #254 a bit, the handling of resizable containers that have some jagged vectors and those that have none, go quite differently. For the types without jagged vectors, we just have a single `unsigned int` in memory that describes the size of the entire container. But when jagged vectors are involved, it's not the container that's "resizable", but all the individual jagged vectors by themselves. So we instead have a bunch of `unsigned int`-s allocated, one for each "inner vector" of every jagged vector variable.

`vecmem::edm::view::size()` returns a type-less view into the memory block that describes the "dynamic/resizable size" of the container.

If we reached this point in the code, where we checked already that the schema of the two views is "compatible" (so they should have the exact same variables except for const-ness), and both of them are resizable, then the two of them not using the same number of `unsigned int`-s to describe their size is some serious logic bug in the code somewhere. Not on the user's side, but on ours. Since the code should really never ever reach this point without some coding error in this project.

Maybe the exception message should reflect this a bit more. :thinking: But for now. throwing an exception seemed like the best thing to do here. :thinking:

core/include/vecmem/utils/impl/copy.ipp
```diff
@@ -395,6 +398,138 @@ copy::event_type copy::set_sizes(
...
+                // Perform a dumb copy.
+                operator()(from_view.size(), to_view.size(), cptype);
+            } else {
+                // If not, then copy the size(s) recursively.
+                copy_sizes_impl(from_view, to_view, cptype,
+                                std::index_sequence_for<VARTYPES1...>{});
+            }
+        }
+
+        // Create a synhronization event.
```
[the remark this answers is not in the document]
krasznaa: :+1:

core/include/vecmem/utils/impl/copy.ipp
krasznaa: Exactly. That's the logic that the code has been using for 1D and jagged vectors as well. :wink: So that copying is only implemented between two views. Regardless of what exact type of object is holding the memory.

core/include/vecmem/utils/impl/copy.ipp
krasznaa: :thinking: Yeah... You're right, just a single `INDEX` template parameter could be enough. How strongly do you feel about this for the first version of the code?

tests/core/test_core_edm_copy.cpp
krasznaa: I was thinking about it... :thinking: The reason I found the pre-existing bug in `vecmem::copy::set_sizes(...)` is that I made a typo with the sizes in `host_to_fixed_device_jagged`. And then the test failed "in the wrong way".

Okay, let me see how easily I could put in some simple tests where I expect the code to throw some kind of exception. :thinking:

tests/core/test_core_edm_copy.cpp
krasznaa: Not yet. The PR coming after this one, which will be the last one in this saga, will add such tests. (I will need the `vecmem::edm::container` type to make the testing code for that functionality a little less tedious.)

krasznaa (commented):
Hahh! I knew that if only I send you guys a billion notifications, you may look at the code! :smile: But seriously, thanks for the comments!

core/include/vecmem/utils/impl/copy.ipp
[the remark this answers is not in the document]
krasznaa: Forgot to explain synchronization... That's a whole long story by itself. But this is to make `vecmem::sycl::async_copy` and `vecmem::cuda::async_copy` functional. Many of the functions in `vecmem::copy` return a `vecmem::copy::event_type` object. Which is created by the `vecmem::copy::create_event()` virtual function.

To make a long story shorter, have a look at : #209 That has a bit of a description of how this logic works.

core/include/vecmem/utils/impl/copy.ipp
krasznaa: Since the templating could indeed be simplified for all of the implementation functions, I went ahead and did that.

tests/core/test_core_edm_copy.cpp
krasznaa: Adding the test also revealed that `vecmem::copy` was in some cases throwing `std::length_error` and in others `std::runtime_error` for these types of errors.

So I tried to fix those, and added some explicit tests for the SoA copies. That they would throw the expected exceptions. (Didn't add such tests for the lower level tests at this point though...)

paulgessinger:
@krasznaa it depends on how intimidating it looks, and how my day is going 😅

PR #256, opened 2023-12-11
Add the SoA Container Class, main branch (2023.12.11.)

krasznaa:
In the final, 6th step of the #246 saga.

Introduced `vecmem::edm::container`, which is meant to be the primary UI for the users for managing an SoA EDM.

Introduced `vecmem::testing::simple_soa_container` and `vecmem::testing::jagged_soa_container` to demonstrate how one is meant to define SoA containers in the end.

With the help of the new types, re-wrote the SoA copy tests, and implemented them for CUDA, HIP and SYCL copies as well. To make the tests work, added some default functions to `vecmem::edm::buffer`, and some additional "schema traits" for some general quality-of-life improvements.

It's a draft for now, as I still need to add actual device code tests as well. But not today anymore...


krasznaa:
This PR is mostly testing code in the end. The new code is really just `vecmem::edm::container`. So if you have a look, it's in the end enough to just look at that. And how the "user interface" is put on top of `vecmem::edm::host` and `vecmem::edm::device` in it. :thinking:

PR #259, opened 2023-12-27
(Thrust) Pool Memory Resource, main branch (2023.12.27.)

krasznaa:
As a holiday project I looked through the [Thrust memory resources](https://github.com/NVIDIA/cccl/tree/main/thrust/thrust/mr) a little.

As a first thing, I selected [thrust::mr::disjoint_unsynchronized_pool_resource](https://github.com/NVIDIA/cccl/blob/main/thrust/thrust/mr/disjoint_pool.h) for inclusion into our project. I was actually lead astray at first with [thrust::mr::unsynchronized_pool_resource](https://github.com/NVIDIA/cccl/blob/main/thrust/thrust/mr/pool.h). 😛 That one had the "simplest" header name, so it took me a while to realize that it would only work with host-accessible memory resources. (As it allocates memory for the "meta data" that it needs, using the same resource that provides the memory to the users.) Since that code uses fewer allocations in total to do its job, it is kept as a separate memory resource in Thrust. But in this project I don't think we need a caching memory resource that only works with host-accessible upstream resources. 🤔 So in the end I chose the code from [thrust::mr::disjoint_unsynchronized_pool_resource](https://github.com/NVIDIA/cccl/blob/main/thrust/thrust/mr/disjoint_pool.h) for inclusion. Letting go of [thrust::mr::unsynchronized_pool_resource](https://github.com/NVIDIA/cccl/blob/main/thrust/thrust/mr/pool.h).

Since the new code needs some of the same underlying calculations as `vecmem::binary_page_memory_resource`, I moved the functions used by both into a new `integer_math.hpp` header. (I stole the naming from Thrust as well. See: https://github.com/NVIDIA/cccl/blob/main/thrust/thrust/detail/integer_math.h 😛)

Note that I went with simple [std::vector](https://en.cppreference.com/w/cpp/container/vector)-s in `vecmem::details::pool_memory_resource_impl`. The Thrust code uses a separate memory resource here for managing the host-accessible allocations. But since this is strictly an internal state of this memory resource, I didn't see much point in using `vecmem::vector` instead of `std::vector`. In fact, I feared that the former would be slower. (Since in the code all allocations/deallocations with `vecmem::vector` would absolutely need to go through virtual function calls. The compiler would never be able to deduce at compile time what exact memory resource is being used with the `vecmem::vector`-s. 🤔)

I believe we'll have some fun with this code. 😉 I'll add benchmarking for it in a later PR.


core/src/utils/integer_math.hpp
```diff
@@ -0,0 +1,96 @@
...
+}
+
+/// Compute the base-2 logarithm of a number
+///
+/// @param x The number to compute the logarithm of
+/// @return The base-2 logarithm of @c x
+///
+inline std::size_t log2(std::size_t x) {
+
+    static constexpr std::size_t num_bits = 8 * sizeof(std::size_t);
```
stephenswat, line 58: At the risk of being the ultimate nitpicker, the C++ standard doesn't guarantee that a byte is exactly 8 bits. Technically it would be better to use `CHAR_BIT` instead of 8. :nerd_face:

core/src/memory/details/pool_memory_resource_impl.cpp
stephenswat, line 1: I won't review this code because it came from Thrust directly, but it may be useful to note that this code is originally licensed under the Apache 2.0 licence which reads:

> 4. Redistribution. You may reproduce and distribute copies of the
>       Work or Derivative Works thereof in any medium, with or without
>       modifications, and in Source or Object form, provided that You
>       meet the following conditions:
> 
>       (a) You must give any other recipients of the Work or
>           Derivative Works a copy of this License; and
> 
>       (b) You must cause any modified files to carry prominent notices
>           stating that You changed the files; and
> 
>       (c) You must retain, in the Source form of any Derivative Works
>           that You distribute, all copyright, patent, trademark, and
>           attribution notices from the Source form of the Work,
>           excluding those notices that do not pertain to any part of
>           the Derivative Works; and
> 
>       (d) If the Work includes a "NOTICE" text file as part of its
>           distribution, then any Derivative Works that You distribute must
>           include a readable copy of the attribution notices contained
>           within such NOTICE file, excluding those notices that do not
>           pertain to any part of the Derivative Works, in at least one
>           of the following places: within a NOTICE text file distributed
>           as part of the Derivative Works; within the Source form or
>           documentation, if provided along with the Derivative Works; or,
>           within a display generated by the Derivative Works, if and
>           wherever such third-party notices normally appear. The contents
>           of the NOTICE file are for informational purposes only and
>           do not modify the License. You may add Your own attribution
>           notices within Derivative Works that You distribute, alongside
>           or as an addendum to the NOTICE text from the Work, provided
>           that such additional attribution notices cannot be construed
>           as modifying the License.
> 
>       You may add Your own copyright statement to Your modifications and
>       may provide additional or different license terms and conditions
>       for use, reproduction, or distribution of Your modifications, or
>       for any such Derivative Works as a whole, provided Your use,
>       reproduction, and distribution of the Work otherwise complies with
>       the conditions stated in this License.

To meet the legal requirements, I think we should extend the header to mention that this code is a modified version of original code, including a link to the APL2.0 as well as a mention and link to the original project.

core/src/utils/integer_math.hpp
```diff
@@ -0,0 +1,96 @@
...
+    // Return the result.
+    return result;
+}
+
+/**
+ * @brief Rounds a size up to the nearest power of two, and returns the power
+ * (not the size itself).
+ */
+inline std::size_t round_up(std::size_t size) {
+    for (std::size_t i = 0; i < 32; i++) {
```
stephenswat, line 87: Instead of 32, this should read `sizeof(std::size_t)`.

Also I think this contains a potential infinite loop if is called on `INT_MAX` (or its equivalent for `std::size_t`), but that's probably fine.

stephenswat (commented):
Minor comments.

core/src/utils/integer_math.hpp
krasznaa: 👍

core/src/memory/details/pool_memory_resource_impl.cpp
krasznaa: I added a comment to the Doxygen documentation of `vecmem::pool_memory_resource` with a link to the Apache 2 license. This should be enough... right?

core/src/utils/integer_math.hpp
krasznaa: Added an assertion as well, just in case. Since the edge case is actually a little larger in my mind. 🤔 There are a bunch of numbers above the largest power of 2 value that `std::size_t` can express. (Basically, anything that has the "highest bit" on, and any of the power bits on as well.)

stephenswat (approved):
This now looks good to me, thanks. :slightly_smiling_face:

PR #260, opened 2023-12-27
Synchronized Memory Resource, main branch (2023.12.27.)

krasznaa:
After #259, this is "the other" memory resource that I want to steal from the [Thrust memory resources](https://github.com/NVIDIA/cccl/tree/main/thrust/thrust/mr).

In Thrust there is an "unsynchronized" (thread-unsafe) and "synchronized" (thread-safe) version of every memory resource. They need to do that (in the end...), because [thrust::mr::memory_resource](https://github.com/NVIDIA/cccl/blob/main/thrust/thrust/mr/memory_resource.h) is templated. But since [std::pmr::memory_resource](https://en.cppreference.com/w/cpp/memory/memory_resource) is not, instead of adding a "thread-safe wrapper" around all of our currently thread-unsafe memory resources, we can also do what this PR does. Just add a single `vecmem::synchronized_memory_resource` that makes sure that all allocations and de-allocations happen one-at-a-time for its upstream memory resource.

The code is trivially simple in the end, but I think it could be useful in a couple of cases. 🤔

Note that Thrust also provides a "thread local version" of every one of its memory resources. But that I don't want to copy for our project. 🤔 `thread_local` is a bit heavy. So I want to leave it up to the frameworks using our memory resources, to deal with instantiating multiple memory resource objects for their multiple threads. As the frameworks themselves should be able to do it much more efficiently. (At least with [TBB](https://github.com/oneapi-src/oneTBB) this can be done more efficiently than with `thread_local`.)

### DIFF
--- a/core/CMakeLists.txt
+++ b/core/CMakeLists.txt
@@ -137,6 +137,9 @@ vecmem_add_library( vecmem_core core
    # Terminal memory resource.
    "src/memory/terminal_memory_resource.cpp"
    "include/vecmem/memory/terminal_memory_resource.hpp"
+   # Synchronized memory resource.
+   "src/memory/synchronized_memory_resource.cpp"
+   "include/vecmem/memory/synchronized_memory_resource.hpp"
    # Utilities.
    "include/vecmem/utils/copy.hpp"
    "include/vecmem/utils/impl/copy.ipp"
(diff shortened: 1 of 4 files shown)


stephenswat (approved):
This looks sane.

krasznaa:
Let's merge this one in at first then. :wink:

PR #265, opened 2024-02-12
Jagged Vector Copy Fix, v1.3.0-patches branch (2024.02.12.)

krasznaa:
I wanted to have a new tag with #264 included in it, but making `v1.4.0` so quickly after `v1.3.0` seemed to be a poor solution. 🤔 So I thought I'd try it out, how we may want to handle patches to older releases in the future. In case we ever get into such a working mode. 🤔 (Admittedly I was thinking about this since a while, so I took this opportunity to try this out.)

I went with the branch naming used by ROOT. But you should speak up, if you dislike this for some reason.

As you can see, the new release will be called `v1.3.1`. With the separate `v1.3.0-patches` branch, I could update the project version to be `1.3.1` already. 😉

### DIFF
--- a/core/include/vecmem/utils/impl/copy.ipp
+++ b/core/include/vecmem/utils/impl/copy.ipp
@@ -1,7 +1,7 @@
 /*
  * VecMem project, part of the ACTS project (R&D line)
  *
- * (c) 2021-2023 CERN for the benefit of the ACTS project
+ * (c) 2021-2024 CERN for the benefit of the ACTS project
  *
  * Mozilla Public License Version 2.0
  */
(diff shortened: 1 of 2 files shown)


paulgessinger:
Just for reference, we do this regularly in the ACTS core repository. 

I suppose the difference is that we tag on dedicated branches there. 

Should we discuss this tomorrow @krasznaa?

krasznaa:
After talking to @paulgessinger a bit, I renamed the "patch branch" to something that I also like a bit more, but other than that we agreed to go ahead with the simple setup that I discussed with @stephenswat yesterday. 🤔

So, let's try this for now...

PR #269, opened 2024-02-28
SoA Device Test Improvements, main branch (2024.02.28.)

krasznaa:
Skip some SoA tests on devices with no FP64 support.

The device tests with `vecmem::testing::jagged_soa_container` require the device to be able to execute FP64 operations. (Since the SoA container has a jagged vector of doubles.) Which is not the case for absolutely every SYCL device that we test. So on those these tests should just be skipped.

I considered removing the FP64 instructions from the test. But in the end decided that doing some tests with FP64 values was still worth it. To test for some possible alignment issues in the SoA code. So just skipping these tests all together on devices with no FP64 support seemed better than making the tests less thorough on all devices.

Unfortunately implementing this was a bit of a hassle. :frowning: As it turns out, one can only use `GTEST_SKIP()` in the main test body. You cannot just use that macro inside of a function that gets called as part of the test. :frowning: So I decided to add a simple `bool` return value to all language-specific functions, and let the main test body call `GTEST_SKIP()` based on those return values. :thinking:

### DIFF
--- a/tests/common/soa_device_tests.ipp
+++ b/tests/common/soa_device_tests.ipp
@@ -1,7 +1,7 @@
 /*
  * VecMem project, part of the ACTS project (R&D line)
  *
- * (c) 2023 CERN for the benefit of the ACTS project
+ * (c) 2023-2024 CERN for the benefit of the ACTS project
  *
  * Mozilla Public License Version 2.0
  */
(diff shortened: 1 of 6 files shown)


stephenswat (approved):
Looks good :+1:

PR #270, opened 2024-02-29
Copy (Test) Improvements, main branch (2024.02.29.)

krasznaa:
Managed to hit "Enter" with my pinkie a bit too early... :frowning:

This is a follow-up from #268.
  - First off, I made the "SoA copy tests" behave in the same way as the "general copy tests" do since #268. Which also allowed me to start testing the asynchronous copy types in the tests.
  - After finding a bunch of new synchronization issues, and still having problems with finding the last problem(s), I made "copy events" a little smarter.
    * Introduced the `vecmem::abstract_event::ignore()` function, to allow clients to explicitly ignore a given synchronization event in their code if they want to. (In some rare cases this can be useful.)
    * Made both `::sycl_event` and `::cuda_event` implicitly wait for their underlying events in their destructors, if the user didn't do so explicitly.
    * Since the latter (not explicitly waiting on synchronization events) can often be a bug in one's code, leading to performance degradation, introduced the `VECMEM_FAIL_ON_ASYNC_ERRORS` CMake option for building the project in a way that it would call [std::terimate()](https://en.cppreference.com/w/cpp/error/terminate) in case it detects such an error. After considering throwing an exception at first, I went with calling `std::terminate()` in the end, because throwing exceptions in destructors "cleanly" is just a lot of hassle. I still needed to do something "drastic", since the way that we build the project, with pretty aggressive symbol hiding, there is no easy way of debugging such issues in a client project easily without the code doing something "drastic". :thinking:
  - With all this in place, fixed all of the issues with asynchronous copying in the project that I could find.
    * It lead to `vecmem::copy` becoming yet a bit more complicated. In order to avoid "internal waits" in its functions as much as possible.

With all of this in place, all copy related issues are gone from my tests now... :crossed_fingers:


cuda/src/utils/cuda/async_copy.cpp
```diff
@@ -26,13 +27,39 @@ namespace {
...
+        if (m_event != nullptr) {
+            // If so, wait implicitly now.
+            VECMEM_DEBUG_MSG(1, "Asynchronous CUDA event was not waited on!");
+            wait();
+#ifdef VECMEM_FAIL_ON_ASYNC_ERRORS
+            // If the user wants to fail on asynchronous errors, do so now.
+            std::terminate();
+#endif  // VECMEM_FAIL_ON_ASYNC_ERRORS
+        }
+    }
```
krasznaa, line 45: @stephenswat, how do you feel about such a behaviour in these types? As I wrote in the PR, after some pondering, I found this as the best setup to allow debugging issues with user code. :thinking:

krasznaa (commented):
Unfortunately the code created a ["SEH exception"](https://learn.microsoft.com/en-us/cpp/cpp/structured-exception-handling-c-cpp?view=msvc-170) in the CUDA tests on Windows. I'll have to hunt that down before this would go in... :thinking:

krasznaa:
The "SEH exceptions" on Windows turned out to come from this:

https://forums.developer.nvidia.com/t/accessing-managed-memory-during-asynchronous-copies/

So I just disabled some tests on Windows...

krasznaa:
With the conclusions coming out of

https://forums.developer.nvidia.com/t/accessing-managed-memory-during-asynchronous-copies/

I considered for a bit if the `vecmem::copy` code could potentially detect when it can and can't do certain copy operations asynchronously. But in the end, that would be super difficult to do. (The "copy code" should not know what type of memory resources it is dealing with.) So let's just stay with disabling some of the tests on Windows, and then remembering this limitation in case some user stumbles upon it in the future.

PR #272, opened 2024-03-06
HIP updates: managed_memory_resource and missing header

StewMH:
Fixes compilation in the traccc alpaka HIP build.

### DIFF
--- a/core/include/vecmem/containers/impl/device_array.ipp
+++ b/core/include/vecmem/containers/impl/device_array.ipp
@@ -1,10 +1,15 @@
 /* VecMem project, part of the ACTS project (R&D line)
  *
- * (c) 2021 CERN for the benefit of the ACTS project
+ * (c) 2021-2024 CERN for the benefit of the ACTS project
  *
  * Mozilla Public License Version 2.0
  */
 #pragma once
+
+// HIP include(s)
+#if defined(__HIP_DEVICE_COMPILE__)
+#include <hip/hip_runtime.h>
+#endif
 
 // System include(s).
 #include <cassert>
--- a/core/include/vecmem/memory/impl/device_atomic_ref.ipp
+++ b/core/include/vecmem/memory/impl/device_atomic_ref.ipp
@@ -7,6 +7,11 @@
  */
 #pragma once
 
+// HIP include
+#if defined(__HIP_DEVICE_COMPILE__)
+#include <hip/hip_runtime.h>
+#endif
+
 // SYCL include(s).
 #if defined(CL_SYCL_LANGUAGE_VERSION) || defined(SYCL_LANGUAGE_VERSION)
 #include <CL/sycl.hpp>
--- a/core/include/vecmem/memory/impl/device_atomic_ref.ipp
+++ b/core/include/vecmem/memory/impl/device_atomic_ref.ipp
@@ -1,7 +1,7 @@
 /*
  * VecMem project, part of the ACTS project (R&D line)
  *
- * (c) 2022-2023 CERN for the benefit of the ACTS project
+ * (c) 2022-2024 CERN for the benefit of the ACTS project
  *
  * Mozilla Public License Version 2.0
  */
--- a/hip/include/vecmem/memory/hip/managed_memory_resource.hpp
+++ b/hip/include/vecmem/memory/hip/managed_memory_resource.hpp
@@ -1,7 +1,7 @@
 /*
  * VecMem project, part of the ACTS project (R&D line)
  *
- * (c) 2021-2023 CERN for the benefit of the ACTS project
+ * (c) 2021-2024 CERN for the benefit of the ACTS project
  *
  * Mozilla Public License Version 2.0
  */
--- a/hip/include/vecmem/memory/hip/managed_memory_resource.hpp
+++ b/hip/include/vecmem/memory/hip/managed_memory_resource.hpp
@@ -1,0 +1,53 @@
+/*
+ * VecMem project, part of the ACTS project (R&D line)
+ *
+ * (c) 2021-2023 CERN for the benefit of the ACTS project
+ *
+ * Mozilla Public License Version 2.0
+ */
+
+#pragma once
+
+// Local include(s).
+#include "vecmem/memory/memory_resource.hpp"
+#include "vecmem/vecmem_hip_export.hpp"
+
+namespace vecmem::hip {
+
+/**
+ * @brief Memory resource that wraps managed HIP allocation.
+ *
+ * This is an allocator-type memory resource that allocates managed HIP
+ * memory, which is accessible directly to devices as well as to the host.
+ */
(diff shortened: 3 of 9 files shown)


hip/include/vecmem/memory/hip/managed_memory_resource.hpp
```diff
@@ -0,0 +1,53 @@
+/*
+ * VecMem project, part of the ACTS project (R&D line)
+ *
+ * (c) 2021-2024 CERN for the benefit of the ACTS project
```
krasznaa, line 4: There is no one true way with this, but in this project we set the "starting date" in the copyright statements to be the year when the file was created. (Of course when we rename, or largely copy a file, the definition is more gray that this...)

So in this case, I'd vote to just have `2024` here by itself. :wink:

tests/hip/test_hip_memory_resources.cpp
```diff
@@ -11,24 +11,30 @@
 #include "../common/memory_resource_test_host_accessible.hpp"
 #include "vecmem/memory/hip/device_memory_resource.hpp"
 #include "vecmem/memory/hip/host_memory_resource.hpp"
+#include "vecmem/memory/hip/managed_memory_resource.hpp"
```
krasznaa, line 14: Please update the copyright statement. :wink:

core/include/vecmem/memory/impl/device_atomic_ref.ipp
```diff
@@ -7,6 +7,11 @@
  */
 #pragma once
 
+// HIP include
```
krasznaa, line 10: Please update the copyright statement. :wink:

krasznaa (changes requested):
I was quite mystified as to why the unit tests of this project would not have shown this issue before. But then I found:

https://github.com/acts-project/vecmem/blob/main/tests/hip/test_hip_containers_kernels.hip#L8-L10

In the test where we exercise `vecmem::device_atomic_ref` in HIP device code, we already include `<hip/hip_runtime.h>` to make `assert(...)` work. So the code was/is blind to `device_atomic_ref.h` needing this include as well. :frowning:

Was HIP managed memory introduced with ROCm 5? I remember having looked for it back when I first wrote this code with ROCm 4. But I didn't find such functionality back then. :thinking: Or couse ROCm 4 is not even installed on any of my machines anymore, so I couldn't even double-check just now if this was indeed the case or not. :confused:

You should fix the formatting error(s) flagged by the CI, update the copyright statements, and this is good to go from my side. :+1:

krasznaa:
Actually... :thinking: Could you please do this?

```diff
diff --git a/core/include/vecmem/memory/impl/atomic.ipp b/core/include/vecmem/memory/impl/atomic.ipp
index 1aebc28..0008511 100644
--- a/core/include/vecmem/memory/impl/atomic.ipp
+++ b/core/include/vecmem/memory/impl/atomic.ipp
@@ -1,12 +1,17 @@
 /*
  * VecMem project, part of the ACTS project (R&D line)
  *
- * (c) 2021 CERN for the benefit of the ACTS project
+ * (c) 2021-2024 CERN for the benefit of the ACTS project
  *
  * Mozilla Public License Version 2.0
  */
 #pragma once
 
+// HIP include(s).
+#if defined(__HIP_DEVICE_COMPILE__)
+#include <hip/hip_runtime.h>
+#endif
+
 // SYCL include(s).
 #if defined(CL_SYCL_LANGUAGE_VERSION) || defined(SYCL_LANGUAGE_VERSION)
 #include <CL/sycl.hpp>
diff --git a/core/include/vecmem/memory/impl/device_atomic_ref.ipp b/core/include/vecmem/memory/impl/device_atomic_ref.ipp
index 021c648..0b3fe7e 100644
--- a/core/include/vecmem/memory/impl/device_atomic_ref.ipp
+++ b/core/include/vecmem/memory/impl/device_atomic_ref.ipp
@@ -1,12 +1,17 @@
 /*
  * VecMem project, part of the ACTS project (R&D line)
  *
- * (c) 2022-2023 CERN for the benefit of the ACTS project
+ * (c) 2022-2024 CERN for the benefit of the ACTS project
  *
  * Mozilla Public License Version 2.0
  */
 #pragma once
 
+// HIP include(s).
+#if defined(__HIP_DEVICE_COMPILE__)
+#include <hip/hip_runtime.h>
+#endif
+
 // SYCL include(s).
 #if defined(CL_SYCL_LANGUAGE_VERSION) || defined(SYCL_LANGUAGE_VERSION)
 #include <CL/sycl.hpp>
diff --git a/tests/hip/test_hip_containers_kernels.hip b/tests/hip/test_hip_containers_kernels.hip
index e3240e8..5cb83df 100644
--- a/tests/hip/test_hip_containers_kernels.hip
+++ b/tests/hip/test_hip_containers_kernels.hip
@@ -1,14 +1,10 @@
 /* VecMem project, part of the ACTS project (R&D line)
  *
- * (c) 2021-2023 CERN for the benefit of the ACTS project
+ * (c) 2021-2024 CERN for the benefit of the ACTS project
  *
  * Mozilla Public License Version 2.0
  */
 
-// HIP include(s). Note that this needs to come first, as it affects how
-// other headers later on include/see system headers like <cassert>.
-#include <hip/hip_runtime.h>
-
 // Local include(s).
 #include "../../hip/src/utils/hip_error_handling.hpp"
 #include "test_hip_containers_kernels.hpp"
@@ -21,6 +17,9 @@
 #include "vecmem/memory/device_atomic_ref.hpp"
 #include "vecmem/utils/tuple.hpp"
 
+// HIP include(s).
+#include <hip/hip_runtime.h>
+
 /// Kernel performing a linear transformation using the vector helper types
 __global__ void linearTransformKernel(
     vecmem::data::vector_view<const int> constants,
```

We should make our tests more sensitive to such errors. And also fix `atomic.ipp`. :wink:

StewMH:
Thanks Attila. I couldn't get your patch to apply but I've hopefully done something similar.

core/include/vecmem/containers/impl/device_array.ipp
```diff
@@ -1,11 +1,16 @@
...
- * (c) 2021 CERN for the benefit of the ACTS project
+ * (c) 2021-2024 CERN for the benefit of the ACTS project
  *
  * Mozilla Public License Version 2.0
  */
 #pragma once
 
+// HIP include(s)
+#if defined(__HIP_DEVICE_COMPILE__)
+#include <hip/hip_runtime.h>
```
krasznaa, line 11: This was a good catch! :thinking:

Though now I fear that we're missing such statements in a bunch of other `.ipp` files as well. :frowning: (Whatever is used on the device, and uses `assert(...)`. Which is a lot of this project's code.)

But well, the tests of the project now build successfully, so let's just stay with this for now.

krasznaa (approved):
After a small additional tweak, I'm happy with this.

PR #274, opened 2024-05-04
Add ability to reserve bulk space in device vector

stephenswat:
A very common pattern in massively parallel computing is to reserve space for an entire thread block at once and to insert those values later. This commit adds a `reserve` and `reserve_unsafe` method to `device_vector`, which allows clients to do this.

The semantics of `reserve` are fully compliant with the C++ standard starting from C++20; C++ standards below that will work just fine. Any program which calls this function and compiles should be well formed, as defined in P0593R6 and P2590R2.

The `reserve_unsafe` method does the same, but _technically_ leaves the program in an ill-defined state. Note that this form of ill-defined memory has been used for donkeys years and is actually safe. It's just technically not standards compliant.

### DIFF
--- a/core/include/vecmem/containers/impl/device_vector.ipp
+++ b/core/include/vecmem/containers/impl/device_vector.ipp
@@ -276,7 +276,7 @@ VECMEM_HOST_AND_DEVICE auto device_vector<TYPE>::bulk_append_implicit(
     // This can only be done on a resizable vector.
     assert(m_size != nullptr);
 
-    static_assert(details::is_implicit_lifetime_v<TYPE>,
+    static_assert(details::is_implicit_lifetime<TYPE>::value,
                   "Type `TYPE` in `device_vector<T>::bulk_append_implicit` is "
                   "not an implicit lifetype type, so slots cannot be safely "
                   "reserved. Note that the definition of implicit lifetimes "
@@ -387,7 +387,7 @@ VECMEM_HOST_AND_DEVICE void device_vector<TYPE>::resize_implicit(
     // This can only be done on a resizable vector.
     assert(m_size != nullptr);
 
-    static_assert(details::is_implicit_lifetime_v<TYPE>,
+    static_assert(details::is_implicit_lifetime<TYPE>::value,
                   "Type `TYPE` in `device_vector<T>::resize_implicit` is not "
                   "an implicit lifetype type, so slots cannot be safely "
                   "reserved. Note that the definition of implicit lifetimes "
--- a/core/include/vecmem/containers/impl/device_vector.ipp
+++ b/core/include/vecmem/containers/impl/device_vector.ipp
@@ -11,6 +11,8 @@
 
 // System include(s).
 #include <cassert>
+#include <memory>
+#include <type_traits>
 
 namespace vecmem {
 
@@ -226,6 +228,86 @@ VECMEM_HOST_AND_DEVICE auto device_vector<TYPE>::push_back(
 }
 
 template <typename TYPE>
+VECMEM_HOST_AND_DEVICE auto device_vector<TYPE>::bulk_append(size_type n)
+    -> size_type {
+    // This can only be done on a resizable vector.
+    assert(m_size != nullptr);
+
+    static_assert(std::is_default_constructible<TYPE>::value,
+                  "Type `T` in `device_vector<T>::bulk_append` is not a "
+                  "default-constructible type.");
+
+    device_atomic_ref<size_type> asize(*m_size);
+    const size_type index = asize.fetch_add(n);
+    assert((index + n) <= m_capacity);
+
+    for (size_type i = 0; i < n; ++i) {
+        construct(index + i, value_type());
+    }
+
+    return index;
+}
+
+template <typename TYPE>
+VECMEM_HOST_AND_DEVICE auto device_vector<TYPE>::bulk_append(size_type n,
+                                                             const_reference v)
+    -> size_type {
+    // This can only be done on a resizable vector.
+    assert(m_size != nullptr);
+
+    static_assert(std::is_copy_constructible<TYPE>::value,
+                  "Type `TYPE` in device_vector<T>::bulk_append(SIZE, TYPE)` "
+                  "must be copy-constructible.");
+
+    device_atomic_ref<size_type> asize(*m_size);
+    const size_type index = asize.fetch_add(n);
+    assert((index + n) <= m_capacity);
+
+    for (size_type i = 0; i < n; ++i) {
+        construct(index + i, value_type(v));
+    }
+
+    return index;
+}
+
+template <typename TYPE>
+VECMEM_HOST_AND_DEVICE auto device_vector<TYPE>::bulk_append_implicit(
+    size_type n) -> size_type {
+    // This can only be done on a resizable vector.
+    assert(m_size != nullptr);
+
+    static_assert(details::is_implicit_lifetime_v<TYPE>,
+                  "Type `TYPE` in `device_vector<T>::bulk_append_implicit` is "
+                  "not an implicit lifetype type, so slots cannot be safely "
+                  "reserved. Note that the definition of implicit lifetimes "
+                  "differs between C++<=17, C++20, and C++>=23.");
+
+    device_atomic_ref<size_type> asize(*m_size);
+    const size_type index = asize.fetch_add(n);
+    assert((index + n) <= m_capacity);
+
+#if defined(__cpp_lib_start_lifetime_as) && \
+    __cpp_lib_start_lifetime_as >= 202207L
+    std::start_lifetime_as_array<TYPE>(m_ptr[index], n);
+#endif
+
+    return index;
+}
+
+template <typename TYPE>
+VECMEM_HOST_AND_DEVICE auto device_vector<TYPE>::bulk_append_implicit_unsafe(
+    size_type n) -> size_type {
+    // This can only be done on a resizable vector.
+    assert(m_size != nullptr);
+
+    device_atomic_ref<size_type> asize(*m_size);
+    const size_type index = asize.fetch_add(n);
+    assert((index + n) <= m_capacity);
+
+    return index;
+}
+
+template <typename TYPE>
 VECMEM_HOST_AND_DEVICE auto device_vector<TYPE>::pop_back() -> size_type {
 
     // This can only be done on a resizable vector.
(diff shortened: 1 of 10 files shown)


core/include/vecmem/containers/impl/device_vector.ipp
```diff
@@ -225,6 +227,50 @@ VECMEM_HOST_AND_DEVICE auto device_vector<TYPE>::push_back(
...
+#else
+    static_assert(details::is_implicit_lifetime_v<TYPE>,
+                  "Type `T` in `device_vector<T>::reserve` is not an implicit "
+                  "lifetype type (according to a non-C++23-compliant "
+                  "definition), so slots cannot be safely reserved.");
+#endif
+
+    device_atomic_ref<size_type> asize(*m_size);
+    const size_type index = asize.fetch_add(n);
+    assert((index + n - 1) < m_capacity);
```
krasznaa, line 251: Why not

```suggestion
    assert((index + n) <= m_capacity);
```

? The compiler may have a slightly easier time with that one...

core/include/vecmem/containers/impl/device_vector.ipp
```diff
@@ -225,6 +227,50 @@ VECMEM_HOST_AND_DEVICE auto device_vector<TYPE>::push_back(
...
+    __cpp_lib_is_implicit_lifetime >= 202302L
+    static_assert(std::is_implicit_lifetyime_v<TYPE>,
+                  "Type `T` in `device_vector<T>::reserve` is not an implicit "
+                  "lifetype type (according to the C++23-compliant "
+                  "definition), so slots cannot be safely reserved.");
+#else
+    static_assert(details::is_implicit_lifetime_v<TYPE>,
+                  "Type `T` in `device_vector<T>::reserve` is not an implicit "
+                  "lifetype type (according to a non-C++23-compliant "
+                  "definition), so slots cannot be safely reserved.");
```
krasznaa, line 246: We don't actually require this for device types at the moment. :thinking: This is why the `device_vector<T>::construct(...)` function exists. I was expecting that you'd make use of it here.

Making it possible to use types that have some simple constructors (for setting default values on variables for instance), would be highly desirable in my mind.

krasznaa (changes requested):
I agree that we need these sort of functions. But I don't like the naming proposal. :frowning:

We are trying to mimic the behaviour of `std::vector` here. But you'd never start using elements of an `std::vector` that have been "reserved", but not "created" yet.

I'd suggest a different naming scheme for these functions. Let's say:

```c++
size_type enlarge(size_type n, const_reference value);
size_type enlarge_unsafe(size_type n);
```

These are indeed names that `std::vector` doesn't use. But they are after all meant to do stuff that `std::vector` doesn't do. :thinking:

core/include/vecmem/containers/impl/device_vector.ipp
stephenswat: Right, but there's a problem with that strategy: if you want to reserve, say, 1024 elements for a whole block of threads, the constructor approach requires the calling thread to loop through a constructor 1024 times which could be potentially super slow.

The advantage of using implicit lifetime types like this is that the object is in a well-defined (but unknown) stage the second the memory is allocated, so serial construction is not necessary.

The unsafe method allows the client to construct the objects in parallel for performance.

In principle I think having a set of functions which enlarges a vector by $n$ elements and the constructs them in place also has a use, but it's a different use from this.

core/include/vecmem/containers/impl/device_vector.ipp
stephenswat: That's fair!

stephenswat:
Agreed the naming could be improved; I see a few functions that we could implement which would need a name:

1. Implicitly insert $n$ elements, _not_ calling their constructors for performance reasons, but the objects are well-formed because they are implicit lifetime objects.
2. Implicitly insert $n$ elements, _not_ calling their constructors for performance reasons, in an unsafe way because they are _not_ implicit lifetime objects.
3. Insert $n$ elements, default-constructing each of them.
4. Insert $n$ elements, copy-constructing them from a given value.

The first two are implemented, I agree the second two could also be useful; but they need names.

stephenswat:
Could be something like `bulk_insert_implicit`, `bulk_insert_implicit_unsafe`, `bulk_insert` (with 2 overloads)?

krasznaa:
As long as we're workshopping the names, I'd prefer `bulk_append_implicit`, etc. Since these will only ever add new elements at the end of the vector.

stephenswat:
I have renamed the functions and added the additional two functions I described above.

tests/core/test_core_device_containers.cpp
```diff
@@ -179,45 +179,71 @@ TEST_F(core_device_container_test, resizable_vector_buffer) {
     EXPECT_EQ(device_vector.back(), 15);
 
     device_vector.assign(20, 123);
-    EXPECT_EQ(device_vector.size(), static_cast<vector_size_type>(20));
```
krasznaa, line 182: The size check should still remain, no? :confused:

core/include/vecmem/containers/impl/device_vector.ipp
krasznaa: :+1:

core/include/vecmem/utils/type_traits.hpp
```diff
@@ -113,5 +113,22 @@ auto max(T&& t, Ts&&... ts) {
...
+    : std::disjunction<
+          std::is_scalar<B>, std::is_array<B>, std::is_aggregate<B>,
+          std::conjunction<
+              std::is_trivially_destructible<B>,
+              std::disjunction<std::is_trivially_default_constructible<B>,
+                               std::is_trivially_copy_constructible<B>,
+                               std::is_trivially_move_constructible<B>>>> {};
+#else
+template <class B>
+struct is_implicit_lifetime : std::false_type {};
```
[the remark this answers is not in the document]
stephenswat: I changed this to `std::true_type`; if someone wants to compile this pre-C++17 I guess they can.

tests/core/test_core_device_containers.cpp
stephenswat: Yes, not sure why this happened.

stephenswat:
I think this should be good to go now. I'd suggest we get this and #275 in and tag a release.

core/include/vecmem/containers/impl/device_vector.ipp
```diff
@@ -225,6 +227,91 @@ VECMEM_HOST_AND_DEVICE auto device_vector<TYPE>::push_back(
...
 
+template <typename TYPE>
+VECMEM_HOST_AND_DEVICE auto device_vector<TYPE>::bulk_append(size_type n)
+    -> size_type {
+    // This can only be done on a resizable vector.
+    assert(m_size != nullptr);
+
+    static_assert(std::is_default_constructible<TYPE>::value,
+                  "Type `T` in `device_vector<T>::bulk_append` is not a "
+                  "default-constructible type.");
```
krasznaa, line 238: What happens when we instantiate `vecmem::device_vector` in a non-default-constructible type? 🤔 Even if the user doesn't try to use this function, I think (based on what we saw in #275) that this assertion will activate.

I fear that we can only do this using SFINAE...

To make sure, please add a unit test with a type that's not default constructible. 🤔

core/include/vecmem/containers/impl/device_vector.ipp
```diff
@@ -225,6 +227,91 @@ VECMEM_HOST_AND_DEVICE auto device_vector<TYPE>::push_back(
...
+                  "an implicit "
+                  "lifetype type (according to the C++23-compliant "
+                  "definition), so slots cannot be safely reserved.");
+#else
+    static_assert(details::is_implicit_lifetime_v<TYPE>,
+                  "Type `T` in `device_vector<T>::bulk_append_implicit` is not "
+                  "an implicit "
+                  "lifetype type (according to a non-C++23-compliant "
+                  "definition), so slots cannot be safely reserved.");
+#endif
```
krasznaa, line 288: I fear these would need to be converted into SFINAE as well. I'd do it by making `vecmem::details::is_implicit_lifetime_v` "behave correctly` with all C\+\+ standards. So that the preprocessor statements would be in the declaration of that trait class, and not in its users.

krasznaa (changes requested):
I'm still very much on board, but we need to make sure that the new functions don't break previously working functionality.

core/include/vecmem/containers/impl/device_vector.ipp
stephenswat: This is a very fun edge case in C++ that was actually resolved in C++23 with P2593R1. Before that, `static_assert(false)` was _always_ ill-formed, even in templates that are not instantiated. After P2593R1, `static_assert(false)` is only ill-formed if the template is actually instantiated, and can also be hidden in `if constexpr`. For dependent `static_asserts` like this one, it's actually fine: the error is only emitted when the template is actually instantiated.

core/include/vecmem/utils/type_traits.hpp
```diff
@@ -113,5 +113,45 @@ auto max(T&& t, Ts&&... ts) {
...
+/// trait is available.
+///
+/// @warning On pre-C++17 translation units, this type trait is always assumed
+/// to be true.
+///
+/// @{
+#if defined(__cpp_lib_is_implicit_lifetime) && \
+    __cpp_lib_is_implicit_lifetime >= 202302L
+template <class B>
+struct is_implicit_lifetime : std::is_implicit_lifetime<TYPE> {};
```
krasznaa, line 132: Not sure why you use `B` in all these places, but:

```suggestion
struct is_implicit_lifetime : std::is_implicit_lifetime<B> {};
```

Or better yet, why not

```c++
template <class B>
using is_implicit_lifetime = std::is_implicit_lifetime<B>;
```

?

krasznaa (changes requested):
The unit test looks good.

The typo just underlines that we'll need to add C\+\+20 CI tests sooner than later. 🤔

core/include/vecmem/utils/type_traits.hpp
stephenswat: Well spotted, `is_implicit_lifetime` is actually a C++23 feature so not even a C++20 CI build would have caught that. :smile:

krasznaa:
Are you still tweaking it, or is it ready? 😕

stephenswat:
This is now ready to go. :+1:

krasznaa:
Good. Then let me just put some unit tests on top of it, which I invite you to check. 😉

krasznaa (approved):
I'm happy with it myself, but @stephenswat should have a look as well.

stephenswat:
Looks sane. :+1:

stephenswat:
Let's finally get this in. :smile:

core/include/vecmem/containers/impl/device_vector.ipp
```diff
@@ -299,6 +381,34 @@ VECMEM_HOST_AND_DEVICE void device_vector<TYPE>::resize(size_type new_size,
     asize.store(new_size);
 }
 
+template <typename TYPE>
+VECMEM_HOST_AND_DEVICE void device_vector<TYPE>::resize_implicit(
+    size_type new_size) {
+    // This can only be done on a resizable vector.
+    assert(m_size != nullptr);
+
+    static_assert(details::is_implicit_lifetime_v<TYPE>,
```
krasznaa, line 390: Unfortunately I forgot about testing this function. I believe this line would also produce a warning with CUDA 10.4 right now.

```suggestion
    static_assert(details::is_implicit_lifetime<TYPE>::value,
```

I should have checked for this in my previous commit. :frowning:

stephenswat: What? Why? The CUDA 10.4 problem is that the `_v` predicates in the standard library were only introduced in C++17, but we define this one ourselves.

PR #275, opened 2024-06-06
Increase the robustness of `device_atomic_ref`

stephenswat:
While working on https://github.com/acts-project/traccc/pull/595, I found out that the vecmem implementation of atomic CAS is fundamentally broken on CUDA platforms :worried:. Currently, the functionality is `compare_exchange_strong` is broken because it relies on the CUDA `atomicCAS` builtin which functions fundamentally differently from the C++ STL version of the equivalent code. Indeed, the C++ version returns true on a succesful swap and false otherwise. The CUDA version always returns the old value. As such, if the old value is false-like, e.g. 0, the `compare_exchange_strong` function will always appear to fail, even if it succeeded. This commit fixes the above issue.

I also removed the backup implementation of CAS as it was not atomic in any way and was basically lying to users about working atomically :worried:.

### DIFF
--- a/core/include/vecmem/memory/impl/device_atomic_ref.ipp
+++ b/core/include/vecmem/memory/impl/device_atomic_ref.ipp
@@ -42,7 +42,6 @@ template <>
 struct sycl_address_space<device_address_space::local> {
     static constexpr cl::sycl::access::address_space add =
         cl::sycl::access::address_space::local_space;
-
     template <typename T>
     using ptr_t = cl::sycl::local_ptr<T>;
 };
@@ -73,7 +72,62 @@ struct sycl_address_space<device_address_space::local> {
         ARG1, ARG2)
 #endif
 
+#if defined __has_builtin
+#if __has_builtin(__atomic_load_n)
+#define VECMEM_HAVE_BUILTIN_ATOMIC_LOAD
+#endif
+#if __has_builtin(__atomic_store_n)
+#define VECMEM_HAVE_BUILTIN_ATOMIC_STORE
+#endif
+#if __has_builtin(__atomic_exchange_n)
+#define VECMEM_HAVE_BUILTIN_ATOMIC_EXCHANGE
+#endif
+#if __has_builtin(__atomic_compare_exchange_n)
+#define VECMEM_HAVE_BUILTIN_ATOMIC_COMPARE_EXCHANGE
+#endif
+#if __has_builtin(__atomic_fetch_add)
+#define VECMEM_HAVE_BUILTIN_ATOMIC_FETCH_ADD
+#endif
+#if __has_builtin(__atomic_fetch_sub)
+#define VECMEM_HAVE_BUILTIN_ATOMIC_FETCH_SUB
+#endif
+#if __has_builtin(__atomic_fetch_and)
+#define VECMEM_HAVE_BUILTIN_ATOMIC_FETCH_AND
+#endif
+#if __has_builtin(__atomic_fetch_xor)
+#define VECMEM_HAVE_BUILTIN_ATOMIC_FETCH_XOR
+#endif
+#if __has_builtin(__atomic_fetch_or)
+#define VECMEM_HAVE_BUILTIN_ATOMIC_FETCH_OR
+#endif
+#endif
+
 namespace vecmem {
+
+#if defined(__ATOMIC_RELAXED) && defined(__ATOMIC_CONSUME) && \
+    defined(__ATOMIC_ACQUIRE) && defined(__ATOMIC_RELEASE) && \
+    defined(__ATOMIC_ACQ_REL) && defined(__ATOMIC_SEQ_CST)
+constexpr int __memorder_vecmem_to_builtin(memory_order o) {
+    switch (o) {
+        case memory_order::relaxed:
+            return __ATOMIC_RELAXED;
+        case memory_order::consume:
+            return __ATOMIC_CONSUME;
+        case memory_order::acquire:
+            return __ATOMIC_ACQUIRE;
+        case memory_order::release:
+            return __ATOMIC_RELEASE;
+        case memory_order::acq_rel:
+            return __ATOMIC_ACQ_REL;
+        case memory_order::seq_cst:
+            return __ATOMIC_SEQ_CST;
+        default:
+            assert(false);
+            return 0;
+    }
+}
+#define VECMEM_HAVE_MEMORDER_DEFINITIONS
+#endif
 
 template <typename T, device_address_space address>
 VECMEM_HOST_AND_DEVICE device_atomic_ref<T, address>::device_atomic_ref(
@@ -90,13 +144,13 @@ VECMEM_HOST_AND_DEVICE auto device_atomic_ref<T, address>::operator=(
     value_type data) const -> value_type {
 
     store(data);
-    return load();
+    return data;
 }
 
 template <typename T, device_address_space address>
 VECMEM_HOST_AND_DEVICE void device_atomic_ref<T, address>::store(
-    value_type data, memory_order) const {
-
+    value_type data, memory_order order) const {
+    (void)order;
 #if (defined(__CUDA_ARCH__) || defined(__HIP_DEVICE_COMPILE__)) && \
     (!(defined(SYCL_LANGUAGE_VERSION) || defined(CL_SYCL_LANGUAGE_VERSION)))
     volatile pointer addr = m_ptr;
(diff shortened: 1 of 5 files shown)


krasznaa:
As you noticed, one cannot just use `static_assert(...)` in a "non-templated" function of a templated class. As soon as the class is instantiated, the assertion kicks in. It doesn't only happen when the function is called. :frowning:

If you forego the removal of the naive, non-atomic implementation, then I'll be happy to get this fix in. But I'd rather not open the can of worms with how `vecmem::device_atomic_ref` should behave on the host before C\+\+20. :frowning: I don't think that will lead us anywhere useful.

stephenswat:
My bad, I had hoped that any of the templates would have been on the function, not on the class, but sadly not. Anyway there will be a bit more work to do here anyway, so I'll come up with a more comprehensive solution.

krasznaa:
Note that I've been thinking for a while now about introducing `cuda::atomic_ref` in this code. :thinking: Similar to how we use `sycl::atomic_ref`, "under the right circumstances" the code should just use `cuda::atomic_ref`, as is. You should check if you could make that happen.

stephenswat:
Okay, the scope of this PR has grown a little bit to fix a whole bunch of other issues with the atomic references. Also adds additional compile-time checks on the functionality of atomic references as well as runtime tests.

core/include/vecmem/concepts/atomic_ref.hpp
```diff
@@ -0,0 +1,62 @@
+/*
+ * VecMem project, part of the ACTS project (R&D line)
+ *
+ * (c) 2022-2023 CERN for the benefit of the ACTS project
```
krasznaa, line 4:
```suggestion
 * (c) 2024 CERN for the benefit of the ACTS project
```

core/include/vecmem/concepts/atomic_ref.hpp
```diff
@@ -0,0 +1,62 @@
...
+ *
+ * Mozilla Public License Version 2.0
+ */
+#pragma once
+
+#if __cpp_concepts >= 201907L
+#include <concepts>
+
+#include "vecmem/memory/device_atomic_ref.hpp"
```
krasznaa, line 13: You should move `vecmem::memory_order` into a separate header. As I don't like this circular include.

core/include/vecmem/concepts/atomic_ref.hpp
```diff
@@ -0,0 +1,62 @@
...
+#endif
+
+namespace vecmem::concepts {
+#if __cpp_concepts >= 201907L
+/*
+ * Concept to verify that a type functions as an atomic reference.
+ */
```
krasznaa, line 20: Please make it into a Doxygen comment.

core/include/vecmem/memory/device_atomic_ref.hpp
```diff
@@ -213,4 +213,17 @@ class device_atomic_ref {
...
+static_assert(
+    vecmem::concepts::atomic_ref<vecmem::device_atomic_ref<uint32_t> >,
+    "Atomic reference on uint32_t is incompletely defined.");
+static_assert(
+    vecmem::concepts::atomic_ref<vecmem::device_atomic_ref<uint64_t> >,
+    "Atomic reference on uint64_t is incompletely defined.");
+static_assert(
+    vecmem::concepts::atomic_ref<vecmem::device_atomic_ref<std::size_t> >,
+    "Atomic reference on std::size_t is incompletely defined.");
+#endif
```
krasznaa, line 227: :thinking: Is this the place to do this check? I'm honestly not sure...

core/include/vecmem/memory/impl/device_atomic_ref.ipp
```diff
@@ -116,137 +187,216 @@ VECMEM_HOST_AND_DEVICE auto device_atomic_ref<T, address>::load(
...
     (!(defined(SYCL_LANGUAGE_VERSION) || defined(CL_SYCL_LANGUAGE_VERSION)))
     return atomicSub(m_ptr, data);
 #elif defined(CL_SYCL_LANGUAGE_VERSION) || defined(SYCL_LANGUAGE_VERSION)
     return __VECMEM_SYCL_ATOMIC_CALL1(fetch_sub, m_ptr, data);
+#elif defined(VECMEM_HAVE_BUILTIN_ATOMIC_SUB) && \
+    defined(VECMEM_HAVE_MEMORDER_DEFINITIONS)
+    return __atomic_fetch_sub(m_ptr, data, __memorder_vecmem_to_builtin(order));
+#elif defined(VECMEM_HAVE_BUILTIN_ATOMIC_ADD) && \
+    defined(VECMEM_HAVE_MEMORDER_DEFINITIONS)
+    return __atomic_fetch_sub(m_ptr, -data,
```
krasznaa, line 325:
```suggestion
    return __atomic_fetch_add(m_ptr, -data,
```

core/include/vecmem/memory/impl/device_atomic_ref.ipp
```diff
@@ -7,6 +7,15 @@
...
 #pragma once
 
+#include <cassert>
+
+// MSVC include
+#if defined(_MSC_VER)
+// Not how you're supposed to do it, but whatever.
+#include <windows.h>
+#include <winnt.h>
+#endif
```
krasznaa, line 17: I'm also not a fan of this. :frowning: Why not try to use the MSVC compiler intrinsics instead? :thinking: That's also not great of course, but with this we include the whole universe on Windows. :frowning:

core/include/vecmem/memory/impl/device_atomic_ref.ipp
```diff
@@ -116,137 +187,216 @@ VECMEM_HOST_AND_DEVICE auto device_atomic_ref<T, address>::load(
...
     return __VECMEM_SYCL_ATOMIC_CALL0(load, m_ptr);
+#elif defined(VECMEM_HAVE_BUILTIN_ATOMIC_LOAD) && \
+    defined(VECMEM_HAVE_MEMORDER_DEFINITIONS)
+    return __atomic_load_n(m_ptr, __memorder_vecmem_to_builtin(order));
+#elif defined(__clang__) || defined(__GNUC__) || defined(__GNUG__) || \
+    defined(__CUDACC__)
+    return __atomic_load_n(m_ptr, __memorder_vecmem_to_builtin(order));
+#elif defined(_MSC_VER)
+    // Probably pollutes your caches but who cares about MSVC anyway.
+    return InterlockedAdd(m_ptr, 0);
```
krasznaa, line 198: The following would probably behave better:

```suggestion
    return InterlockedOr(m_ptr, static_cast<T>(0));
```

krasznaa (changes requested):
I'm absolutely on board with making this code better. I'm very happy that you're looking into it.

Please fix up all the remaining issues, and then I'll be willing to push in this macro-hell. :thinking: But after that, I'll absolutely want to clean this up.
  - Instead of doing preprocessor magic everywhere, I'll want to have a few different classes called let's say `vecmem::details::cuda::atomic_ref`, `vecmem::details::win32::atomic_ref`, etc.
    * The idea being that that would hopefully result in more understandable compiler errors when some preprocessor decision inevitably goes wrong in the future.
  - At that point we could push the implementation of the "host versions" into `.cpp` files, to avoid exposing the user to let's say `<windows.h>`. We only want to provide this class for a short list of primitive types anyway.

If you want to give that setup a try, I won't stop you. :wink: But as I started, I'm willing to let the code in with this design as well as a first step. (After the actual issues have been fixed.)

core/include/vecmem/memory/device_atomic_ref.hpp
stephenswat: I think it's fine here. You get better diagnostics if you check for concepts in a template argument, but we don't really use these atomic references for that (yet?). Still, I think checking for these is very important if we're going to sometimes use `sycl::atomic_ref`, etc.

core/include/vecmem/memory/impl/device_atomic_ref.ipp
stephenswat: Good catch. :smile:

core/include/vecmem/memory/impl/device_atomic_ref.ipp
stephenswat: Sure, that will work. :slightly_smiling_face:

stephenswat:
Okay, let's see what the MSVC CI thinks of this.

stephenswat:
Okay so MSVC doesn't support atomics on unsigned integers. :laughing:

stephenswat:
Or at least, I cannot find them.

stephenswat:
There's an implementation here, in the MSVC STL: https://github.com/microsoft/STL/blob/main/stl/inc/atomic

But I cannot be arsed to reimplement all of that...

stephenswat:
Someone explain to me how this commit breaks the synchronized memory resource on release builds in MSVC and in those builds alone.

stephenswat:
Ah, of course.

Including `intrin.h` breaks locks and mutexes.

https://github.com/stephenswat/vecmem/actions/runs/9440139525/job/25999074005

:clown_face: :clown_face: :clown_face: :clown_face: :clown_face: :clown_face: :clown_face: :clown_face: :clown_face: :clown_face: :clown_face: :clown_face: :clown_face: :clown_face: :clown_face: :clown_face: :clown_face: :clown_face: :clown_face: :clown_face: :clown_face: :clown_face: :clown_face: :clown_face: :clown_face: :clown_face: :clown_face: :clown_face: :clown_face: :clown_face: :clown_face: :clown_face: :clown_face: :clown_face: :clown_face: :clown_face: :clown_face: :clown_face: :clown_face: :clown_face: :clown_face: :clown_face: :clown_face: :clown_face: :clown_face: :clown_face: :clown_face:

stephenswat:
I got rid of the MSVC intrinsics and replaced them by a non-atomic implementation of CAS, but at least now there is only one point of fake atomicity in this code.

core/include/vecmem/memory/impl/device_atomic_ref.ipp
```diff
@@ -116,137 +187,254 @@ VECMEM_HOST_AND_DEVICE auto device_atomic_ref<T, address>::load(
     return value;
 #elif defined(CL_SYCL_LANGUAGE_VERSION) || defined(SYCL_LANGUAGE_VERSION)
     return __VECMEM_SYCL_ATOMIC_CALL0(load, m_ptr);
+#elif defined(VECMEM_HAVE_BUILTIN_ATOMIC_LOAD) && \
+    defined(VECMEM_HAVE_MEMORDER_DEFINITIONS)
+    return __atomic_load_n(m_ptr, __memorder_vecmem_to_builtin(order));
```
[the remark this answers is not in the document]
krasznaa: To be sorted out in a follow-up PR.

krasznaa (approved):
Let's get this in. Cleanup/improvements to come afterwards.

PR #276, opened 2024-06-10
Add concept for atomic references

stephenswat:
This commit adds a new concept for atomic references and, in order to do so, moves the definition of `memory_order` into a separate file.

### DIFF
--- a/core/include/vecmem/concepts/atomic_ref.hpp
+++ b/core/include/vecmem/concepts/atomic_ref.hpp
@@ -1,0 +1,62 @@
+/*
+ * VecMem project, part of the ACTS project (R&D line)
+ *
+ * (c) 2024 CERN for the benefit of the ACTS project
+ *
+ * Mozilla Public License Version 2.0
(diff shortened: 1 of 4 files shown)


stephenswat:
Trying to figure out where exactly the error in #275 is coming from.

stephenswat:
Okay now I'm _really_ confused.

krasznaa (approved):
I'll run one more test on Windows once I'm home, but after that will push this in as-is. :thinking:

krasznaa:
So, local tests are still running fine. And while searching a bit more, I found:
  - https://github.com/google/googletest/issues/4556
  - https://github.com/actions/runner-images/issues/10004

(I also opened https://github.com/orgs/community/discussions/128019, but apparently that was not the correct place for this.)

So, it appears to be a quite fresh issue. Let's hope that it quickly gets solved. As I'd really hate to disable some of the tests on Windows just because of this. :thinking:

PR #279, opened 2024-06-11
Windows CI Fix, main branch (2024.06.11.)

krasznaa:
As discussed in https://github.com/actions/runner-images/issues/10004, CI jobs on Windows started misbehaving recently. This is a work-around for the issue, adding `_DISABLE_CONSTEXPR_MUTEX_CONSTRUCTOR` as a definition to `vecmem::core`. (To allow it to use [std::mutex](https://en.cppreference.com/w/cpp/thread/mutex) correctly on Windows.)

### DIFF
--- a/core/CMakeLists.txt
+++ b/core/CMakeLists.txt
@@ -1,6 +1,6 @@
 # VecMem project, part of the ACTS project (R&D line)
 #
-# (c) 2021-2023 CERN for the benefit of the ACTS project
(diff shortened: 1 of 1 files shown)


stephenswat (approved):
I'd have preferred something based on detecting whether we are in a CI job or not, but this is fine. Seems like this flag only affects MSVC anyway.

PR #280, opened 2024-06-11
Introduce doxygen-awesome-css, main branch (2024.06.08.)

krasznaa:
I was told about [doxygen-awesome-css](https://github.com/jothepro/doxygen-awesome-css) last week by @EdwardMoyse. And then he even went ahead with introducing it into [atlas/athena](https://gitlab.cern.ch/atlas/athena) in https://gitlab.cern.ch/atlas/athena/-/merge_requests/72075. So I thought I should do it for our repository as well. :wink:

What this setup brings is to switch from this sort of documentation pages:

image
image
image

, to a look like:

image
image
image

Now... I personally use a dark theme, that's why all the pages show up like this. But they should by default adapt to the viewer's preference I believe. Plus, the claim is that the pages should show up a lot better on phones/tablets. Though this latter I didn't try so far.

I'm actually open to a discussion about this. How do you guys feel about this sort of a look? :thinking:

### DIFF
--- a/.github/workflows/pages.yml
+++ b/.github/workflows/pages.yml
@@ -1,6 +1,6 @@
 # VecMem project, part of the ACTS project (R&D line)
 #
-# (c) 2023 CERN for the benefit of the ACTS project
+# (c) 2023-2024 CERN for the benefit of the ACTS project
 #
 # Mozilla Public License Version 2.0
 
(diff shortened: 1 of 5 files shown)


stephenswat:
It's certainly better than the default style, although I do prefer the RTD (Sphinx?) format over this.

beomki-yeo:
I am also fine with the change. The default one looks quite old school

PR #281, opened 2024-06-12
Revert "Windows CI Fix, main branch (2024.06.11.)"

krasznaa:
Reverts acts-project/vecmem#279.

The claim in https://github.com/actions/runner-images/issues/10004 is that the GitHub Windows runners are now back to not needing this workaround anymore. So, let's see...

### DIFF
--- a/core/CMakeLists.txt
+++ b/core/CMakeLists.txt
@@ -316,12 +316,6 @@ if( VECMEM_HAVE_STD_ALIGNED_ALLOC )
       PRIVATE VECMEM_HAVE_STD_ALIGNED_ALLOC )
 endif()
 
-# This is a (hopefully) temporary workaround for fixing the CI builds on
-# Windows. See https://github.com/actions/runner-images/issues/10004 for more
-# details.
-target_compile_definitions( vecmem_core
-   PRIVATE _DISABLE_CONSTEXPR_MUTEX_CONSTRUCTOR )
-
 # Test the public headers of vecmem::core.
 if( BUILD_TESTING AND VECMEM_BUILD_TESTING )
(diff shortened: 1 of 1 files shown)


stephenswat:
Apparently not. :worried:

krasznaa:
I guess we can keep it open to have an easy way of testing this going forward. :thinking:

core/CMakeLists.txt
```diff
@@ -1,6 +1,6 @@
 # VecMem project, part of the ACTS project (R&D line)
 #
-# (c) 2021-2024 CERN for the benefit of the ACTS project
+# (c) 2021-2023 CERN for the benefit of the ACTS project
```
stephenswat, line 3: Actually, changing this year back to 2023 makes no sense at all. Should remain at 2024.

stephenswat (changes requested):

core/CMakeLists.txt
krasznaa: :confused: I used GitHub's built-in PR reversal feature. That undoes everything from the previous PR.

stephenswat (approved):

krasznaa:
Are these latest failures fixed by #287? I thought that was for a slightly different issue. 🤔 I don't remember such missing glibc symbols as errors so far.

stephenswat:
> Are these latest failures fixed by https://github.com/acts-project/vecmem/pull/287? I thought that was for a slightly different issue. 🤔 I don't remember such missing glibc symbols as errors so far.

Yeah, the glibc issues are fixed by #287, until Microsoft removes support for node 10 fully.

stephenswat:
Looks good. :+1:

PR #283, opened 2024-06-17
Separate SYCL compilation details from vecmem core

stephenswat:
The SYCL build system for vecmem suffers from a pretty nasty problem where it has to be compiled using a SYCL-capable compiler. This means that an installation of SYCL cannot be conveniently classified by the single compiler used to build it. Fixing this will take a few more PR's but a more pressing issue is that the vecmem core library is actually infected with compilation details from the SYCL library. More specifically, the core library carries information about the function used to print in SYCL code, leading to the fact that the core library can change if the SYCL library is enabled, disabled, or is compiled with a different compiler which is very undesirable. This commit migrates all the SYCL-related logic from the core library to the SYCL library to resolve this issue.


krasznaa (commented):
I accept that the code should / could be improved. But this is not the way. This forces clients to link against `vecmem::sycl`. Even if they only need `vecmem::core`. Since for instance SYCL device code would never need to link against `vecmem::sycl`. Just as how CUDA device code has no use for `vecmem::cuda`.

The only good way I see out of this is similar to what Thrust, Vc, etc. do for their configuration. Where they require the client to use some CMake function that they supply, which would set up all the flags for using their code, in the user's environment. I was just shying away from such a large change so far.

stephenswat:
Actually, as is usually the case, the best solution would be to minimize the role of CMake in all of this and just rely on the preprocessor to solve this problem for us. I'll update the PR later.

krasznaa:
Selecting the correct print function via preprocessor logic is going to result in much more fragile code unfortunately. :frowning: No, I'm not in favour of that.

krasznaa:
Replaced by #294.

PR #284, opened 2024-07-01
Deduplicate `empty.sycl` file

stephenswat:
The current setup uses `try_compile` on the source file `empty.sycl` twice, which leads to the following error when using Ninja, enabling SYCL, and having the tests enabled:

```
error: sycl/cmake/empty.sycl is defined as an output multiple times
```

This commit takes the path of least resistance and clones the file into `empty_ubsan.sycl` so that these compile tests no longer collide.


krasznaa (approved):
Having now looked at it myself as well, unfortunately I don't see a better solution either. Though to be honest, I don't even fully understand why Ninja gets confused here. Since `empty.sycl` is not a generated file. We use it as input to multiple `try_compile(...)` commands. I consider it a bug in CMake that this leads to an error with Ninja...

But yeah, path of least resistance...

PR #285, opened 2024-07-01
Make atomic reference constructor explicit

stephenswat:
Right now, the constructor `device_atomic_ref(T&)` is implicit, which does not match the definition provided in C++20. While this seems harmless, it can break code that relies on this implicit conversion if it switches to using `std::atomic_ref`, which vecmem may decide to do if C++20 is enabled. This commit makes the constructor explicit so it matches the definitions in the STL.


krasznaa (approved):
:+1:

PR #286, opened 2024-07-16
 Mark copy events as `nodiscard`

stephenswat:
Currently, there is no compile-time mechanism to ensure that copy objects are properly handled, i.e. either waited for or ignored. This commit adds the `nodiscard` attribute to methods of the copy objects which return events to ensure that the compiler will issue a warning if the object is ignored.

### DIFF
--- a/core/include/vecmem/utils/attributes.hpp
+++ b/core/include/vecmem/utils/attributes.hpp
@@ -1,0 +1,19 @@
+/*
+ * VecMem project, part of the ACTS project (R&D line)
+ *
+ * (c) 2021-2023 CERN for the benefit of the ACTS project
+ *
+ * Mozilla Public License Version 2.0
+ */
+
+#pragma once
+
+#ifdef __has_cpp_attribute
+#if __has_cpp_attribute(nodiscard) >= 201603L
+#define VECMEM_NODISCARD [[nodiscard]]
+#else
+#define VECMEM_NODISCARD
+#endif
+#else
+#define VECMEM_NODISCARD
+#endif
--- a/benchmarks/core/benchmark_copy.cpp
+++ b/benchmarks/core/benchmark_copy.cpp
@@ -50,11 +50,11 @@ void jaggedVectorUnknownHtoDCopy(::benchmark::State& state) {
     const data::jagged_vector_data<int> source_data = get_data(source);
     // Create the "destination buffer".
     data::jagged_vector_buffer<int> dest(sizes, host_mr);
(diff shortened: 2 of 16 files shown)


krasznaa (commented):
:thinking: I considered doing this a while ago. But decided that it would be too painful for the users to always have to deal with adding `wait()` and `ignore()` calls, for **all** `vecmem::copy` variants. Even all the synchronous ones.

Since forgetting to add these calls is "only" a performance penalty in the current setup, I introduced `VECMEM_FAIL_ON_ASYNC_ERRORS` in #270.

Now... I do not think that that solution is perfect. Not at all. But I'm very afraid of making the UI "too painful" to use. And I fear that using `[[nodiscard]]` on all of these functions is very painful. :frowning:

stephenswat:
Hmm, not necessarily. I modified traccc using this PR and it only came to 20 files changed, 86 insertions(+), 69 deletions(-). So not that painful, really.

stephenswat:
> Since forgetting to add these calls is "only" a performance penalty in the current setup, I introduced VECMEM_FAIL_ON_ASYNC_ERRORS in https://github.com/acts-project/vecmem/pull/270.

Unfortunately, they can also be runtime errors in SYCL code if you use unordered streams.

krasznaa:
Unfortunately there are some HIP issues after all. 😦 (Some missing return value handling was left in.)

core/include/vecmem/utils/attributes.hpp
```diff
@@ -0,0 +1,19 @@
...
+
+#ifdef __has_cpp_attribute
+#if __has_cpp_attribute(nodiscard) >= 201603L
+#define VECMEM_NODISCARD [[nodiscard]]
+#else
+#define VECMEM_NODISCARD
+#endif
+#else
+#define VECMEM_NODISCARD
+#endif
```
krasznaa, line 19: Hmm... Now that I checked this because of another reason...

Why did you bother with all of this here? We already require clients to use a minimum of C\+\+17 with their host compiler to be able to use the core functionality of this project. (That of memory resources.) So we can just use `[[nodiscard]]` directly. Really no need for macros here. :thinking:

krasznaa (changes requested):
Plus, the HIP changes...

stephenswat:
Updated to pass the CI. @krasznaa

krasznaa (approved):
Fine, let's go with the unnecessary `VECMEM_NODISCARD`. We can remove it later on... 🤔

PR #287, opened 2024-07-16
Fix checkout support in CI for Ubuntu 18.04

stephenswat:
The combination of us using the past-EoL Ubuntu 18.04 and Microsoft being a bit gung-ho in their changes to GitHub Actions means that Node16, and by extension checkouts@v3 is no longer supported. Since Ubuntu 18.04 doesn't support the newer Node20, we are stuck. Thankfully, we can set the `ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION` environment variable which is project to keep the CI running until they remove support for Node16 fully.

See #286 for the related CI failure, and https://github.blog/changelog/2024-03-07-github-actions-all-actions-will-run-on-node20-instead-of-node16-by-default/ for the GitHub blog post about this.


krasznaa (approved):
Sure. I trust that you researched the issue sufficiently.

krasznaa (changes requested):
Well, I should've checked. Unfortunately the CI doesn't seem to like these lines. :frowning: Something needs to be done differently as it seems.

stephenswat:
There we go, now it's happy.

krasznaa (approved):
👍

Issue #288, opened 2024-08-03
Atomic Failure on Windows with CUDA, main branch (2024.08.03.)


krasznaa:
Unfortunately #275 seems to have introduced a failure in a setup that's not too easy to test in the CI. 😦 While building the latest code on Windows, with CUDA available, I bump into:

```
...
[build]   tmpxft_000062a0_00000000-7_test_cuda_jagged_vector_view_kernels.cudafe1.cpp
[build]   tmpxft_0000606c_00000000-7_test_cuda_edm_kernels.cudafe1.cpp
[build] C:\Users\krasz\ATLAS\vecmem\vecmem\core\include\vecmem/memory/impl/device_atomic_ref.ipp(187): error C3861: '__memorder_vecmem_to_builtin': identifier not found [C:\Users\krasz\ATLAS\vecmem\vecmem\build\tests\cuda\vecmem_test_cuda.vcxproj]
[build]   C:\Users\krasz\ATLAS\vecmem\vecmem\core\include\vecmem/memory/impl/device_atomic_ref.ipp(187): note: the template instantiation context (the oldest one first) is
[build]   C:\Users\krasz\ATLAS\vecmem\vecmem\tests\cuda\../common/jagged_soa_container_helpers.hpp(21): note: see reference to class template instantiation 'vecmem::testing::jagged_soa_interface<vecmem::edm::device<vecmem::edm::schema<vecmem::edm::type::scalar<int>,vecmem::edm::type::vector<float>,vecmem::edm::type::jagged_vector<double>,vecmem::edm::type::scalar<float>,vecmem::edm::type::jagged_vector<int>,vecmem::edm::type::vector<int>>>>' being compiled
[build]   C:\Users\krasz\ATLAS\vecmem\vecmem\tests\common\jagged_soa_container.hpp(18): note: see reference to class template instantiation 'vecmem::edm::device<vecmem::edm::schema<vecmem::edm::type::scalar<int>,vecmem::edm::type::vector<float>,vecmem::edm::type::jagged_vector<double>,vecmem::edm::type::scalar<float>,vecmem::edm::type::jagged_vector<int>,vecmem::edm::type::vector<int>>>' being compiled
[...]

PR #291, opened 2024-08-08
Atomic Reference Reorganization, main branch (2024.08.08.)

krasznaa:
This is a proposal for fixing the MSVC+CUDA compilation issue described in #288. As discussed already in #275, I wanted to split the implementation of `vecmem::device_atomic_ref` into multiple classes since a while. #288 was the trigger to make it happen.

This PR introduces the following new classes:
  - `vecmem::dummy_device_atomic_ref`: A fallback implementation, for host code that's not covered by something better. Does not actually do atomic operations.
  - `vecmem::posix_device_atomic_ref`: An implementation using GCC / Clang built-ins for atomic operation support in host code.
  - `vecmem::cuda::device_atomic_ref`: An implementation using the basic CUDA C atomic operations.
  - `vecmem::hip::device_atomic_ref`: The same as `vecmem::cuda::device_atomic_ref` (is a typedef of the latter), with an additional include statement needed by HIP.
  - `vecmem::sycl::builtin_device_atomic_ref`: A typedef using `cl::sycl::atomic_ref`.
  - `vecmem::sycl::custom_device_atomic_ref`: A full implementation of an atomic reference, using "low level" SYCL atomic operations.

The `vecmem/memory/device_atomic_ref.hpp` file now just chooses between these different implementations, based on various criteria, and sets up `vecmem::device_atomic_ref` as a typedef to one of the above types, or to [std::atomic_ref](https://en.cppreference.com/w/cpp/atomic/atomic_ref) with host code when using C\+\+20.

As we discussed with @stephenswat, in the end the `VECMEM_SUPPORT_POSIX_ATOMIC_REF` test, which currently runs only during the build, will need to become something that `vecmem-config.cmake` would re-run for the host compiler in the context of the client project's build. But since some other flags will need to be moved to behave like that as well, I decided to do that in a future PR. And in this one I just figure out whether the host compiler supports `vecmem::posix_device_atomic_ref`, during the build of vecmem itself.

I am still working on the code, but wanted to expose it to @stephenswat at this point.


core/CMakeLists.txt
```diff
@@ -316,11 +327,27 @@ if( VECMEM_HAVE_STD_ALIGNED_ALLOC )
...
+endif()
+unset( CMAKE_REQUIRED_INCLUDES )
+
 # Test the public headers of vecmem::core.
 if( BUILD_TESTING AND VECMEM_BUILD_TESTING )
-   file( GLOB_RECURSE vecmem_core_public_headers
+   file( GLOB vecmem_core_public_headers
       RELATIVE "${CMAKE_CURRENT_SOURCE_DIR}/include"
-      "include/*/*.hpp" )
+      "include/vecmem/*/*.hpp" )
```
krasznaa, line 350: Not every single header may be seen by the host compiler anymore. So the test now only looks at "top level" headers. I'm not 100% satisfied with this though. :thinking:

core/CMakeLists.txt
```diff
@@ -316,11 +327,27 @@ if( VECMEM_HAVE_STD_ALIGNED_ALLOC )
...
+   #include <vecmem/memory/details/posix_device_atomic_ref.hpp>
+   int main() {
+      int foo = 0;
+      vecmem::posix_device_atomic_ref<int> ref{foo};
+      return 0;
+   }
+   " VECMEM_SUPPORT_POSIX_ATOMIC_REF )
+if( VECMEM_SUPPORT_POSIX_ATOMIC_REF )
+   target_compile_definitions( vecmem_core
+      PUBLIC VECMEM_SUPPORT_POSIX_ATOMIC_REF )
```
krasznaa, line 342: As mentioned in the description of the PR, this test is currently made during the build of vecmem. In a follow-up PR I'll add this test separately in `vecmem-config.cmake` as well. Along with a couple of additional tests.

core/include/vecmem/memory/details/device_address_space.hpp
```diff
@@ -0,0 +1,14 @@
...
+ * (c) 2022-2024 CERN for the benefit of the ACTS project
+ *
+ * Mozilla Public License Version 2.0
+ */
+#pragma once
+
+namespace vecmem {
+
+/// Custom definition for the device memory adress space
+enum class device_address_space { global = 0, local = 1 };
```
krasznaa, line 12: It might make more sense to have this file outside of the `details/` namespace in the end. I was probably a bit too eager with putting it in here. :thinking:

core/include/vecmem/memory/details/hip_device_atomic_ref.hpp
```diff
@@ -0,0 +1,33 @@
...
+// HIP include(s).
+#include <hip/hip_runtime.h>
+
+// Set up __VECMEM_THREADFENCE correctly for the vecmem::hip::device_atomic_ref
+// code.
+#ifdef __HIP_DEVICE_COMPILE__
+#define __VECMEM_THREADFENCE __threadfence()
+#else
+#define __VECMEM_THREADFENCE
+#endif  // defined(__HIP_DEVICE_COMPILE__)
```
krasznaa, line 19: I thought I would be able to avoid all such tricky logic with this code layout, but as it turns out, no...

You see, once I made `vecmem::posix_device_atomic_ref` only work in host code, it turned out that at least HIP (but I think also CUDA...) was instantiating that type for at least some steps in building `__global__` functions.

But once I made sure that the CUDA and HIP compilers would always see `vecmem::cuda::device_atomic_ref` and `vecmem::hip::device_atomic_ref`, even when building host code, I ran into complaints from `__threadfence()` being declared a "device only" function, so we're not allowed to call it from a `VECMEM_HOST_AND_DEVICE` function.

When hiding the `__threadfence()` call behind `__CUDA_ARCH__` and `__HIP_DEVICE_COMPILE__`, this compiler complaint disappears. Apparently `nvcc` doesn't check for this type of consistency when only doing the device code compilation stage of the build. :confused:

Long story short, apparently this part of our code has to stay inherently fragile after all.
  - As long as we want to allow `vecmem::device_vector` to work in host code (and we absolutely do!), `vecmem::cuda::device_atomic_ref` has to use `VECMEM_HOST_AND_DEVICE` on its functions.
  - `__threadfence()` is a "device only" function, so calling it from a `VECMEM_HOST_AND_DEVICE` function is not actually allowed.

So yeah, I don't see a clean way out of this fragility at the moment. Not without exposing a completely different class during the host and device code compilation stages to `nvcc`. Which is also not any less fragile... :frowning:

core/include/vecmem/memory/details/sycl_custom_device_atomic_ref.hpp
```diff
@@ -0,0 +1,125 @@
...
+        std::is_integral<value_type>::value,
+        "vecmem::sycl::custom_device_atomic_ref only accepts built-in "
+        "integral types");
+
+    /// @}
+
+    /// Constructor, with a pointer to the managed variable
+    explicit custom_device_atomic_ref(reference ref);
+    /// Copy constructor
+    custom_device_atomic_ref(const custom_device_atomic_ref& parent);
```
krasznaa, line 57: I removed all uses of the `VECMEM_HOST_AND_DEVICE` macro from this class. As a SYCL compiler (which is the only one that should ever see this code) would never care about that macro anyway. :thinking:

krasznaa (commented):
I wanted to point out a few things for you Stephen, since I'm aware of a few weaknesses myself... :thinking:

stephenswat (commented):
Looks good, minor comments and questions.

core/include/vecmem/memory/details/device_address_space.hpp
krasznaa: 👍

core/include/vecmem/memory/details/hip_device_atomic_ref.hpp
krasznaa: Switching (back) to `__CUDA_ARCH__` and `__HIP_DEVICE_COMPILE__` made this macro unnecessary once again... 🤔

core/include/vecmem/memory/device_atomic_ref.hpp
```diff
@@ -1,202 +1,117 @@
...
-                       details::sycl_address_space<address>::add>;
+using device_atomic_ref = sycl::custom_device_atomic_ref<T, address>;
 
 }  // namespace vecmem
 
-#elif ((!defined(__CUDA_ARCH__)) && (!defined(__HIP_DEVICE_COMPILE__)) && \
-       (!defined(CL_SYCL_LANGUAGE_VERSION)) &&                            \
-       (!defined(SYCL_LANGUAGE_VERSION)) && __cpp_lib_atomic_ref)
+#endif  // defined(VECMEM_HAVE_SYCL_ATOMIC_REF)
+#elif defined(__CUDACC__)
```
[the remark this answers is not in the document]
krasznaa: As discussed, against what I was expecting, `__CUDA_ARCH__` and `__HIP_DEVICE_COMPILE__` do result in the expected behaviour after all. Even though `nvcc` and `hipcc` do absolute encounter the host versions of these classes during the build.

Hey, the code works with using those macros, and it allowed me to remove the ugly `__VECMEM_THREADFENCE` hack, so let's not question the universe...

core/include/vecmem/memory/impl/dummy_device_atomic_ref.ipp
```diff
@@ -0,0 +1,147 @@
...
+
+    // This is **NOT** a sane implementation of CAS!
+    value_type old = *m_ptr;
+    if (old == expected) {
+        *m_ptr = desired;
+        return true;
+    } else {
+        expected = old;
+        return false;
+    }
```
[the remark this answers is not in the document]
krasznaa: Unfortunately using a deprecation warning on this function (which is the only one I could think of) triggers a bunch of warnings in the Windows build at the moment. 😦 So I really don't want to make this into a warning right now. 🤔

In the future, when I feel masochistic, I'll want to take the Windows specific code that you tried to develop in #275, and make it finally work. At that point we'll be able to put actual warnings on the dummy class, as it will really only be used on platforms that we really-really didn't think of. 🤔

stephenswat (approved):

PR #294, opened 2024-09-09
Flag Determination Redesign, main branch (2024.09.09.)

krasznaa:
As discussed in some previous PRs, the build system making decisions about how clients should be using some of the templated code in `vecmem::core` and `vecmem::sycl`, just based on the capabilities of the compilers seen by this project's build, is not technically correct.

While clients of a built/installed vecmem project are not meant to use fundamentally different compilers to compile/link against the libraries, some flexibility **is** allowed. (And with some code rewrite, even more flexibility could potentially be achieved.)

Here I introduced two helper functions, `vecmem_setup_core(...)` and `vecmem_setup_sycl(...)`, which would take care of figuring out what `INTERFACE` flags should be set up on `vecmem::core` and `vecmem::sycl`. This way these functions can be called by both [core/CMakeLists.txt](core/CMakeLists.txt) / [sycl/CMakeLists.txt](sycl/CMakeLists.txt) and any subsequent `find_package(vecmem)` calls as well.

In this new setup, when a client calls `find_package(vecmem)` in an environment in which a SYCL compiler is not available, they would see something like the following:

```
[bash][Celeborn]:vecmem > cmake -DCMAKE_PREFIX_PATH=$PWD/install-intel/usr/local -S test -B build-test
-- The CXX compiler identification is GNU 13.2.0
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Found vecmem: /home/krasznaa/ATLAS/projects/vecmem/install-intel/usr/local/lib/cmake/vecmem-1.8.0/vecmem-config.cmake (found version "1.8.0")
-- Performing Test VECMEM_SUPPORT_POSIX_ATOMIC_REF
-- Performing Test VECMEM_SUPPORT_POSIX_ATOMIC_REF - Success
-- Looking for a SYCL compiler
-- Looking for a SYCL compiler - NOTFOUND
-- Configuring done (4.1s)
-- Generating done (0.0s)
-- Build files have been written to: /home/krasznaa/ATLAS/projects/vecmem/build-test
[bash][Celeborn]:vecmem >
```

While in an environment in which a SYCL compile is available, this changes to:

```
[bash][Celeborn]:vecmem > cmake -DCMAKE_PREFIX_PATH=$PWD/install-intel/usr/local -S test -B build-test
-- The CXX compiler identification is IntelLLVM 2024.2.1
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /home/krasznaa/software/intel/oneapi-2024.2.1/compiler/2024.2/bin/compiler/clang++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Found vecmem: /home/krasznaa/ATLAS/projects/vecmem/install-intel/usr/local/lib/cmake/vecmem-1.8.0/vecmem-config.cmake (found version "1.8.0")
-- Performing Test VECMEM_SUPPORT_POSIX_ATOMIC_REF
-- Performing Test VECMEM_SUPPORT_POSIX_ATOMIC_REF - Success
-- Looking for a SYCL compiler
-- Looking for a SYCL compiler - /home/krasznaa/software/intel/oneapi-2024.2.1/compiler/2024.2/bin/compiler/clang++
-- The SYCL compiler identification is IntelLLVM 2024.2.1
-- Check for working SYCL compiler: /home/krasznaa/software/intel/oneapi-2024.2.1/compiler/2024.2/bin/compiler/clang++
-- Check for working SYCL compiler: /home/krasznaa/software/intel/oneapi-2024.2.1/compiler/2024.2/bin/compiler/clang++ - works
-- Performing Test VECMEM_HAVE_SYCL_EXT_ONEAPI_PRINTF
-- Performing Test VECMEM_HAVE_SYCL_EXT_ONEAPI_PRINTF - Success
-- Performing Test VECMEM_HAVE_SYCL_ONEAPI_PRINTF
-- Performing Test VECMEM_HAVE_SYCL_ONEAPI_PRINTF - Failed
-- Performing Test VECMEM_HAVE_SYCL_ATOMIC_REF
-- Performing Test VECMEM_HAVE_SYCL_ATOMIC_REF - Success
-- Performing Test VECMEM_HAVE_SYCL_LOCAL_ACCESSOR
-- Performing Test VECMEM_HAVE_SYCL_LOCAL_ACCESSOR - Success
-- Configuring done (86.5s)
-- Generating done (0.0s)
-- Build files have been written to: /home/krasznaa/ATLAS/projects/vecmem/build-test
[bash][Celeborn]:vecmem >
```

Very similar to what [Vc](https://github.com/VcDevel/Vc) does. Whether or not that's a good thing... 🤔

(I was testing the build in WSL, to keep making sure that Windows compatibility would stay alive. Hence the horrendous configuration times. 😛)

To make all of this happen relatively conveniently, I also introduced the `vecmem_check_sycl_source_compiles(...)` function, with an API mimicking [check_cxx_source_compiles(...)](https://cmake.org/cmake/help/latest/module/CheckCXXSourceCompiles.html). Converting most of the checks to using it, from the existing `vecmem_check_sycl_code_compiles(...)` calls.


CMakeLists.txt
```diff
@@ -12,6 +12,10 @@ project( vecmem VERSION 1.8.0 LANGUAGES CXX )
 include( GNUInstallDirs )
 include( CTest )
 
+# Set the location of the installed CMake files.
+set( CMAKE_INSTALL_CMAKEDIR
+   "${CMAKE_INSTALL_LIBDIR}/cmake/vecmem-${PROJECT_VERSION}" )
```
paulgessinger, line 17: Is this conventional?

cmake/vecmem-check-sycl-source-compiles.cmake
```diff
@@ -0,0 +1,36 @@
+# VecMem project, part of the ACTS project (R&D line)
+#
+# (c) 2024 CERN for the benefit of the ACTS project
+#
+# Mozilla Public License Version 2.0
+
+# CMake version requirement.
+cmake_minimum_required( VERSION 3.14 )
```
paulgessinger, line 8: One thing I found with these is that they reset policy switches, so my assumption was that it's best practice to avoid them. 

Did you consider this?

core/cmake/vecmem-setup-core.cmake
```diff
@@ -0,0 +1,109 @@
...
+   endif()
+   unset( CMAKE_REQUIRED_INCLUDES )
+
+   # Check if a SYCL compiler is available.
+   vecmem_check_language( SYCL )
+
+   # Set the SYCL definitions only if the build system knows about SYCL.
+   if( CMAKE_SYCL_COMPILER )
+
+      # Test which SYCL printf function(s) is/are available.
```
paulgessinger, line 45: Ouff, that's quite a mess. But I guess it has to be done.

sycl/CMakeLists.txt
```diff
@@ -39,8 +40,11 @@ target_link_libraries( vecmem_sycl PUBLIC vecmem::core )
...
    CXX_VISIBILITY_PRESET  "hidden"
    SYCL_VISIBILITY_PRESET "hidden" )
-vecmem_check_sycl_code_compiles( VECMEM_HAVE_SYCL_VISIBILITY_MS_COMPAT
-   "${CMAKE_CURRENT_SOURCE_DIR}/cmake/empty.sycl"
+vecmem_check_sycl_source_compiles( "
+   #include <CL/sycl.hpp>
+   int main() { return 0; }
+   "
+   VECMEM_HAVE_SYCL_VISIBILITY_MS_COMPAT
    CMAKE_FLAGS -DCOMPILE_DEFINITIONS=-fvisibility-ms-compat )
```
paulgessinger, line 48: Not part of the diff, but why does this have to go through `CMAKE_FLAGS`?

sycl/cmake/vecmem-setup-sycl.cmake
```diff
@@ -0,0 +1,45 @@
...
+include( vecmem-check-sycl-source-compiles )
+
+# CMake include(s).
+include( CheckCXXSourceCompiles )
+
+# Helper function setting up the public definitions for vecmem::sycl
+#
+# Usage: vecmem_setup_sycl( vecmem::sycl )
+#
+function( vecmem_setup_sycl libName )
```
paulgessinger, line 18: Isn't it a bit strange to have a function (there's a similar one for the core lib) that accepts a library target but realistically is meant to be used with one single library that it applies to?

They get used in the installed config, but that doesn't require the library as an argument, does it?

paulgessinger (commented):
So if the initial build saw a sycl compiler and the downstream build does not see one, it's not an error? Is there any message about this.

I guess it would just be more confusing to try to store the sycl compiler location in the installed config and then trying to recover that one?

krasznaa:
> So if the initial build saw a sycl compiler and the downstream build does not see one, it's not an error? Is there any message about this.
> 
> I guess it would just be more confusing to try to store the sycl compiler location in the installed config and then trying to recover that one?

Generally, to use either `vecmem::core` or `vecmem::sycl`, you just need a C\+\+ compiler. So yes, in principle a client project of let's say traccc could well use its SYCL library without having SYCL set up for its own build. (It's also true for vecmem with `vecmem::sycl`, it's just that you couldn't do a whole lot with just the ability to allocate different types of SYCL memory blocks.)

In both `vecmem::core` and `vecmem::sycl` there are a couple of headers with template code that can behave "in a SYCL specific way". The flags that the CMake code is trying to figure out here, are for making those template classes work correctly.

Long story short, as you can do this right now as well, with this PR in place you'll still be able to compile pure C\+\+ code against the installed vecmem libraries. Whether or not `vecmem::sycl` was built.

CMakeLists.txt
krasznaa: This is the convention I went with for all of the R&D projects.

From my experience `lib/cmake` is the most used location for any project that doesn't need a `share/` directory otherwise. For projects that do use `share/`, the CMake files usually sit under `share/cmake/`. (Contrary to **all of that**, Gaudi and ATLAS use just `cmake/`. Go figure. :stuck_out_tongue:)

As for the final directory name, I like the idea of being able to have different versions of vecmem sitting side-by-side. That's why I don't just have `${CMAKE_INSTALL_LIBDIR}/cmake/vecmem` here. :thinking: (It's not something that we would necessarily ever do, but this at least allows for it.)

See https://cmake.org/cmake/help/latest/command/find_package.html#config-mode-search-procedure for all the possibilities that CMake understands. This here is one of the valid installation locations.

cmake/vecmem-check-sycl-source-compiles.cmake
krasznaa: It indeed resets earlier requirements. But just for the scope of this file. The file including this one should not get its requirements updated. At least I **think** how it's supposed to happen.

This file absolutely needs `3.14` as a minimum. So as long as that minimum doesn't leak into the code including this one, I'm happy with overriding the caller's requirement for just the scope of this file. :thinking:

core/cmake/vecmem-setup-core.cmake
krasznaa: This is the logic we had in place for a number of years by now. Unfortunately `printf(...)` is a oneAPI extension to the SYCL standard. The logging functionality would need to be seriously redesigned to make it work with [sycl::stream](https://www.intel.com/content/www/us/en/docs/oneapi/optimization-guide-gpu/2023-0/doing-io-in-the-kernel.html). Since I was able to use for instance [AdaptiveCpp](https://github.com/AdaptiveCpp/AdaptiveCpp), which does not support `printf(...)`, by just not turning on any debug messages, it was fine so far to only get debugging output with oneAPI.

sycl/CMakeLists.txt
krasznaa: Because [try_compile(...)](https://cmake.org/cmake/help/latest/command/try_compile.html) is a pain. :frowning: See the description of `CMAKE_FLAGS` under:

https://cmake.org/cmake/help/latest/command/try_compile.html#options

So I just kept the "standard CMake interface" for the SYCL function as well. As broken as it may be...

PR #296, opened 2024-09-25
SoA Proxy, main branch (2024.09.25.)

krasznaa:
With the final straw being https://github.com/acts-project/traccc/pull/712, this is an attempt at providing "proxies" for the SoA containers that would allow an "AoS-like" access to the containers in the client code.

The code requires the user to create the same type of interface struct/class for their container, as they did so far. But now that interface is no longer only used providing a user-friendly API for the containers, but also to provide the same API for the proxy objects. Allowing user code to interchangeably use
  - `container.variable().at(i)` or;
  - `container.at(i).variable()`

semantics to access the same variable.

We discussed about this already with @stephenswat, and I'm absolutely not convinced that I did the correct thing here, but I made the proxies behave in the following way:
  - "constant proxies" keep their variables by value for the scalar and vector variable types;
  - "non-const proxies" keep their variables by reference.

My idea being that this way we could force the loading of all variables of a given container in a single place in device code, when we are only reading from those variables. But as @stephenswat pointed out, I might very well be doing something silly with this. And should maybe just use references in the proxies everywhere. :thinking:

The device code, and a lot of additional testing code is still to come, but I wanted to showcase already how I want to implement this feature.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -86,6 +86,8 @@ if( VECMEM_BUILD_DOCS )
       "${CMAKE_CURRENT_SOURCE_DIR}/cuda/include"
       "${CMAKE_CURRENT_SOURCE_DIR}/hip/include"
       "${CMAKE_CURRENT_SOURCE_DIR}/sycl/include" )
+   set( DOXYGEN_PREDEFINED
+      "__cplusplus=201700L" )
    set( DOXYGEN_EXCLUDE
       "${CMAKE_CURRENT_SOURCE_DIR}/tests"
       "${CMAKE_CURRENT_SOURCE_DIR}/benchmarks"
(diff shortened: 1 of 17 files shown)


krasznaa:
I now tried this code with https://github.com/acts-project/traccc/pull/712. @stephenswat was correct. It's more performant to just use references everywhere in the proxies, and leave it up to the GPU to figure out what it would pull into registers.

As noted on https://github.com/acts-project/traccc/pull/712, it turns out that memory copies with `vecmem::edm` are not quite as amazing as I was hoping for. :frowning: Specifically, `vecmem::edm::host` to `vecmem::edm::buffer` H->H copies (in order to collect all columns into a single allocation for the H->D copy) are not nearly as quick as I thought. :confused:

I intend to study that more deeply in some new benchmarks that I'll add in a separate PR. :thinking:

krasznaa:
After having slept on it a few times already, I'm still pretty happy with this code. So let's go forward with it. 😉

PR #300, opened 2024-10-27
Introduce vecmem::metal, main branch (2024.10.27.)

krasznaa:
What should one do on a weekend if not write some useless code... 🤔

I wanted to see for a while now whether [Metal](https://developer.apple.com/metal/) was the least bit compatible with how this project works. And after a weekend of goofing around, the answer is eeehhh... 😕

Metal is not a C(\+\+) language/library to begin with, but an Objective-C one. Apple wrote a wrapper itself for calling Metal library functions from C\+\+ code directly, but I didn't get the impression from the documentation that I found, that it would be particularly supported. So I went for interfacing with Metal by adding Objective-C\+\+ source files to this repository. 😮

Similar to SYCL/Vulkan/etc., Metal would like to handle all its memory allocations through language specific buffer objects, which would control the lifetime of those allocations. Which is not quite how memory resources work. But it's also not completely unusable with memory resources. 🤔 The memory resource "just" needs to manage buffer objects itself.

This is what I did as a demonstration in `vecmem::metal::shared_memory_resource`. Which, on a very basic level, does seem to work on my M1 Mac Mini. 🎉

```
[zsh][Thranduil]:build > ./bin/vecmem_test_metal
[vecmem] metal/src/utils/device_wrapper.mm:37 Created an "owning wrapper" around device: Apple M1
[vecmem] metal/src/utils/device_wrapper.mm:37 Created an "owning wrapper" around device: Apple M1
[==========] Running 4 tests from 2 test suites.
[----------] Global test environment set-up.
[----------] 1 test from metal_basic_memory_resource_tests/memory_resource_test_basic
[ RUN      ] metal_basic_memory_resource_tests/memory_resource_test_basic.allocations/shared_resource
[       OK ] metal_basic_memory_resource_tests/memory_resource_test_basic.allocations/shared_resource (0 ms)
[----------] 1 test from metal_basic_memory_resource_tests/memory_resource_test_basic (0 ms total)

[----------] 3 tests from metal_host_accessible_memory_resource_tests/memory_resource_test_host_accessible
[ RUN      ] metal_host_accessible_memory_resource_tests/memory_resource_test_host_accessible.int_value/shared_resource
[       OK ] metal_host_accessible_memory_resource_tests/memory_resource_test_host_accessible.int_value/shared_resource (0 ms)
[ RUN      ] metal_host_accessible_memory_resource_tests/memory_resource_test_host_accessible.double_value/shared_resource
[       OK ] metal_host_accessible_memory_resource_tests/memory_resource_test_host_accessible.double_value/shared_resource (0 ms)
[ RUN      ] metal_host_accessible_memory_resource_tests/memory_resource_test_host_accessible.custom_value/shared_resource
[       OK ] metal_host_accessible_memory_resource_tests/memory_resource_test_host_accessible.custom_value/shared_resource (0 ms)
[----------] 3 tests from metal_host_accessible_memory_resource_tests/memory_resource_test_host_accessible (0 ms total)

[----------] Global test environment tear-down
[==========] 4 tests from 2 test suites ran. (0 ms total)
[  PASSED  ] 4 tests.
[zsh][Thranduil]:build >
```

But allocating memory by itself is of course not much of a thing. 🤔  And to actually implement Metal shaders that would do some computation, is a whole other level of complexity. 😢 See:
  - https://developer.apple.com/metal/sample-code/
  - https://developer.apple.com/documentation/metal/performing_calculations_on_a_gpu

I actually want to come back to this during future weekends, and goof some more, but of course none of this seems actually usable to us. Building "Metal shader libraries" is a bit of an involved process, which really only works with [XCode](https://developer.apple.com/xcode/) at the moment. Though with that generator, CMake can be used for it like:

```cmake
set_source_files_properties("add.metal"
    PROPERTIES
    LANGUAGE METAL
)
```

(Without having to write code like what we have for SYCL and HIP. Which others have done already, see: https://dpogue.ca/articles/cmake-metal.html)

So yeah, pretty silly, but I thought I would entertain you guys with it nevertheless. 😄


metal/src/utils/device_wrapper.mm
```diff
@@ -0,0 +1,142 @@
...
+    : m_device(nullptr), m_managedDevice() {
+
+    // Check whether the parent owns its own device or not.
+    if (parent.m_managedDevice) {
+        // If so, make a copy of it, and own that copy in this object as well.
+        // This makes sure that reference counts to the MTLDevice would be
+        // managed properly.
+        m_managedDevice =
+            std::make_unique<details::opaque_device>(*(parent.m_managedDevice));
+        m_device = (__bridge void*)m_managedDevice->m_device;
```
paulgessinger, line 67: What does `__bridge` do here?

paulgessinger (approved):
Neat. One could probably also do it by bridging with Swift, but this is probably even more hairy.

metal/src/memory/shared_memory_resource.mm
```diff
@@ -0,0 +1,80 @@
...
+
+void shared_memory_resource::do_deallocate(void* p, std::size_t, std::size_t) {
+
+    // Find the buffer, and release it. It's undefined behaviour to deallocate
+    // a buffer that was not allocated by this memory resource. So we might as
+    // well crash when that happens.
+    assert(m_data);
+    auto it = m_data->m_buffers.find(p);
+    assert(it != m_data->m_buffers.end());
+    m_data->m_buffers.erase(it);
```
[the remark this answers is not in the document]
krasznaa: My own experience with this is limited, but from what I understand, Apple uses ARC these days.
  - https://clang.llvm.org/docs/AutomaticReferenceCounting.html
  - https://en.wikipedia.org/wiki/Automatic_Reference_Counting

But I've indeed not quite made sure that it would be active in this case. 🤔 I **think** ARC is in effect all the time unless it's turned off, but at the same time I've also seen [@autoreleasepool](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/MemoryMgmt/Articles/mmAutoreleasePools.html) in some of the examples. But I think that's not needed for the garbage collection... 🤔 (I believe that's a slightly different thing.)

I'll need to check this in action, but I **think** one just needs to make `id` objects go out of scope, and the runtime should take care of deleting them eventually.

metal/src/utils/device_wrapper.mm
krasznaa: I'm on **very** shaky ground here. 🤔 I was trying to find out from the internets how to do the equivalent of what's happening in `vecmem::sycl::queue_wrapper`. Where a `void*` pointer is used to point at a `cl::sycl::queue` object in memory.

The compiler was happy to let me convert an `id<MTLDevice>` "object" to a `void*` pointer, but I was not sure how to "safely" do the reverse. Since the `reinterpret_cast` that we do in `vecmem::sycl::queue_wrapper` is not really an option here. (I think...)

After some looking around, I found: https://clang.llvm.org/docs/AutomaticReferenceCounting.html#bridged-casts Tried to use it, and the code compiled without warnings, and ran successfully. 🤔 So I went with this formalism. However it's absolutely true that this code might be doing something silly at this point.

krasznaa:
It seems that this goofing around is finishing here... 🤔 Since it's not even possible to use even `vecmem::data::vector_view` in a `.metal` file as it seems.

```
[build] CompileMetalFile /Users/krasznaa/ATLAS/projects/vecmem/vecmem/tests/metal/test_metal_containers_kernels.metal (in target 'vecmem_test_metal' from project 'vecmem')
[build]     cd /Users/krasznaa/ATLAS/projects/vecmem/vecmem
[build]     /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/metal -c -target air64-apple-macos15.1 -I/Users/krasznaa/ATLAS/projects/vecmem/vecmem/build/bin/Debug/include -F/Users/krasznaa/ATLAS/projects/vecmem/vecmem/build/bin/Debug -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX15.1.sdk -fmetal-math-mode\=fast -fmetal-math-fp32-functions\=fast -serialize-diagnostics /Users/krasznaa/ATLAS/projects/vecmem/vecmem/build/build/vecmem_test_metal.build/Debug/Metal/test_metal_containers_kernels-f8a2a21537653cdc54ebcc13175da71a.dia -o /Users/krasznaa/ATLAS/projects/vecmem/vecmem/build/build/vecmem_test_metal.build/Debug/Metal/test_metal_containers_kernels-f8a2a21537653cdc54ebcc13175da71a.air -I/Users/krasznaa/ATLAS/projects/vecmem/vecmem/build/metal/CMakeFiles -I/Users/krasznaa/ATLAS/projects/vecmem/vecmem/metal/include -I/Users/krasznaa/ATLAS/projects/vecmem/vecmem/build/core/CMakeFiles -I/Users/krasznaa/ATLAS/projects/vecmem/vecmem/core/include -MMD -MT dependencies -MF /Users/krasznaa/ATLAS/projects/vecmem/vecmem/build/build/vecmem_test_metal.build/Debug/Metal/test_metal_containers_kernels-f8a2a21537653cdc54ebcc13175da71a.dat /Users/krasznaa/ATLAS/projects/vecmem/vecmem/tests/metal/test_metal_containers_kernels.metal
[build] In file included from /Users/krasznaa/ATLAS/projects/vecmem/vecmem/tests/metal/test_metal_containers_kernels.metal:9:
[build] In file included from /Users/krasznaa/ATLAS/projects/vecmem/vecmem/core/include/vecmem/containers/data/vector_view.hpp:10:
[build] /Users/krasznaa/ATLAS/projects/vecmem/vecmem/core/include/vecmem/utils/type_traits.hpp:10:10: fatal error: 'algorithm' file not found
[build] #include <algorithm>
[build]          ^~~~~~~~~~~
[build] 1 error generated.
```

Note that had to do the following to just force the build to pick up the project's include directories for the compilation:

```cmake
# Force the Metal source code into working...
set_property( SOURCE "test_metal_containers_kernels.metal"
   PROPERTY LANGUAGE METAL )
set_property( SOURCE "test_metal_containers_kernels.metal"
   APPEND
   PROPERTY COMPILE_OPTIONS -I$<JOIN:$<TARGET_PROPERTY:vecmem_metal,INCLUDE_DIRECTORIES>, -I> )
```

For some reason `COMPILE_OPTIONS` is passed through to Xcode, but `INCLUDE_DIRECTORIES` isn't. 😕

Well, it was fun while it lasted... 🤔

PR #303, opened 2024-11-07
SonarCloud Fixes, main branch (2024.11.07.)

krasznaa:
This is a first set of fixes triggered by SonarCloud's results on our code. It's more or less exploratory at this point, as I want to see how easy it is to track issues in that system.

It was not directly pointed out by SonarCloud, but these changes also allowed me to remove one constructor of `vecmem::device_vector` since the non-const -> const conversion can actually happen at the level of the views. We don't need to have this type of templated code inside of `device_vector` itself. 🤔

@stephenswat, I also received some notes about "your" code. Have a look if you like what I did to resolve those reports.


stephenswat (approved):
LGTM.

PR #304, opened 2024-11-08
Remove Deprecated Types, main branch (2024.11.08.)

krasznaa:
I bumped into `vecmem::const_device_array` and `vecmem::const_device_vector` today.

These are not types that we would've ever actually used in client projects. As the API has been pretty much from the beginning, that constness is expressed through the template types in `vecmem::device_vector`, etc. types.


krasznaa:
image

No way... That was time very well spent with you this afternoon! 😄

PR #307, opened 2024-12-12
Allow const and non-const reverse iterator comparisons

stephenswat:
As @beomki-yeo discovered in https://github.com/acts-project/traccc/pull/788, constant reverse iterators cannot be currently compared to non-constant reverse iterators as is required by the C++ standard in section [container.requirements.general]. This commit adds the necessary comparison code, and adds some tests for the new behaviour.

### DIFF
--- a/tests/core/test_core_containers.cpp
+++ b/tests/core/test_core_containers.cpp
@@ -71,6 +71,35 @@ TEST_F(core_container_test, device_vector) {
     }
 }
 
+/// Test(s) for @c vecmem::device_vector
+TEST_F(core_container_test, device_vector_iterators) {
+
+    using V = vecmem::device_vector<int>;
+    V test_vector(vecmem::get_data(m_reference_vector));
+
+    V::const_iterator itfc1 = test_vector.cbegin();
+    V::const_iterator itfc2 = test_vector.cend();
+    V::iterator itfn1 = test_vector.begin();
+    V::iterator itfn2 = test_vector.end();
+    V::const_reverse_iterator itrc1 = test_vector.crbegin();
+    V::const_reverse_iterator itrc2 = test_vector.crend();
+    V::reverse_iterator itrn1 = test_vector.rbegin();
+    V::reverse_iterator itrn2 = test_vector.rend();
+
+    EXPECT_EQ(itfn1, itfc1);
+    EXPECT_EQ(itfc1, itfn1);
+    EXPECT_NE(itfn2, itfc1);
+    EXPECT_NE(itfc1, itfn2);
+    EXPECT_EQ(itrn1, itrc1);
+    EXPECT_EQ(itrc1, itrn1);
+    EXPECT_NE(itrn2, itrc1);
+    EXPECT_NE(itrc1, itrn2);
+    EXPECT_EQ(itfc2, itfn2);
+    EXPECT_NE(itfc2, itfn1);
+    EXPECT_EQ(itrc2, itrn2);
+    EXPECT_NE(itrc2, itrn1);
+}
+
 /// Test(s) for @c vecmem::static_vector
 TEST_F(core_container_test, static_vector) {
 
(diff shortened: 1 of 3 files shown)


tests/core/test_core_containers.cpp
```diff
@@ -71,6 +71,48 @@ TEST_F(core_container_test, device_vector) {
...
+    EXPECT_TRUE(itfn2 != itfc1);
+    EXPECT_TRUE(itfc1 != itfn2);
+    EXPECT_TRUE(itrn1 == itrc1);
+    EXPECT_TRUE(itrc1 == itrn1);
+    EXPECT_TRUE(itrn2 != itrc1);
+    EXPECT_TRUE(itrc1 != itrn2);
+    EXPECT_TRUE(itfc2 == itfn2);
+    EXPECT_TRUE(itfc2 != itfn1);
+    EXPECT_TRUE(itrc2 == itrn2);
+    EXPECT_TRUE(itrc2 != itrn1);
```
krasznaa, line 106: Let's rather use `EXPECT_EQ(...)` / `EXPECT_NEQ(...)` on these. I believe those should use the comparison operators correctly. 🤔

tests/core/test_core_containers.cpp
```diff
@@ -71,6 +71,48 @@ TEST_F(core_container_test, device_vector) {
...
+TEST_F(core_container_test, device_nonconst_vector) {
+
+    vecmem::device_vector<int> test_vector(
+        vecmem::get_data(m_reference_vector));
+    EXPECT_EQ(test_vector.size(), m_reference_vector.size());
+    EXPECT_EQ(test_vector.empty(), m_reference_vector.empty());
+    EXPECT_TRUE(std::equal(m_reference_vector.begin(), m_reference_vector.end(),
+                           test_vector.begin()));
+    EXPECT_EQ(std::accumulate(test_vector.begin(), test_vector.end(), 0),
+              std::accumulate(test_vector.rbegin(), test_vector.rend(), 0));
```
krasznaa, line 84: This is a duplication of tests we have already. I don't see the point in duplicating them.

tests/core/test_core_containers.cpp
```diff
@@ -71,6 +71,48 @@ TEST_F(core_container_test, device_vector) {
...
+
+    decltype(test_vector)::const_iterator itfc1 = test_vector.cbegin();
+    decltype(test_vector)::const_iterator itfc2 = test_vector.cend();
+    decltype(test_vector)::iterator itfn1 = test_vector.begin();
+    decltype(test_vector)::iterator itfn2 = test_vector.end();
+    decltype(test_vector)::const_reverse_iterator itrc1 = test_vector.crbegin();
+    decltype(test_vector)::const_reverse_iterator itrc2 = test_vector.crend();
+    decltype(test_vector)::reverse_iterator itrn1 = test_vector.rbegin();
+    decltype(test_vector)::reverse_iterator itrn2 = test_vector.rend();
```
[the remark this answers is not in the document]
stephenswat: Replaced by a typedef.

tests/core/test_core_containers.cpp
stephenswat: Sure, I like being explicit here as we're specifically testing the `==` and `!=` operators but this probably works, too.

tests/core/test_core_containers.cpp
```diff
@@ -71,6 +71,48 @@ TEST_F(core_container_test, device_vector) {
...
+    EXPECT_TRUE(itfc2 != itfn1);
+    EXPECT_TRUE(itrc2 == itrn2);
+    EXPECT_TRUE(itrc2 != itrn1);
+
+    for (std::size_t i = 0; i < m_reference_vector.size(); ++i) {
+        const vecmem::device_vector<int>::size_type ii =
+            static_cast<vecmem::device_vector<int>::size_type>(i);
+        EXPECT_EQ(test_vector.at(ii), m_reference_vector.at(i));
+        EXPECT_EQ(test_vector[ii], m_reference_vector[i]);
+    }
```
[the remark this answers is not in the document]
stephenswat: We have these tests for const vectors, but not yet for non-const vectors. As far as I'm concerned that's a nice and easy thing to add.

tests/core/test_core_containers.cpp
stephenswat: See comment above.

PR #309, opened 2025-02-20
SoA Proxy Updates, main branch (2025.02.20.)

krasznaa:
Triggered by @beomki-yeo's insistence in https://github.com/acts-project/traccc/pull/878, I spent a bit of time trying to come up with a way of filling SoA containers conveniently "by hand".

I considered a number of options, but in the end went with making `vecmem::edm::proxy` even a bit more complicated. 🤔 Instead of only being able to refer to (i.e. "proxy"...) an element of a container, now they can also behave as standalone objects, which hold some payload of their own.

I went with this, because if I introduced a new type for this (let's say `vecmem::edm::object`), I would've had to create conversion operators back and forth between the new type and `vecmem::edm::proxy`. So it was easier to just use the same type for the standalone objects as well.

So far I only added `vecmem::edm::host::push_back(...)` with the new machinery, just as a proof of concept. Along with some testing code for it. Please have a look, especially at the unit test code. Since some early feedback could be useful here.

Note that it will never be possible to use such a formalism on containers that have jagged vectors, in device code. (Since a dynamic sized vector cannot be used in device code for this.) But since SoA containers with jagged vectors are not resizable to begin with, that is not a new limitation actually... 🤔

### DIFF
--- a/core/include/vecmem/edm/host.hpp
+++ b/core/include/vecmem/edm/host.hpp
@@ -59,6 +59,11 @@ public:
     using const_proxy_type = interface_type<
         proxy<schema_type, details::proxy_domain::host,
               details::proxy_access::constant, details::proxy_type::reference>>;
+    /// Type type of standalone proxy objects for the container
+    using object_type =
+        interface_type<proxy<schema_type, details::proxy_domain::host,
+                             details::proxy_access::non_constant,
+                             details::proxy_type::standalone>>;
 
     /// @name Constructors and assignment operators
     /// @{
@@ -81,6 +86,8 @@ public:
     /// Reserve memory for the container
     VECMEM_HOST
     void reserve(size_type size);
+    /// Add a new element to the container
+    VECMEM_HOST void push_back(const object_type& element);
 
     /// Get the vector of a specific variable (non-const)
     template <std::size_t INDEX>
--- a/core/include/vecmem/edm/host.hpp
+++ b/core/include/vecmem/edm/host.hpp
@@ -1,6 +1,6 @@
 /* VecMem project, part of the ACTS project (R&D line)
  *
- * (c) 2023-2024 CERN for the benefit of the ACTS project
+ * (c) 2023-2025 CERN for the benefit of the ACTS project
  *
  * Mozilla Public License Version 2.0
  */
@@ -52,11 +52,11 @@ public:
     using interface_type = INTERFACE<T>;
     /// The type of the (non-const) proxy objects for the container elements
     using proxy_type =
-        interface_type<proxy<schema_type, details::proxy_type::host,
+        interface_type<proxy<schema_type, details::proxy_domain::host,
                              details::proxy_access::non_constant>>;
     /// The type of the (const) proxy objects for the container elements
     using const_proxy_type =
-        interface_type<proxy<schema_type, details::proxy_type::host,
+        interface_type<proxy<schema_type, details::proxy_domain::host,
                              details::proxy_access::constant>>;
 
     /// @name Constructors and assignment operators
--- a/core/include/vecmem/edm/host.hpp
+++ b/core/include/vecmem/edm/host.hpp
@@ -53,11 +53,12 @@ public:
     /// The type of the (non-const) proxy objects for the container elements
     using proxy_type =
         interface_type<proxy<schema_type, details::proxy_domain::host,
-                             details::proxy_access::non_constant>>;
+                             details::proxy_access::non_constant,
+                             details::proxy_type::reference>>;
     /// The type of the (const) proxy objects for the container elements
-    using const_proxy_type =
-        interface_type<proxy<schema_type, details::proxy_domain::host,
-                             details::proxy_access::constant>>;
+    using const_proxy_type = interface_type<
+        proxy<schema_type, details::proxy_domain::host,
+              details::proxy_access::constant, details::proxy_type::reference>>;
 
     /// @name Constructors and assignment operators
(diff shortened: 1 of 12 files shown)


core/include/vecmem/edm/host.hpp
```diff
@@ -80,6 +86,8 @@ class host<schema<VARTYPES...>, INTERFACE> {
     /// Reserve memory for the container
     VECMEM_HOST
     void reserve(size_type size);
+    /// Add a new element to the container
+    VECMEM_HOST void push_back(const object_type& element);
```
krasznaa, line 90: Note that, as I discussed with @stephenswat, I was trying to make this function much more generic at first. So that instead of using `object_type`, it would use a generic proxy here. But Clang was really not happy about it. (Getting itself confused about what to do with reference variables at the end of some long templating stacks.)

So, in order to use `push_back(...)`, all variables need to be loaded onto the stack first. Even if one is copying an element from another SoA container. Not what I intended, but I **think** it should not matter much in the end.

core/include/vecmem/edm/impl/host.ipp
```diff
@@ -65,6 +65,22 @@ VECMEM_HOST void host<schema<VARTYPES...>, INTERFACE>::reserve(
...
+    // Make sure that there are some (jagged) vector types in the container.
+    static_assert(
+        std::disjunction_v<type::details::is_vector<VARTYPES>...>,
+        "This function requires at least one (jagged) vector variable.");
+
+    // Add a new element to the container.
+    const size_type index = size();
+    resize(index + 1);
+    // Set the new element.
+    at(index) = element;
```
krasznaa, line 81: The implementation here is super naive, but again, I thought this could be good enough. But let me know if you guys know of a non-too-verbose way of doing this in a more efficient way.

tests/core/test_core_edm_host.cpp
```diff
@@ -213,3 +212,91 @@ TEST_F(core_edm_host_test, const_proxy) {
...
+TEST_F(core_edm_host_test, push_back_jagged) {
+
+    // Create an empty host container.
+    host_type host{m_resource};
+
+    // Add a couple of elements to it. Since interface has no constructor of
+    // its own, because with jagged vectors that's not so easy to do, one needs
+    // to use double braces in the following expression.
+    host.push_back({{1, 2.f, vecmem::vector<double>{3., 4., 5.}}});
+    host.push_back({{2, 3.f, vecmem::vector<double>{4., 5., 6.}}});
```
krasznaa, line 260: I'm not absolutely happy about the formalism here, but I couldn't easily come up with one that would not require double curly braces. 🤔

krasznaa: Turns out that it was not all that hard to implement it in a way that results in a more natural formalism for the client code. 🤔 (All the infrastructure was already in place to make it happen.)

stephenswat:
Is this still under development or is it ready for review?

krasznaa:
Please have a look. By now it's in a shape that I'm fairly happy with.

core/include/vecmem/utils/tuple.hpp
```diff
@@ -93,6 +93,25 @@ struct tuple<T, Ts...> {
...
+    ///
+    template <typename U, typename... Us,
+              std::enable_if_t<vecmem::details::negation<std::is_same<
+                                   tuple<T, Ts...>, tuple<U, Us...>>>::value &&
+                                   sizeof...(Ts) == sizeof...(Us),
+                               bool> = true>
+    VECMEM_HOST_AND_DEVICE constexpr tuple &operator=(
+        const tuple<U, Us...> &parent) {
+        if (static_cast<const void *>(this) !=
+            static_cast<const void *>(&parent)) {
```
krasznaa, line 108: Now that I think about it, since this function now makes sure that the auto-generated assignment operator was left in place, this `if(...)` check is not actually needed...

core/include/vecmem/edm/impl/host.ipp
```diff
@@ -65,6 +65,22 @@ VECMEM_HOST void host<schema<VARTYPES...>, INTERFACE>::reserve(
...
 }
 
+template <typename... VARTYPES, template <typename> class INTERFACE>
+VECMEM_HOST void host<schema<VARTYPES...>, INTERFACE>::push_back(
+    const object_type& element) {
+
+    // Make sure that there are some (jagged) vector types in the container.
+    static_assert(
+        std::disjunction_v<type::details::is_vector<VARTYPES>...>,
+        "This function requires at least one (jagged) vector variable.");
```
stephenswat, line 75: Why does pushing back into an SoA require at least one jagged vector? :confused:

stephenswat: I'm not sure I follow, one should be able to push back into an SoA of just scalar values, no?

krasznaa: `vecmem::edm::type::scalar` means that we have one variable for the entire container. I.e.:

```c++
struct soa_host_container {
   int foo; // This is a scalar variable
   vecmem::vector<int>; // This is a vector variable
   vecmem::jagged_vector<int>; // This is a jagged vector
};
```

As I learned with the xAOD developments more than a decade ago, one can get easily confused about types when constantly switching between what the container has, and what the proxies have.

stephenswat: Okay, I'll take your word on it.

stephenswat (approved):

PR #312, opened 2025-02-25
Change default platform to AMD

StewMH:
Remove this warning:
```
Warning: HIP_PLATFORM=hcc is deprecated.Please use HIP_PLATFORM=amd.
```
and some related failures in debug builds.

Longer term might be good to introduce a flag to use native CMake HIP support only.


krasznaa:
I pondered about doing this many times over the years. Well, I was considering changing the default from `"hcc"` to `"amd"`. I didn't consider removing all mentions of the old names...

You see, AMD did this update in a sort of weird way. At some `5.X` version of HIP (don't remember anymore when this happened) they introduced the `"amd"` and `"nvidia"` values, and deprecated `"hcc"` and `"nvcc"` right away. And ever since, up until their current latest software version, they still recognize the `"hcc"` and `"nvcc"` values. And still print warnings about them.

There is no reasonable way in the code for choosing a default based on the version of HIP found. So I decided to keep the old `"hcc"` value as the default, as that would work with all the HIP versions that I ever tested.

Note that you don't need to accept the warnings in your build. :thinking: You can just set

```sh
export HIP_PLATFORM=amd
```

in your build environment, and the warnings will be gone. As long as you use a HIP version that accepts that value of course.

I don't see much upside in reducing the compatibility of the code. As I started, I considered it many times over the years. Always getting to the same conclusion that the warnings are a lesser evil than dripping support for older compilers.

With that being said, I could possibly get on board with changing the default. But for sure I don't want to erase `"hcc"` and `"nvcc"` from the code. :thinking:

StewMH:
Right, the problem is that this now breaks:
```
#elif !defined(__HIP_PLATFORM_AMD__) && defined(__HIP_PLATFORM_NVIDIA__)
#include "hip/nvidia_detail/nvidia_hip_runtime_api.h"
#else
#error("Must define exactly one of __HIP_PLATFORM_AMD__ or __HIP_PLATFORM_NVIDIA__");
#endif
```
https://github.com/ROCm/HIP/blob/4f263b6e90770c55c694af6beb0924db084b952f/include/hip/hip_runtime_api.h#L9337
with ROCM 6.3. At least I managed to break it last week with a debug build, although I cannot currently reproduce this.

I will revert to keeping the options in.

krasznaa (approved):
Technically I don't think that the C\+\+ code of HIP would've changed much any time recently. They've been using macros with `AMD` and `NVIDIA` in them since a while.

Surprisingly though even `hipcc` seems to have forgotten about those values some time ago already... :thinking:

https://github.com/ROCm/llvm-project/blame/amd-staging/amd/hipcc/bin/hipcc.pl#L118

So sure, let's change the default finally. :+1:

PR #313, opened 2025-02-28
SoA Fixes, main branch (2025.02.28.)

krasznaa:
While preparing an update for [traccc](https://github.com/acts-project/traccc), I came across a number of issues in our code. :frowning:

First off, making an SoA container with a `detray::bound_track_parameters` element does not work currently. Because the `get_data(...)` calls implemented for `vecmem::edm::host` make compilers not know which `get_data(...)` function to use for such a Detray type. Because Detray happens to implement some super generic `detray::get_data` functions itself.

Originally I used scope-less `get_data(...)` calls on purpose in those functions, to allow users to slide in with their own definitions if needed. But now I rather think it's better to go like this, forwarding specifically to `::vecmem::get_data(...)` calls instead.

The other (much bigger) development of the PR is to make it possible to perform assignments between the elements of SoA containers. Like:

```c++
auto container1 = ...;
auto container2 = ...;
container1.at(...) = container2.at(...);
```

To make it happen a number of fixes needed to be made in both `vecmem::edm::proxy`, `vecmem::tuple` and `vecmem::device_vector`. :thinking: Many of which only came up when I added some explicit tests for this functionality. Since the traccc code started working with much more incorrect vecmem code as well. :stuck_out_tongue:

Basically they all now define constructors and assignment operators explicitly for both receiving a parent of the exact same type, and also a parent that is "a little different". (But still of an acceptable type.) For the latter I started using [std::is_constructible](https://en.cppreference.com/w/cpp/types/is_constructible) and [std::is_assignable](https://en.cppreference.com/w/cpp/types/is_assignable) instead of the quite broken tests that are in place right now.

### DIFF
--- a/core/include/vecmem/containers/impl/device_vector.ipp
+++ b/core/include/vecmem/containers/impl/device_vector.ipp
@@ -1,6 +1,6 @@
 /* VecMem project, part of the ACTS project (R&D line)
  *
- * (c) 2021-2024 CERN for the benefit of the ACTS project
+ * (c) 2021-2025 CERN for the benefit of the ACTS project
  *
  * Mozilla Public License Version 2.0
  */
@@ -29,8 +29,10 @@ VECMEM_HOST_AND_DEVICE device_vector<TYPE>::device_vector(
 }
 
 template <typename TYPE>
+template <typename OTHERTYPE,
+          std::enable_if_t<std::is_convertible<OTHERTYPE, TYPE>::value, bool>>
 VECMEM_HOST_AND_DEVICE device_vector<TYPE>::device_vector(
-    const device_vector& parent)
+    const device_vector<OTHERTYPE>& parent)
     : m_capacity(parent.m_capacity),
       m_size(parent.m_size),
       m_ptr(parent.m_ptr) {
@@ -52,9 +54,26 @@ VECMEM_HOST_AND_DEVICE device_vector<TYPE>& device_vector<TYPE>::operator=(
     }
 
     // Copy the other object's payload.
-    m_capacity = rhs.m_capacity;
-    m_size = rhs.m_size;
-    m_ptr = rhs.m_ptr;
+    resize(rhs.size());
+    for (size_type i = 0; i < size(); ++i) {
+        this->at(i) = rhs.at(i);
+    }
+
+    // Return a reference to this object.
+    return *this;
+}
+
+template <typename TYPE>
+template <typename OTHERTYPE,
+          std::enable_if_t<std::is_convertible<OTHERTYPE, TYPE>::value, bool>>
+VECMEM_HOST_AND_DEVICE device_vector<TYPE>& device_vector<TYPE>::operator=(
+    const device_vector<OTHERTYPE>& rhs) {
+
+    // Copy the other object's payload.
+    resize(rhs.size());
+    for (size_type i = 0; i < size(); ++i) {
+        this->at(i) = rhs.at(i);
+    }
 
     // Return a reference to this object.
     return *this;
--- a/core/include/vecmem/utils/tuple.hpp
+++ b/core/include/vecmem/utils/tuple.hpp
@@ -51,13 +51,21 @@ struct tuple<T, Ts...> {
 
     /// Default constructor
     constexpr tuple() = default;
+    /// Default copy constructor
+    constexpr tuple(const tuple &) = default;
+    /// Default move constructor
+    constexpr tuple(tuple &&) noexcept = default;
 
     /// Copy constructor
     ///
     /// @param parent The parent to copy
     ///
     template <typename U, typename... Us,
-              std::enable_if_t<sizeof...(Ts) == sizeof...(Us), bool> = true>
+              std::enable_if_t<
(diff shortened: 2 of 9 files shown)


core/include/vecmem/utils/tuple.hpp
```diff
@@ -93,15 +100,29 @@ struct tuple<T, Ts...> {
...
     ///
     template <typename U, typename... Us,
-              std::enable_if_t<
-                  (!std::is_same<tuple<T, Ts...>, tuple<U, Us...>>::value) &&
-                      sizeof...(Ts) == sizeof...(Us),
-                  bool> = true>
+              std::enable_if_t<vecmem::details::conjunction<
+                                   std::is_assignable<T, U &&>,
+                                   std::is_assignable<Ts, Us &&>...>::value,
+                               bool> = true>
```
stephenswat, line 125: This doesn't seem right; I suspect this should be `decay_t<U>` rather than `U&&`.

core/include/vecmem/utils/tuple.hpp
```diff
@@ -51,13 +51,20 @@ struct tuple<T, Ts...> {
...
     /// Copy constructor
     ///
     /// @param parent The parent to copy
     ///
     template <typename U, typename... Us,
-              std::enable_if_t<sizeof...(Ts) == sizeof...(Us), bool> = true>
+              std::enable_if_t<vecmem::details::conjunction<
+                                   std::is_constructible<T, U &&>,
+                                   std::is_constructible<Ts, Us &&>...>::value,
+                               bool> = true>
```
stephenswat, line 67: Same as below.

core/include/vecmem/containers/impl/device_vector.ipp
```diff
@@ -52,9 +54,26 @@ VECMEM_HOST_AND_DEVICE device_vector<TYPE>& device_vector<TYPE>::operator=(
...
+        this->at(i) = rhs.at(i);
+    }
+
+    // Return a reference to this object.
+    return *this;
+}
+
+template <typename TYPE>
+template <typename OTHERTYPE,
+          std::enable_if_t<std::is_convertible<OTHERTYPE, TYPE>::value, bool>>
```
stephenswat, line 68:
```suggestion
          std::enable_if_t<std::is_convertible<OTHERTYPE, TYPE>::value, bool> = true>
```

core/include/vecmem/utils/tuple.hpp
krasznaa: :thinking: How sure are you that this is not the correct formalism?

Let's ignore that we use exactly this formalism in a number of places already. (https://github.com/acts-project/vecmem/blob/main/core/include/vecmem/utils/tuple.hpp#L90-L91, https://github.com/acts-project/detray/blob/main/core/include/detray/utils/tuple.hpp#L43-L44, https://github.com/acts-project/detray/blob/main/core/include/detray/utils/tuple.hpp#L49-L50, etc. Much of which you have written.)

But the more I think of it, the more I think that using a universal reference here is the right approach. Since `&&` in this context should not mean an rvalue reference, but a universal one. (Though I admit, I'm standing on slightly shaky grounds here...)

I tried, just removing the `&&` from all those type declarations seems to work as well. (`std::decay` I didn't try, as removing constness and references from these types would mess with these tests for sure.) But I'm not entirely sure if that's just by chance.

I'll do a little more testing still...

krasznaa: After quite some further reflection, I still believe that using `U&&` as a universal reference in these expressions is the correct thing to do. 🤔

core/include/vecmem/utils/tuple.hpp
```diff
@@ -93,15 +100,29 @@ struct tuple<T, Ts...> {
...
     /// Assignment operator from a (slightly) different tuple type
     ///
     /// @param parent The parent to copy
     ///
     template <typename U, typename... Us,
-              std::enable_if_t<
-                  (!std::is_same<tuple<T, Ts...>, tuple<U, Us...>>::value) &&
-                      sizeof...(Ts) == sizeof...(Us),
-                  bool> = true>
+              std::enable_if_t<vecmem::details::conjunction<
```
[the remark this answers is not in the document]
krasznaa: Added the size check back.

core/include/vecmem/utils/tuple.hpp
stephenswat: No, it's not... Consider the fact that you are predicating `m_head = parent.m_head;` on `std::is_assignable<T, U &&>`. Now imagine a type `Q = T = U` that is not copyable but is movable. That type would satisfy your SFINAE but then fail on the assignment (without perfect fowarding).

krasznaa: The **big** problem here is that `U` is often a reference. (In fact in most cases in this code.) So if I write `decay_t<U>`, that's also not the type that would take part in the assignment.

But at this point I'd rather have something that works with our current code, than to have something that is perfect. So let's use `std::decay_t<T>`. We can correct it later if we run into an issue with it.

stephenswat (approved):
Sure :shrug:

PR #314, opened 2025-03-19
Prevent CMake flags from propagating to parent

stephenswat:
See https://github.com/acts-project/detray/pull/956.

### DIFF
--- a/cmake/vecmem-compiler-options-cpp.cmake
+++ b/cmake/vecmem-compiler-options-cpp.cmake
@@ -4,6 +4,10 @@
 #
 # Mozilla Public License Version 2.0
 
+cmake_minimum_required(VERSION 3.21)
+
+# Only set these compiler flags if we are the top level project.
+if(PROJECT_IS_TOP_LEVEL)
 # Include the helper function(s).
 include( vecmem-functions )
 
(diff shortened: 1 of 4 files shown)


krasznaa:
Since this is clearly a contentious issue, let me put it into writing.

I do not believe that this setup would be more robust than what we have now. I very much believe that we need to embrace the behaviour of CMake cache variables, and use them such that the client projects would behave as needed.

Let's consider for instance the case where a client project would not be using C\+\+ itself. Let's say it would be only building Rust code. (But still being able to use the vecmem libraries through some mechanism.) In such a setup vecmem would not build at all all of a sudden. Since nobody would be setting `CMAKE_CXX_STANDARD` to anything.

I firmly believe that the logic that cache variables provide is a robust one. Where a cache variable is only set using [set(...)](https://cmake.org/cmake/help/latest/command/set.html) when the variable doesn't exist yet. Allowing projects to set meaningful defaults, which can be overridden. Putting a different logic on these variables, like the one proposed here, would just make our project "less standard" than I'd like.

stephenswat:
> Putting a different logic on these variables, like the one proposed here, would just make our project "less standard" than I'd like.

I maintain that this argument boils down to "everyone else does it wrong, so we should too". It's a race to the bottom, and I'm sad to see we are participating in it.

PR #317, opened 2025-03-31
SonarCloud Inspired Changes, main branch (2025.03.31.)

krasznaa:
Implemented many of SonarCloud's suggestions for the (base) containers.

Simplified the view types by not using `const` and non-`const` accessor functions separately anymore. Making the types more similar to the implementation of [std::span](https://en.cppreference.com/w/cpp/container/span).
  - Previously the `const_pointer` types were incorrectly defined. `vecmem::vector_view<T>::const_pointer` was not `const T*`, but rather `T* const`. This is why the code worked at all. 🤔 It turns out that having separate `const` and non-`const` accessor functions on the views is not meaningful in the end. With the final nail in the coffin being lambdas. 🤔 Since the following just didn't work once I fixed the previously incorrectly defined types:

```c++
vecmem::vector_view<int> view;
[view]() {
   int* ptr = view.ptr();
}();
```

(Since captured objects are **always** `const`. So not allowing access to the non-const pointer through a const view object makes the use of these views **very** inconvenient with SYCL.)

Stopped using protected variables in the view types. Making the derived classes set up their base objects through operator=(...).

### DIFF
--- a/core/include/vecmem/containers/data/jagged_vector_buffer.hpp
+++ b/core/include/vecmem/containers/data/jagged_vector_buffer.hpp
@@ -1,7 +1,7 @@
 /*
  * VecMem project, part of the ACTS project (R&D line)
  *
- * (c) 2021-2023 CERN for the benefit of the ACTS project
+ * (c) 2021-2025 CERN for the benefit of the ACTS project
  *
  * Mozilla Public License Version 2.0
  */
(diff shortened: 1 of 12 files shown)


stephenswat (approved):

krasznaa:
image

Only about 600 to go... 😛

PR #320, opened 2025-04-16
HIP: Add Async Copy

CrossR:
This is a simple duplication of the existing CUDA async copy bits, to HIP. I've also updated the tests to match the CUDA ones.

I've got a branch of the traccc Alpaka code that is now using non-blocking queues (i.e. CUDA/HIP streams), and async copy, but we need the equivalent code for HIP, for it to work there as well.

### DIFF
--- a/tests/hip/test_hip_copy.cpp
+++ b/tests/hip/test_hip_copy.cpp
@@ -63,9 +63,7 @@ static ::testing::Environment* const async_hip_copy_env =
 
 /// The configurations to run the tests with. Skip tests with asynchronous
 /// copies on "managed host memory" on Windows. (Using managed memory as
-/// "device memory" does seem to be fine.) There seems to be some issue with
-/// HIP, which results in SEH exceptions when scheduling an asynchronous
-/// copy in such a setup, while a previous asynchronous copy is still running.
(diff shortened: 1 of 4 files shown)


tests/hip/test_hip_copy.cpp
```diff
@@ -19,22 +19,74 @@
...
+        hip_stream.reset();
+        hip_async_device_copy_ptr = nullptr;
+    }
+};
+
+/// Register the environment
+static ::testing::Environment* const async_hip_copy_env =
+    ::testing::AddGlobalTestEnvironment(new AsyncHIPCopyEnvironment{});
+
+/// The configurations to run the tests with. Skip tests with asynchronous
```
StewMH, line 64: Not sure if Windows fails here? I have no way to test it but we may as well define the tests for Windows in case someone does...

krasznaa: While HIP is now technically available on Windows, we don't test any of the GPU libraries on Windows in the CI. As it's really not that easy to install CUDA, oneAPI, or any of these things as part of the CI. (Not impossible as far as I can tell, but far from easy.)

So we don't need to worry about any of that in the CUDA, HIP and SYCL libraries.

PR #325, opened 2025-05-02
Tuple Trivial Copy, main branch (2025.05.02.)

krasznaa:
While trying to integrate `v1.15.0` into [traccc](https://github.com/acts-project/traccc), I came across a very surprising issue, courtesy of [Alpaka](https://github.com/alpaka-group/alpaka).

```
...
[ 96%] Building CXX object device/alpaka/CMakeFiles/traccc_alpaka.dir/src/clusterization/clusterization_algorithm.cpp.o
In file included from /home/krasznaa/ATLAS/projects/traccc/traccc/device/alpaka/src/clusterization/clusterization_algorithm.cpp:11:
In file included from /home/krasznaa/ATLAS/projects/traccc/traccc/device/alpaka/src/clusterization/../utils/barrier.hpp:11:
In file included from /home/krasznaa/ATLAS/projects/traccc/traccc/device/alpaka/src/clusterization/../utils/utils.hpp:10:
In file included from /home/krasznaa/ATLAS/projects/traccc/build/_deps/alpaka-src/include/alpaka/alpaka.hpp:13:
In file included from /home/krasznaa/ATLAS/projects/traccc/build/_deps/alpaka-src/include/alpaka/acc/AccCpuOmp2Blocks.hpp:27:
In file included from /home/krasznaa/ATLAS/projects/traccc/build/_deps/alpaka-src/include/alpaka/acc/Traits.hpp:14:
/home/krasznaa/ATLAS/projects/traccc/build/_deps/alpaka-src/include/alpaka/kernel/Traits.hpp:290:27: error: static assertion failed due to requirement 'isKernelArgumentTriviallyCopyable<vecmem::edm::view<vecmem::edm::schema<vecmem::edm::type::vector<const detray::geometry::barcode>, vecmem::edm::type::vector<const unsigned long>, vecmem::edm::type::vector<const float>, vecmem::edm::type::vector<const float>, vecmem::edm::type::vector<const float>, vecmem::edm::type::vector<const float>, vecmem::edm::type::vector<const float>, vecmem::edm::type::vector<const unsigned char>>>>': The kernel argument T must be trivially copyable!
  290 |             static_assert(isKernelArgumentTriviallyCopyable<T>, "The kernel argument T must be trivially copyable!");
      |                           ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/krasznaa/ATLAS/projects/traccc/build/_deps/alpaka-src/include/alpaka/kernel/Traits.hpp:344:18: note: in instantiation of function template specialization 'alpaka::detail::assertKernelArgIsTriviallyCopyable<vecmem::edm::view<vecmem::edm::schema<vecmem::edm::type::vector<const detray::geometry::barcode>, vecmem::edm::type::vector<const unsigned long>, vecmem::edm::type::vector<const float>, vecmem::edm::type::vector<const float>, vecmem::edm::type::vector<const float>, vecmem::edm::type::vector<const float>, vecmem::edm::type::vector<const float>, vecmem::edm::type::vector<const unsigned char>>>>' requested here
  344 |         (detail::assertKernelArgIsTriviallyCopyable<std::decay_t<TArgs>>(), ...);
      |                  ^
...
```

For some reason Alpaka decided that kernel arguments must all be trivially copyable. Which `vecmem::edm::view` isn't ever since #313, when `vecmem::tuple` received some custom assignment operators.

Trying to fix this led me down a deep rabbit hole... :thinking: Did you know for instance that

```c++
std::is_trivially_copyable_v<std::tuple<int, float>> == false
```

? There is also no requirement in the C\+\+ standard for `std::tuple` to behave that way, so this is perfectly legal behaviour.

Note also that CUDA, HIP and SYCL didn't complain about any of this. It's "just" Alpaka that decided that trivial copyability should be a requirement. :confused:

As discussed in #313, the custom assignment operator was really only needed to make `vecmem::edm::proxy` work as intended. So what I did in the end was to make `vecmem::tuple` trivially copyable, and then use a slightly more elaborate formalism in `vecmem::edm::proxy` for assignments between proxy objects.

But while I was doing all of this, I came across some revelations about C\+\+. Since I was **extremely** surprised that:

```c++
std::is_assignable_v<int*, int*> == false
```

I.e. one of the assignment operators introduced in #313 was not active most of the time. (Which also explained why I couldn't make the constructor of `vecmem::tuple` explicit in the past.)

So I tried to once again give a best try to implement SFINAE correctly for `vecmem::tuple`. Let's hope I came closer to the truth this time around. :thinking:

Pinging @StewMH and @CrossR for the info about Alpaka. :wink:

### DIFF
--- a/tests/core/test_core_device_containers.cpp
+++ b/tests/core/test_core_device_containers.cpp
@@ -1,6 +1,6 @@
 /* VecMem project, part of the ACTS project (R&D line)
  *
- * (c) 2021-2023 CERN for the benefit of the ACTS project
+ * (c) 2021-2025 CERN for the benefit of the ACTS project
  *
  * Mozilla Public License Version 2.0
  */
@@ -27,9 +27,8 @@
 #include <type_traits>
 
 /// Test case for the custom device container types
-class core_device_container_test : public testing::Test {
+struct core_device_container_test : public testing::Test {
 
-protected:
     /// Memory resource for the test(s)
     vecmem::host_memory_resource m_resource;
     /// Helper object for the memory copies.
@@ -38,13 +37,15 @@ protected:
 };  // class core_device_container_test
 
 /// Test that the "simple" data types are trivially constructible.
-TEST_F(core_device_container_test, trivial_construct) {
+TEST_F(core_device_container_test, construct_copy_assign) {
 
     EXPECT_TRUE(std::is_trivially_default_constructible<
                 vecmem::data::jagged_vector_view<const int> >());
     EXPECT_TRUE(std::is_trivially_constructible<
                 vecmem::data::jagged_vector_view<const int> >());
     EXPECT_TRUE(std::is_trivially_copy_constructible<
+                vecmem::data::jagged_vector_view<const int> >());
+    EXPECT_TRUE(std::is_trivially_copyable<
                 vecmem::data::jagged_vector_view<const int> >());
 
     EXPECT_TRUE(std::is_trivially_default_constructible<
@@ -53,6 +54,18 @@ TEST_F(core_device_container_test, trivial_construct) {
                 vecmem::data::jagged_vector_view<int> >());
     EXPECT_TRUE(std::is_trivially_copy_constructible<
                 vecmem::data::jagged_vector_view<int> >());
+    EXPECT_TRUE(
+        std::is_trivially_copyable<vecmem::data::jagged_vector_view<int> >());
+
+    bool helper = std::is_assignable<vecmem::data::jagged_vector_view<int>,
+                                     vecmem::data::jagged_vector_view<int> >();
+    EXPECT_TRUE(helper);
+    helper = std::is_assignable<vecmem::data::jagged_vector_view<const int>,
+                                vecmem::data::jagged_vector_view<const int> >();
+    EXPECT_TRUE(helper);
+    helper = std::is_assignable<vecmem::data::jagged_vector_view<const int>,
+                                vecmem::data::jagged_vector_view<int> >();
+    EXPECT_TRUE(helper);
 
     EXPECT_TRUE(std::is_trivially_default_constructible<
                 vecmem::data::vector_view<const int> >());
@@ -60,6 +73,8 @@ TEST_F(core_device_container_test, trivial_construct) {
                 vecmem::data::vector_view<const int> >());
     EXPECT_TRUE(std::is_trivially_copy_constructible<
                 vecmem::data::vector_view<const int> >());
+    EXPECT_TRUE(
+        std::is_trivially_copyable<vecmem::data::vector_view<const int> >());
 
     EXPECT_TRUE(std::is_trivially_default_constructible<
                 vecmem::data::vector_view<int> >());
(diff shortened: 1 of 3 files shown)


CrossR:
I've bumped into this before, and actually there is at least one other place in the `traccc::alpaka` code this was an issue...however, I went the route of just adding a trait to tell Alpaka "no trust me this is trivially copyable". (More info on someone else hitting the same thing here: https://github.com/alpaka-group/alpaka/issues/2195).

https://github.com/acts-project/traccc/blob/369caa901d771fa14fad2a6aba8dcfca50addd77/device/alpaka/src/finding/finding_algorithm.cpp#L494-L502

I'm glad at least that it seems like following this (somewhat annoying) rabbit hole has lead to more "correct" code at least, but it is a pain, especially as you say when the three other back ends we care about have zero issue with all these types.

core/include/vecmem/utils/impl/tuple.ipp
```diff
@@ -55,7 +55,7 @@ struct tuple_get_impl<0> {
 }  // namespace details
 
 template <std::size_t I, typename... Ts>
-VECMEM_HOST_AND_DEVICE inline constexpr const auto &get(
+VECMEM_HOST_AND_DEVICE constexpr const auto &get(
```
stephenswat, line 58: Any particular reason to get rid of these inlines?

core/include/vecmem/utils/tuple.hpp
```diff
@@ -79,8 +80,8 @@ struct tuple<T, Ts...> {
         std::enable_if_t<
             vecmem::details::conjunction<
                 vecmem::details::negation<std::is_same<tuple<T, Ts...>, U>>,
-                std::is_constructible<T, U &&>,
-                std::is_constructible<Ts, Us &&>...>::value,
```
stephenswat, line 83: Why are you removing the rvalue reference markers here?

core/include/vecmem/utils/tuple.hpp
```diff
@@ -95,36 +96,28 @@ struct tuple<T, Ts...> {
     ///
     template <typename U, typename... Us,
               std::enable_if_t<vecmem::details::conjunction<
-                                   std::is_constructible<T, U &&>,
-                                   std::is_constructible<Ts, Us &&>...>::value,
```
stephenswat, line 99: Same as above.

core/include/vecmem/utils/impl/tuple.ipp
krasznaa: SonarCloud seems very adamant about this issue.

https://sonarcloud.io/project/issues?issues=AZMFu6RHKtxsJUM8k-A0&open=AZMFu6RHKtxsJUM8k-A0&id=acts-project_vecmem

I was just trusting it on this one for the best practice.

core/include/vecmem/utils/tuple.hpp
krasznaa: I was going back and forth with this one. :thinking: It's the same discussion from before. That especially in this constructor we use universal references. So most of the time we don't actually use rvalue references, bur rather lvalue ones.

Is `U` and `Us` not the correct type of reference as-is in this specific function? :thinking:

core/include/vecmem/utils/tuple.hpp
krasznaa: On this one I'm less sure. :frowning: The `tail` variable we definitely take as an rvalue reference. But `head` is still a universal reference. So I'm honestly not sure what the most correct test is here.

With all this said, I don't feel all too strongly about these, so the rvalue references might as well go back.

core/include/vecmem/utils/tuple.hpp
krasznaa: I was trying to figure more out with the help of Google, but didn't get much further. :frowning: Since I found a couple of examples of people using `U&&` in similar situations, I went back to using that. Even if I'm still not entirely sure that that's absolutely the most correct. :thinking:

But at least this formalism allows both this project and [traccc](https://github.com/acts-project/traccc) to build everything successfully as much as I could test it. (As did the formalism without `&&`...) So let's use this one.

stephenswat: Indeed `U&&` is the correct solution. `std::forward<T>` always returns `T&&` so there really is no valid reason for anything but `U&&` to be correct. :confused:

PR #333, opened 2025-05-22
Add memory usage reporting to buddy allocator

stephenswat:
This PR equips the binary page memory resource with functions that allow users to inspect the occupied and total memory held by an allocator.


krasznaa (commented):
This is still a draft, right? We still need a public interface to access this information. Or did I miss something?

stephenswat:
> Or did I miss something?

Nah this is really just for internal use, for making the memory plots for traccc.

PR #343, opened 2025-12-09
Optimal Size Copy, main branch (2025.12.09.)

krasznaa:
This came about following the optimizations that @stephenswat has put into [traccc](https://github.com/acts-project/traccc) for getting the sizes of device buffers asynchronously. See:

https://github.com/acts-project/traccc/blob/main/device/cuda/src/utils/get_size.hpp

I believe `vecmem::copy` is a better place for doing this. Since as long as one gives it a specific memory resource that it should use for the copy it does internally, one can use pretty much the same API that we always had for `copy::get_size(...)` and `copy::get_sizes(...)`.

### DIFF
--- a/tests/common/copy_tests.ipp
+++ b/tests/common/copy_tests.ipp
@@ -1,7 +1,7 @@
 /*
  * VecMem project, part of the ACTS project (R&D line)
  *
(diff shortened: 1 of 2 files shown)


krasznaa:
Superseded by #344.

PR #356, opened 2026-03-25
Disable collecting timing information for cuda and hip events 

m-fila:
The CUDA and HIP events in vecmem are used only for synchronization, not for timing, and are never leaked to the outside. Both [CUDA](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__EVENT.html#group__CUDART__EVENT_1g7b317e07ff385d85aa656204b971a042) and [HIP](https://rocm.docs.amd.com/projects/HIP/en/latest/doxygen/html/group___event.html#gae86a5acb1b22b61bc9ecb9c28fc71b75) documentation mentions that if not needed collecting timing information can be disabled to possibly give better performance

Having said that, I don't see noticeable performance improvement for traccc main branch with CUDA 12.8 on neither Nvidia RTX 3060 nor L40S.

### DIFF
--- a/cuda/src/utils/event_pool.cpp
+++ b/cuda/src/utils/event_pool.cpp
@@ -41,7 +41,8 @@ event_pool::event_pool(std::size_t size) : m_pool(size), m_used_events(0) {
 
     // Create the (initial) events in the pool.
     for (cudaEvent_t& e : m_pool) {
-        VECMEM_CUDA_ERROR_CHECK(cudaEventCreate(&e));
+        VECMEM_CUDA_ERROR_CHECK(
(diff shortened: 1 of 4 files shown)


krasznaa (changes requested):
Thanks for opening the PR! 😄

I'm really not sure about the best flags to use. I guess in the end we may want to allow the `async_copy` constructors to receive an `unsinged int flags` argument, which the user could set themselves if they wanted to. 🤔 Until we do that, I guess using this PR's flag setup would be reasonable.

Just to please my personal preferences, please just change the API of the internal `create_event()` functions. Then we could get this in.

krasznaa (approved):
Let's get this finally in. 😄

PR #358, opened 2026-04-07
Add methods for non-blocking event completion

m-fila:
This change complements the existing blocking/waiting interface of `event` by introducing a non-blocking query mechanism.

Adding `is_ready` method to query completion status of events. In case of synchronous, noop-events it always returns `true`, for asynchronous events it calls corresponding "query" function from the underlying library.

Also, adding `unsafe_get` to `async_size` and `async_sizes` to get the value without waiting on an event in cases when user can guarantee it was already synchronized by some other means (such as querying, synchronizing the whole stream, device).
I was also thinking about adding (or replacing `unsafe_get` with) something like `bool try_get(reference)` which would be "safe" at the cost of always making an extra query

### DIFF
--- a/core/include/vecmem/utils/abstract_event.hpp
+++ b/core/include/vecmem/utils/abstract_event.hpp
@@ -1,7 +1,7 @@
 /*
  * VecMem project, part of the ACTS project (R&D line)
  *
- * (c) 2022-2025 CERN for the benefit of the ACTS project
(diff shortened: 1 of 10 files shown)


krasznaa (commented):
I like it. Apart from some extra assertions, I'm pretty happy with the introduced logic and test. 👍

krasznaa (approved):
👍

PR #359, opened 2026-04-15
CUDA/HIP Event Creation Flag Specification, main branch (2026.04.15.)

krasznaa:
Building on the improvements introduced by @m-fila earlier, I made it possible to specify event flags for `vecmem::cuda::async_copy` and `vecmem::hip::async_copy`.

This should practically never be necessary, but client code **may** want to experiment with using different types of CUDA/HIP events behind the scenes.

### DIFF
--- a/cuda/include/vecmem/utils/cuda/async_copy.hpp
+++ b/cuda/include/vecmem/utils/cuda/async_copy.hpp
@@ -33,9 +33,23 @@ namespace vecmem::cuda {
 class async_copy : public vecmem::copy {
 
 public:
-    /// Constructor with the stream to operate on
+    /// Constructor with the CUDA stream to operate on
+    ///
+    /// @param stream The CUDA stream to perform the copies in
+    ///
     VECMEM_CUDA_EXPORT
     explicit async_copy(const stream_wrapper& stream);
+    /// Constructor with a stream and flags to create CUDA events with
+    ///
+    /// For details on the flags, see the documentation
+    /// of @c cudaEventCreateWithFlags(...) on:
+    /// https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__EVENT.html
+    ///
+    /// @param stream The CUDA stream to perform the copies in
+    /// @param event_flags Flag(s) to create internal CUDA events with
+    ///
+    VECMEM_CUDA_EXPORT
+    async_copy(const stream_wrapper& stream, unsigned int event_flags);
     /// Move constructor
     VECMEM_CUDA_EXPORT
     async_copy(async_copy&&) noexcept;
--- a/cuda/src/utils/event_pool.cpp
+++ b/cuda/src/utils/event_pool.cpp
@@ -35,9 +35,8 @@
 
 namespace {
 cudaEvent_t create_event(unsigned int flags) {
+    // "Safely" create a CUDA event with the specified flags.
     cudaEvent_t event;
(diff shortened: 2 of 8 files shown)


cuda/src/utils/event_pool.cpp
```diff
@@ -34,12 +34,11 @@
 #include <stdexcept>
 
 namespace {
-cudaEvent_t create_event() {
+cudaEvent_t create_event(unsigned int flags) {
     cudaEvent_t event;
     // Disable collecting timing information since the event is used only for
     // synchronization.
-    VECMEM_CUDA_ERROR_CHECK(
```
m-fila, line 41: This comment should be probably removed since this isn't timing specific anymore

hip/src/utils/event_pool.cpp
```diff
@@ -34,12 +34,11 @@
 #include <stdexcept>
 
 namespace {
-hipEvent_t create_event() {
+hipEvent_t create_event(unsigned int flags) {
     hipEvent_t event;
     // Disable collecting timing information since the event is used only for
     // synchronization.
```
m-fila: The comment probably should be removed

cuda/include/vecmem/utils/cuda/async_copy.hpp
```diff
@@ -33,9 +33,23 @@ namespace vecmem::cuda {
...
+    ///
+    /// For details on the flags, see the documentation
+    /// of @c cudaEventCreateWithFlags(...) on:
+    /// https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__EVENT.html
+    ///
+    /// @param stream The CUDA stream to perform the copies in
+    /// @param event_flags Flag(s) to create internal CUDA events with
+    ///
+    VECMEM_CUDA_EXPORT
+    async_copy(const stream_wrapper& stream, unsigned int event_flags);
```
m-fila, line 52: Would it make sense to explicitly mention that `cudaEventBlockingSync` can be used to change the behaviour of `wait` method of events produced by this object? I imagine this will be the most common use cases for flags

m-fila:
For reference, I wrote something similar for experiments with scheduling https://github.com/cern-nextgen/wp1.7-vecmem/pull/2 .

I thought about more constrained API since the cuda/hip events are never exposed directly to a client and client won't know how the events are used internally (for example, is it safe to disable timing or are there any internal calls to `cudaEventElapsedTime` that will break?).
From my understanding the only customization relevant to the users is whether the `wait` is blocking or spinning, the rest depend on how the events are used internally by vecmem.

cuda/src/utils/event_pool.cpp
krasznaa: 👍

hip/src/utils/event_pool.cpp
krasznaa: 👍

cuda/include/vecmem/utils/cuda/async_copy.hpp
krasznaa: I was a bit hesitant about even be as detailed as I ended up being. 🤔

I'd rather not be too specific about the flags that one can use. I think just pointing at the CUDA and HIP documentation should be enough. My idea being that they may very well introduce additional flags in the future as well. So I'd rather just point at that webpage, and not refer to flags that may get outdated at one point.

It wasn't a bad suggestion, but I think I'll leave the documentation like this.

PR #360, opened 2026-04-24
Adding atomic max and min support

mimodekjaer:
I have added `fetch_max` and `fetch_min`, to the atomic.

For CUDA, HIP and SYCL, this is already provided, while it is added in c++ std 26 for CPU, which is why this is made using the `compare_exchange`.


core/include/vecmem/memory/impl/atomic.ipp
```diff
@@ -183,4 +183,40 @@ VECMEM_HOST_AND_DEVICE auto atomic<T>::fetch_xor(value_type data)
...
+#elif defined(CL_SYCL_LANGUAGE_VERSION) || defined(SYCL_LANGUAGE_VERSION)
+    return __VECMEM_SYCL_ATOMIC_CALL1(fetch_max, m_ptr, data);
+#else
+    const value_type result = *m_ptr;
+    if (*m_ptr < data) {
+        *m_ptr = data;
+    }
+    return result;
+#endif
+}
```
krasznaa: Let's not bother about `vecmem::atomic`. 🤔 This class is deprecated since a while. No need to update it for this feature.

tests/core/test_core_atomic_ref.cpp
```diff
@@ -104,3 +104,31 @@ TEST(core_device_atomic_test, atomic_fetch_xor) {
...
+    int i = 0;
+    vecmem::device_atomic_ref<int> a(i);
+    ASSERT_EQ(a.load(), 0);
+    ASSERT_EQ(a.fetch_max(5), 0);
+    ASSERT_EQ(a.load(), 5);
+    ASSERT_EQ(a.fetch_max(3), 5);
+    ASSERT_EQ(a.load(), 5);
+#else
+    GTEST_SKIP() << "atomic_fetch_max not supported";
+#endif
```
krasznaa: Curious that you had to worry about this in the test. 🤔

By default we should be using C\+\+17 to build the host code for the tests. So `std::atomic_ref` should not even be used here. The test unfortunately really only exercises our own `vecmem::posix_device_atomic_ref` code. 😦 

But we should at least exercise that. 🤔 If you want to be super safe, you might want to disable these tests just for C\+\+20 and C\+\+23. But for other standards we would expect it to work, right?

krasznaa (commented):
This would be needed for the GBTS code I guess?

I forgot that we don't have these tests for HIP and SYCL. 😦 Could you add those as well while you're at it?

core/include/vecmem/memory/impl/atomic.ipp
mimodekjaer: Okay, do you want me to remove the file all together or just the changes?

tests/core/test_core_atomic_ref.cpp
mimodekjaer: Yes it is only an issue when building on CPU for C++20 and C++23, assuming that the fall back works correctly. Because from what I understand vecmem supports windows which does not support posix.

So either we accept that `fetch_max` and `fetch_min` is not supported at all on C++20 and C++23 or implementation will prioritize the posix verison on CPU for the  C++20 and C++23, so the it limits the difference to only versions that does not support posix.

For me it makes no difference it is up to you and I will gladly add the way you think is better.

core/include/vecmem/memory/impl/atomic.ipp
krasznaa: Let's remove the type in a separate PR. In this one you should just not update/touch it.

tests/core/test_core_atomic_ref.cpp
krasznaa: Hmm... I don't think there is a clearly good or bad solution here. 🤔

But maybe the most reasonable logic could be that in host code we use
  - `std::atomic_ref` with C\+\+23\+;
  - `vecmem::posix_device_atomic_ref` if it's usable;
  - `vecmem::dummy_device_atomic_ref` if none of the above.

I think this would allow client code in all circumstances to start using `fetch_min(...)` and `fetch_max(...)`, and get reasonable results.

core/include/vecmem/memory/device_atomic_ref.hpp
```diff
@@ -95,6 +95,16 @@ using device_atomic_ref = posix_device_atomic_ref<T, address>;
...
+#elif defined(__cpp_lib_atomic_ref)
+
+namespace vecmem {
+
+/// Use @c std::atomic_ref in host code with C++20 and C++23
+template <typename T, device_address_space = device_address_space::global>
+using device_atomic_ref = std::atomic_ref<T>;
+
+}  // namespace vecmem
+
```
krasznaa: I'm very confused about this part. 😕

I believe you correctly modified line 74. That if we have `fetch_min` and `fetch_max` available on `std::atomic_ref`, then we should use that type in host code.

But if we don't have min/max available from `std::atomic_ref`, and we're not on a POSIX system, then we should fall back on `vecmem::dummy_device_atomic_ref`. Which will not perform atomic operations, but will at least allow client code to compile. (And these types are anyway never really used in host code to begin with.)

tests/core/test_core_atomic_ref.cpp
```diff
@@ -104,3 +104,33 @@ TEST(core_device_atomic_test, atomic_fetch_xor) {
     ASSERT_EQ(a.fetch_xor(0b0000), 0b1001);
     ASSERT_EQ(a.load(), 0b1001);
 }
+
+TEST(core_device_atomic_test, atomic_fetch_max) {
+#if defined(VECMEM_SUPPORT_POSIX_ATOMIC_REF) || \
+    defined(__cpp_lib_atomic_min_max)
```
krasznaa: We shouldn't need any of these checks here as long as the logic in `device_atomic_ref.hpp` is done correctly. 🤔  As the test should just fall back on the dummy version when we're not on POSIX.

The whole point of my complaining was so that these unit tests would not need any preprocessor protections.

krasznaa (changes requested):
And to make your life even more miserable, please also update

https://github.com/acts-project/vecmem/blob/main/core/include/vecmem/concepts/atomic_ref.hpp

with the ``fetch_min` / `fetch_max` functions. To ensure that we would always provide a type to the user that would have these functions.

core/include/vecmem/memory/device_atomic_ref.hpp
mimodekjaer: Sorry I just misunderstood you, I will make the change.

tests/core/test_core_atomic_ref.cpp
stephenswat: Does `posix_device_atomic_ref` work on Windows? Looking at the implementation it is actually poorly named because it has nothing to do with POSIX at all. But if this thing doesn't work on Windows then you are effectively suggesting to end Windows support for vecmem. FYI.

And looking at past Windows CI runs, it doesn't seem to:

```
-- Performing Test VECMEM_SUPPORT_POSIX_ATOMIC_REF - Failed
```

tests/core/test_core_atomic_ref.cpp
stephenswat: Again, unless you want to definitively end support for Windows, this is a huge mistake. The dummy atomic reference, if anything, should be removed and replaced by a compiler error; the last thing we need to be doing is to make it _more_ likely that the user quietly receives a broken atomic implementation!

tests/core/test_core_atomic_ref.cpp
krasznaa: The Windows build gives me both good input about the build system of the project, and it gives a good set of extra checks that GCC and Clang don't do. So the Windows build is not going anywhere.

I half-intend to look into a `vecmem::win32_device_atomic_ref` at one point. But it's just really not a priority at the moment.

Until then, I myself am not offended by the Windows build being a bit more fragile than the other ones. Since again, the hose version of this type is really only there for feature completeness. It's not there to provide serious multi-threading for host code.

krasznaa (approved):
I'm happy with it looking like this. 👍

PR #361, opened 2026-04-24
AdaptiveCpp Support, main branch (2026.04.24.)

krasznaa:
[AdaptiveCpp](https://github.com/AdaptiveCpp/AdaptiveCpp) uses a different set of macros than [oneAPI](https://www.intel.com/content/www/us/en/developer/tools/oneapi/overview.html) does. This update makes sure that the "generic device code" that we have in [traccc](https://github.com/acts-project/traccc), would work correctly with AdapriveCpp.

### DIFF
--- a/core/include/vecmem/memory/device_atomic_ref.hpp
+++ b/core/include/vecmem/memory/device_atomic_ref.hpp
@@ -1,7 +1,7 @@
 /*
  * VecMem project, part of the ACTS project (R&D line)
  *
- * (c) 2022-2024 CERN for the benefit of the ACTS project
+ * (c) 2022-2026 CERN for the benefit of the ACTS project
  *
  * Mozilla Public License Version 2.0
  */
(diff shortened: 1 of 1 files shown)


krasznaa:
Let me also ping @illuhad here. 😄 This was one of the simpler things that I had to fix.

illuhad:
AdaptiveCpp also defines these two macros, so in principle the workaround should not be necessary. However, the SYCL specification does not guarantee that these (and any) macros are available without including `sycl.hpp`. AdaptiveCpp defines them in this header, DPC++ (presumably) makes them globally available via the compiler. Both strategies are legal in SYCL.

-> the solution best in line with the SYCL specification would likely be to just include `sycl.hpp` in this file. However, I can understand if you want to e.g. avoid the compile time hit.

krasznaa:
Thanks for the clarification!

Since the header I'm modifying here is meant to work "in any situation", it would be up to the user to make sure that they always include `<sycl/sycl.hpp>` before this header. And that's usually not the order that we use in our own projects. 🤔 

So I'll still go ahead with this update. But it's **very** good to know that (as I thought) the macros that our code was already using, are pretty much the "standard ones". So I was indeed wondering why AdaptiveCpp would not define them. Now I understand what happened. 👍

PR #362, opened 2026-05-21
Fix use of aliased targets in installed code

stephenswat:
Currently, the CMake configuration code tries to set compile definitions on aliased targets such as `vecmem::core` and `vecmem::sycl`, but this is not allowed. Instead, we must retrieve the unaliased target name and apply the definitions directly to that target.

### DIFF
--- a/core/cmake/vecmem-setup-core.cmake
+++ b/core/cmake/vecmem-setup-core.cmake
@@ -20,6 +20,14 @@ function( vecmem_setup_core libName )
    # Check if vecmem::posix_device_atomic_ref is usable.
    get_target_property( _coreIncludes "${libName}"
       INTERFACE_INCLUDE_DIRECTORIES )
+
+   get_property( aliasedLibName TARGET "${libName}" PROPERTY ALIASED_TARGET )
+   if( "${aliasedLibName}" STREQUAL "" )
+      set( unaliasedLibName "${libName}" )
+   else()
+      set( unaliasedLibName "${aliasedLibName}" )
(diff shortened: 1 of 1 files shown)


stephenswat:
This currently renders a broad range of vecmem installations unusable, including the ones I would like to use in the ACTS integration.

krasznaa:
What exact issue do you see? 😕 

That function, while building the core library, is called on `vecmem_core`.

https://github.com/acts-project/vecmem/blob/main/core/CMakeLists.txt#L212

I.e. the concrete, non-aliased name.

It is when using an installed version of vecmem that it gets called on `vecmem::core`.

https://github.com/acts-project/vecmem/blob/main/cmake/vecmem-config.cmake.in#L65

But there `vecmem::core` is the "real" name of the imported library it's not an alias.

So... what exact issue do you see? Because Athena has been using "installed vecmem" happily since years. 🤔

stephenswat:
```
CMake Error at /acts_install/lib/cmake/vecmem-1.24.0/vecmem-setup-core.cmake:34 (target_compile_definitions):
  target_compile_definitions can not be used on an ALIAS target.
  /acts_install/lib/cmake/vecmem-1.24.0/vecmem-config.cmake:79 (vecmem_setup_core)
```

stephenswat:
Okay I think I've found the issue here. :)

stephenswat:
Was a mix of an installed vecmem and a FetchContent vecmem. Lovely that that is even possible. 🙃

PR #363, opened 2026-05-23
Install HIP toolkit CMake code with vecmem

stephenswat:
Currently, the vecmem packaging code does not install the `FindHIPToolkit.cmake` file along with the rest of the CMake code. This means that any code using `find_package(HIPToolkit)` will break when using an installed version of vecmem. This currently breaks the Alpaka HIP-SYCL build in traccc.


krasznaa (approved):
Good catch! 👍

PR #364, opened 2026-08-08
Make jagged container size types consistent with non-jagged ones

### LINKED ISSUE #96: Size type of `vecmem::jagged_device_vector` should be `unsigned int`

manishpaulish:
Closes #96.

`vecmem::jagged_device_vector` and `vecmem::data::jagged_vector_view` used
`std::size_t` as their `size_type`, while `vecmem::device_vector` and
`vecmem::data::vector_view` use `unsigned int`.

Changing the two typedefs on their own builds clean and the whole suite passes,
which is misleading, the affected code is in `.ipp` templates that only
instantiate when something uses them. Adding a test that pins the types forces
instantiation, and surfaces the conversions the wider typedef had been hiding:

| Location | Diagnostic |
|---|---|
| `jagged_vector.ipp` — outer size, all four `get_data` overloads | `-Wconversion` |
| `jagged_vector_buffer.ipp:69,71,73` — `capacities.size()` | `-Wconversion` |
| `jagged_vector_buffer.ipp:102` — same, inside a braced initialiser | `-Wnarrowing` |
| `edm/details/data_traits.hpp:52` — braced initialiser | `-Wnarrowing` |

Per @krasznaa's explanation on the issue, 32-bit atomics are guaranteed on
CUDA/HIP/SYCL, `std::size_t` isn't, the narrowing itself is correct, so I've
kept it and made it explicit rather than removing it.
`vecmem::details::narrow_size()` carries that rationale in its documentation and
asserts in debug builds that the value fits; a jagged vector with more than
`UINT_MAX` rows isn't a supported configuration.

Also adds a `static_assert` test that the jagged and non-jagged size types agree,
so they can't drift apart again without the suite noticing.

**Two things I'd welcome direction on:**

1. I used a shared `narrow_size` helper rather than a `static_cast` at each of
   the six sites, on the grounds that the reason for the narrowing is worth
   stating once. Happy to switch to plain casts if you'd prefer less machinery.
2. I've only built and tested the core library, no CUDA/HIP/SYCL setup here, so
   the device backends are untested beyond compiling against the changed headers.
   Relying on CI for those.

Verified with GCC 11 (Linux) and AppleClang 17 (macOS): full build with no new
warnings, 296/296 tests passing, clang-format v10.0.1 clean.

### DIFF
--- a/core/CMakeLists.txt
+++ b/core/CMakeLists.txt
@@ -172,6 +172,7 @@ vecmem_add_library( vecmem_core core
    "include/vecmem/utils/tuple.hpp"
    "include/vecmem/utils/impl/tuple.ipp"
    "include/vecmem/utils/type_traits.hpp"
+   "include/vecmem/utils/details/narrow_size.hpp"
    "include/vecmem/utils/types.hpp"
    "src/utils/integer_math.hpp" )
 
--- a/core/include/vecmem/containers/data/jagged_vector_view.hpp
+++ b/core/include/vecmem/containers/data/jagged_vector_view.hpp
@@ -50,7 +50,7 @@ class jagged_vector_view {
 
 public:
     /// Size type used in the class
-    using size_type = std::size_t;
+    using size_type = unsigned int;
     /// Value type of the jagged array
     using value_type = vector_view<T>;
     /// Pointer type to the jagged array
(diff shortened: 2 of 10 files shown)


krasznaa (approved):
With a bit longer of a delay...

I like the changes. 🤔 Downstream projects will probably need to adjust themselves a bit, but this inconsistency between the types has bothered us for long enough by now.

Let's indeed try to fix it finally! 😉

Issue #348, opened 2026-01-16
Access to private members when constructing device vector from const view


niermann999:
The member variables accessed in the constructor https://github.com/acts-project/vecmem/blob/521e240a55abe0385d256165581291af72875e9f/core/include/vecmem/containers/impl/device_vector.ipp#L36C25-L36C35 are private in the vector of different type `device_vector<OTHERTYPE>`, leading to errors such as:

````
traccc/build/_deps/vecmem-src/core/include/vecmem/containers/impl/device_vector.ipp:36:25: error: ‘vecmem::device_vector<const unsigned int>::size_type vecmem::device_vector<const unsigned int>::m_capacity’ is private within this context
   36 |     : m_capacity(parent.m_capacity),
```

krasznaa:
With the help of @manishpaulish, this should be fixed now. 🤞

PR #366, opened 2026-08-13
Device Vector Cross-Construct, main branch (2026.08.13.)

krasznaa:
Added a test for constructing `const` device vectors from non-`const` ones. And to cross check, making sure that a non-`const` vector could not be made from a `const` one.

This is to follow up on issue #348. Which I couldn't quite re-produce. :frowning: @niermann999, do you have a case in which you still see this issue showing up?

And pre-commit / clang-format changing a whole lot of lines in the test is not all that great... :frowning:

### DIFF
--- a/core/include/vecmem/containers/device_vector.hpp
+++ b/core/include/vecmem/containers/device_vector.hpp
@@ -1,6 +1,6 @@
 /* VecMem project, part of the ACTS project (R&D line)
  *
- * (c) 2021-2025 CERN for the benefit of the ACTS project
+ * (c) 2021-2026 CERN for the benefit of the ACTS project
  *
  * Mozilla Public License Version 2.0
  */
@@ -84,7 +84,7 @@ public:
     device_vector(const device_vector& parent) = default;
     /// Copy constructor
     template <typename OTHERTYPE,
-              std::enable_if_t<std::is_convertible<OTHERTYPE, TYPE>::value,
+              std::enable_if_t<details::is_same_nc<TYPE, OTHERTYPE>::value,
                                bool> = true>
     VECMEM_HOST_AND_DEVICE device_vector(
         const device_vector<OTHERTYPE>& parent);
--- a/core/include/vecmem/containers/device_vector.hpp
+++ b/core/include/vecmem/containers/device_vector.hpp
@@ -35,6 +35,9 @@ class device;
 template <typename TYPE>
 class device_vector {
 
+    // Make other specializations of the class a friend of this class.
+    template <typename OTHERTYPE>
+    friend class device_vector;
     // Make @c vecmem::edm::device a friend of this class.
     template <typename T, template <typename> class I>
     friend class edm::device;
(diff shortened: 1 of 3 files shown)


manishpaulish:
Hi @krasznaa, you asked for a case that still shows the issue, so here is one. I had a fix for #348 in progress when you opened this, so I tested your branch against my reproducer before dropping mine. Two things came out of it that I think are worth flagging.

**Tightening the constraint to `is_same_nc` looks right to me.** `is_convertible` was letting through conversions that should never have been allowed, and the `EXPECT_FALSE(helper3)` case is a good catch.

**But I do not think it fixes the access error from the issue.** The constraint controls which instantiations are viable; the failure is that the constructor body reads `parent.m_capacity`, `parent.m_size` and `parent.m_ptr` from a different instantiation of the class template, and those are still private in that context. On your branch:

```cpp
unsigned int storage[4] = {1, 2, 3, 4};
vecmem::data::vector_view<unsigned int> view{4, storage};
vecmem::device_vector<unsigned int> mutable_vec{view};
vecmem::device_vector<const unsigned int> const_vec{mutable_vec};
```

```
device_vector.ipp:36:25: error: 'vecmem::device_vector<unsigned int>::m_capacity' is private within this context
device_vector.ipp:37:21: error: 'vecmem::device_vector<unsigned int>::m_size' is private within this context
device_vector.ipp:38:20: error: 'vecmem::device_vector<unsigned int>::m_ptr' is private within this context
```

**The reason the new test does not catch it** is that `std::is_constructible` only checks the declaration. It never instantiates the constructor body, so the private access is never reached. In a single translation unit on your branch, the `static_assert` on `is_constructible` passes and the actual construction of the same pair of types fails, which I think makes it a false green.

A test that constructs rather than asks the trait would cover it:

```cpp
vecmem::device_vector<int> mutable_vector{buffer};
vecmem::device_vector<const int> const_vector{mutable_vector};
EXPECT_EQ(const_vector.size(), mutable_vector.size());
```

What I had locally was a friend declaration, which fixes the access directly:

```cpp
/// Make every instantiation of this class a friend of every other one.
template <typename OTHERTYPE>
friend class device_vector;
```

Combined with your `is_same_nc` change that seems like the complete fix to me, though you may well prefer the accessor route that `reverse_iterator` uses.

**One more thing, separate from this PR.** `vecmem::details::jagged_device_vector_iterator` has the same converting constructor reading `parent.m_ptr`, with no friend declaration either. It fails twice over:

```
jagged_device_vector_iterator.ipp:65:20: error: 'jagged_device_vector_iterator<int>::m_ptr' is private within this context
jagged_device_vector_iterator.ipp:65:20: error: cannot convert 'const vector_view<int>* const' to 'const vector_view<const int>*'
```

The second error is the same const-qualifying pointer conversion that the sibling constructor taking a `vector_view` already handles with a documented `reinterpret_cast`. I have a fix and a test for it if that would be useful, either folded into this PR or as a separate one, whichever you prefer.

For what it is worth, `reverse_iterator` has the same shape of converting constructor but reads the parent through the public `base()`, so it was never affected.

krasznaa:
Thanks for all the description @manishpaulish. 😄 

Please do go ahead an open any number of additional PRs with improvement proposals. I can easily believe that such bugs could still exist in some of the classes.

PR #368, opened 2026-08-17
Migrate GPU CI from GitLab to GitHub Actions

paulgessinger:
Replaces `.gitlab-ci.yml` with `.github/workflows/gpu.yml`, keeping the same two-stage structure: build the CUDA and SYCL-on-NVIDIA configurations on a regular runner, then run the GPU tests on a self-hosted machine labelled `gpu-nvidia`.

| GitLab job | GitHub job |
|---|---|
| `build_cuda` (tags: `docker`) | `build (CUDA)` on `ubuntu-latest`, container `ubuntu2004_cuda:47` |
| `test_cuda` (tags: `docker-gpu-nvidia`) | `test (CUDA)` on `[gpu-nvidia, self-hosted]`, `ctest -R ".*cuda.*"` |
| `build_cuda_sycl` | `build (SYCL)`, container `ubuntu2004_cuda_oneapi:47` |
| `test_cuda_sycl` | `test (SYCL)`, `ctest -R ".*sycl.*"` |

A few adaptations were needed, as a literal translation would not have worked:

- The manual clone of `$CLONE_URL` is replaced by `actions/checkout`. The test job checks out as well, because the build tree refers to the source with absolute paths; running both jobs in containers keeps the workspace at the same path on both sides.
- The build directory is transferred as a tarball, since artifact up-/download does not preserve the executable bit that the test binaries need.
- The container images are bumped from the stale `v13`/`v18` tags to the `:47` tags already used by `builds.yml`.
- The deprecated `SYCL_DEVICE_FILTER: cuda` is replaced by `ONEAPI_DEVICE_SELECTOR=cuda:gpu`, exported after sourcing `ci_setup.sh`, which otherwise pins the selector to `opencl:cpu`.
- Both jobs reuse `.github/ci_setup.sh`, consistent with `builds.yml`.

Two points worth a reviewer's attention:

1. **GPU passthrough.** The test job sets `container.options: --gpus all`, which assumes the nvidia-container-toolkit is available on the runner host.
2. **Fork PRs on self-hosted hardware.** The workflow keeps the repository's `on: [push, pull_request]` convention, which means code from fork PRs can execute on the self-hosted GPU machine. We gate external contributors behind the approval button, and the runners are on ephemeral VMs.


.github/workflows/gpu.yml
```diff
@@ -0,0 +1,144 @@
...
+  # Build the GPU code on a "regular" runner, and hand the build directory
+  # over to the GPU runner for the actual test execution.
+  build:
+
+    strategy:
+      fail-fast: false
+      matrix:
+        PLATFORM:
+          - NAME: "CUDA"
+            CONTAINER: "ghcr.io/acts-project/ubuntu2004_cuda:47"
```
krasznaa: I would prefer to use a single version of all of the images. These can get out of sync over time. But at least at the start they could all be the same. (`51`)

.github/workflows/gpu.yml
```diff
@@ -0,0 +1,144 @@
...
+    steps:
+    - uses: actions/checkout@v4
+    # Run the CMake configuration.
+    - name: Configure
+      run: |
+        source ${GITHUB_WORKSPACE}/.github/ci_setup.sh ${{ matrix.PLATFORM.NAME }}
+        cmake \
+          -DCMAKE_BUILD_TYPE=Release \
+          -DVECMEM_BUILD_${{ matrix.PLATFORM.NAME }}_LIBRARY=TRUE \
+          -DVECMEM_BUILD_BENCHMARKING=TRUE \
```
krasznaa, line 60: Should we run the benchmark executables? 🤔 It may not be a bad idea.

But as long as we don't do that, we may as well just skip building the binaries. We test their build in the "regular" CI jobs already.

krasznaa (commented):
Thanks for the PR! Just some small comments from my side.

.github/workflows/gpu.yml
paulgessinger: So I was trying to not move them too much at this time, but `51` might work.

.github/workflows/gpu.yml
paulgessinger: How long do they usually run?

.github/workflows/builds.yml
```diff
@@ -81,28 +81,28 @@ jobs:
             MSG_LVL: 5
         PLATFORM:
           - NAME: "CPU"
-            CONTAINER: "ghcr.io/acts-project/ubuntu2404:47"
+            CONTAINER: "ghcr.io/acts-project/ubuntu2404:51"
```
krasznaa: This is not what I meant. 😦 I was only talking about the jobs that you are introducing now.

I would not like to change `builds.yml`.

.github/workflows/gpu.yml
```diff
@@ -0,0 +1,153 @@
...
+        cd build
+        source ${GITHUB_WORKSPACE}/.github/ci_setup.sh ${{ matrix.PLATFORM.NAME }}
+        ctest --output-on-failure -R "${{ matrix.PLATFORM.CTEST_REGEX }}"
+    # Run the benchmark(s) on the GPU. The measurement time is cut short on
+    # purpose. The goal here is to exercise the benchmark code on a real device,
+    # not to produce meaningful numbers on a machine shared with other jobs.
+    - name: Benchmark
+      run: |
+        source ${GITHUB_WORKSPACE}/.github/ci_setup.sh ${{ matrix.PLATFORM.NAME }}
+        ./build/bin/${{ matrix.PLATFORM.BENCHMARK }} --benchmark_min_time=0.01
```
krasznaa: Would it not be possible to derive the benchmark name from `matrix.PLATFORM.NAME`? Some formalism must surely exist. 🤔

.github/workflows/gpu.yml
krasznaa: To be seen. They can take a few minutes as I remember.

krasznaa (changes requested):

.github/workflows/builds.yml
paulgessinger: Ah, wouldn't it be best to keep all jobs at a single tag though?

.github/workflows/gpu.yml
paulgessinger: I suppose so yes.

paulgessinger:
@krasznaa updated.

krasznaa (approved):
👍

.github/workflows/gpu.yml
```diff
@@ -0,0 +1,160 @@
...
+#
+# (c) 2021-2026 CERN for the benefit of the ACTS project
+#
+# Mozilla Public License Version 2.0
+
+# Name for this "action".
+name: GPU Tests
+
+# Perform the builds/tests on every push and pull request.
+on: [ push, pull_request ]
```
krasznaa, line 11: Since I got an e-mail about a failure when pushing to your branch...

This will only work "on the main repository", right? Then I believe we need to be more selective here. Otherwise whenever we push stuff into our own feature branches, we'll get a failure e-mail. 🤔

krasznaa (changes requested):

.github/workflows/gpu.yml
paulgessinger: RIght, only on upstream. I limit the test jobs to the upstream repo now.

krasznaa (approved):

PR #370, opened 2026-09-02
Pin pre-commit hooks with SHA

m-fila:
For hardening. Tags aren't immutable so it's safer to use SHA of associated commits which would give an error if they are changed.
Updating with `pre-commit autoupdate --freeze` and `prek update --freeze` (also work with dependabot updates) will also update the human readable annotation.


krasznaa (approved):
No complaints.

PR #371, opened 2026-09-07
assert event wrapped by size(s) isn't nullptr

m-fila:
`async_sizes` and `async_size` happily dereference the pointers they were given. Normally this isn't a problem since they are constructed internally by other vecmem classes. Just adding checks for sanity.


core/include/vecmem/utils/impl/async_size.ipp
```diff
@@ -14,7 +14,10 @@ namespace vecmem {
 
 template <typename SIZE_TYPE>
 async_size<SIZE_TYPE>::async_size(storage_type size, event_type event)
-    : m_size{std::move(size)}, m_event{std::move(event)} {}
+    : m_size{std::move(size)}, m_event{std::move(event)} {
+    assert(m_size != nullptr);
+    assert(m_event != nullptr);
```
krasznaa: I would prefer to do it differently. 🤔 

Instead of requiring that `vecmem::async_size` must only be created with valid arguments, I would rather put the assertions to just before trying to use these pointers.

This would allow us to still create some dummy objects in case there comes the need to do so. Such that as long as the dummy object is not actually used, the code would be happy. But would still fail an assertion as soon as a client would try to use that dummy object.

Long story short, please add these lines into the (accessor) functions instead! 😉

core/include/vecmem/utils/async_size.hpp
```diff
@@ -37,8 +37,9 @@ class async_size : public abstract_event {
 
     /// Constructor taking ownership of a size and event
     ///
-    /// @param size Pointer to the size variable
-    /// @param event Event to wait on before accessing the size
+    /// @param size Pointer to the size variable. Must not be nullptr.
+    /// @param event Event to wait on before accessing the size. Must not be
+    /// nullptr.
```
krasznaa: And as long as we only test for non-null-pointers in the (accessor) functions, we should not change the descriptions either.

krasznaa (changes requested):

krasznaa (approved):
👍

